PR #245, opened 2018-07-23
Feature/video entry

### LINKED ISSUE #199: Embedded video template field
Students should be able to embed a youtube (or similar) video as a field in the entry.

Airgetfrog:
resolves #199 

lots of code duplication between the 3 entry components, would be nice for the future to refactor those but that's something for another project.

had to convert the url to and embed link immediately upon editing to make use of the existing check structure, would have been nice to have the regular url as data and convert only when needed, but the way the check works makes that unnecessarily complicated (no data <-> field type link), and this solution works quite well anyway.

### DIFF
--- a/src/vue/src/components/entry/EntryPreview.vue
+++ b/src/vue/src/components/entry/EntryPreview.vue
@@ -21,7 +21,7 @@
                     <b-form-file v-model="completeContent[i].data" :state="Boolean(completeContent[i].data)" placeholder="Choose a file..."></b-form-file><br><br>
                 </div>
                 <div v-else-if="field.type=='v'">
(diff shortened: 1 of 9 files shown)


src/vue/src/components/entry/EntryPreview.vue
```diff
@@ -4,8 +4,8 @@
 -->
 <template>
     <div>
-        <h2 class="mb-2">{{template.name}}</h2>
-        <b-card class="no-hover">
+        <b-card class="no-hover" :class="$root.getBorderClass(cID)">
+            <h2 class="mb-2">{{template.name}}</h2>
```
Larspolo: {{ template.name }}

PR #263, opened 2018-08-21
Feature/email verification

mjvkeulen:
## Description
Email verification was added for both LTI coupled users and individual registrations. Ontop of this response handling has been made consistent for both front and back end.


src/vue/src/views/EmailVerification.vue
```diff
@@ -0,0 +1,51 @@
...
+    },
+    components: {
+        'content-single-column': contentSingleColumn,
+        icon
+    }
+}
+</script>
+
+<style lang="sass">
+.center-content
```
engelhamer: Can be omitted by using bootstrap classes instead. (.text-center)

src/vue/src/views/EmailVerification.vue
```diff
@@ -0,0 +1,51 @@
...
+    text-align: center
+
+.center-content h2
```
engelhamer: This is not idiomatic Sass.

src/vue/src/views/ErrorPage.vue
```diff
@@ -2,7 +2,7 @@
     <b-row class="error-content">
         <b-col cols="12">
             <h1 id="error-heading">
-                Error {{ code }}: <span id="error-message">{{ message }}</span>
+                Error {{ code }}: <span id="error-message">{{ reasonPhrase }}</span>
```
engelhamer: Since there are no default values anymore there could be a situation in which this results in just ':', which is not very pretty.

src/vue/src/components/assets/PdfDisplay.vue
```diff
@@ -85,7 +85,7 @@ export default {
             }
         },
         password (updatePassword, reason) {
```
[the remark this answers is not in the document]
mjvkeulen: The PDF file can be locked by password, just needs a prompt to enter the password and some handling to unlock it.

src/vue/src/components/profile/PasswordCard.vue
```diff
@@ -25,22 +26,16 @@ export default {
...
-            } else {
-                this.$toasted.error('Passwords do not match')
+                    .then(response => { this.$toasted.success(response.data.description) })
+                    .catch(response => { this.$toasted.error(response.data.description) })
             }
         }
     },
     components: {
-        'icon': icon
+        icon
```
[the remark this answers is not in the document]
mjvkeulen: Updated for all components that allowed it due to our naming convention, as I think the shorthand is easy to understand and usefull.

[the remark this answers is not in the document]
engelhamer: Alright, sounds good!

src/django/VLE/views/create.py
```diff
@@ -287,22 +288,23 @@ def create_lti_user(request):
...
-    if len(password) < 8:
-        return responses.bad_request('Password needs to contain at least 8 characters.')
-    if password == password.lower():
-        return responses.bad_request('Password needs to contain at least 1 capital letter.')
-    if re.match(r'^\w+$', password):
-        return responses.bad_request('Password needs to contain a special character.')
+    try:
+        validators.validate_password(password)
+    except ValidationError:
+        return responses.bad_request('Invalid password format.')
```
Larspolo: Can't you send the error message back iso just invalid? Its nice to know why your password is wrong

src/django/VLE/views/responses.py
```diff
@@ -12,76 +12,76 @@
...
-
-
-def no_content(description='Request succeeded.'):
-    """Return a no content header.
-
-    Arguments:
-    description -- header description (usable for example in the front end)
-    """
-    return response(204, 'No Content', description=description)
+    return json_response(payload=payload, description=description, status=201)
```
Larspolo: Why change from 204 to 201?

src/django/VLE/views/responses.py
```diff
@@ -96,16 +96,49 @@ def response(status, message, description='', payload={}):
...
+
 def keyerror(*keys):
-    """Generate a bad request response when the input JSON has keyerror(s)."""
+    """Generate a bad request response with each given key formatted in the description."""
     if len(keys) == 1:
-        return bad_request('Field {0} is required but is missing.'.format(keys))
+        return bad_request(description='Field {0} is required but is missing.'.format(keys))
     else:
-        return bad_request('Fields {0} are required but one or more are missing.'.format(keys))
+        print('{0}'.format(keys))
```
Larspolo: remove this print

src/vue/src/api/auth.js
```diff
@@ -102,6 +91,19 @@ export default {
...
+    register (username, password, firstname, lastname, email, jwtParams = null) {
+        return connection.conn.post('/create_lti_user/', {
+            username: username,
+            password: password,
+            first_name: firstname,
+            last_name: lastname,
+            email: email,
+            jwt_params: jwtParams
+        })
+            .then(response => { return response.data.user })
```
Larspolo: `.then(response => response.data.user)`

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -174,9 +173,9 @@ export default {
         },
         addNode (infoEntry) {
             journalApi.create_entry(this.jID, infoEntry[0].tID, infoEntry[1])
-                .then(_ => journalApi.get_nodes(this.jID)
-                    .then(response => { this.nodes = response.nodes })
-                    .catch(_ => this.$toasted.error('Error while loading nodes.')))
+                .then(_ => { journalApi.get_nodes(this.jID) })
+                .then(data => { this.nodes = data.nodes })
+                .catch(error => { this.$toasted.error(error.response.data.description) })
```
Larspolo: Why did you change this? Because the second `.then` is from `get_nodes`. Does this still work?

src/vue/src/utils/status_codes.js
```diff
@@ -0,0 +1,13 @@
...
+const NOT_FOUND = 404
+const INTERNAL_SERVER_ERROR = 500
+
+export default {
+    BAD_REQUEST: BAD_REQUEST,
+    UNAUTHORIZED: UNAUTHORIZED,
+    FORBIDDEN: FORBIDDEN,
+    NOT_FOUND: NOT_FOUND,
+    INTERNAL_SERVER_ERROR: INTERNAL_SERVER_ERROR
+}
```
Larspolo: Missing a couple, 200, 201. I know they are not used, but for completion I think its better to include them anyway.

src/vue/src/utils/validation.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        return true
+    },
+
+    validateEmail (email, displayMessage = true) {
+        let re = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
+        var correctEmail = re.test(String(email).toLowerCase())
+
+        if (displayMessage) { Vue.toasted.error('The given email address is not valid!') }
```
Larspolo: Missing a return false, right?

src/vue/src/views/Assignment.vue
```diff
@@ -167,8 +168,8 @@ export default {
             }
 
             function compareUsername (a, b) {
```
Larspolo: This funciton really needs to be changed to a `function compare(a, b)` and then call it with `compare(a.student.name, b.student.name)`

src/django/VLE/views/create.py
mjvkeulen: The exact same test occur front end as well, so if this occurs the user has willingly messed with javascript. Nonetheless this can be done and I will update it accordingly.

src/vue/src/views/Assignment.vue
Larspolo: If you are not interested in fixing this, you can also just add a todo

src/django/VLE/views/responses.py
mjvkeulen: As 201 is created
https://httpstatuses.com/

Might be confused by the deletion of the (unused) no content response, moving the lines around.

src/vue/src/api/auth.js
mjvkeulen: Both do exactly the same

src/vue/src/utils/status_codes.js
mjvkeulen: We miss many, let them be included on the go?

src/vue/src/utils/validation.js
mjvkeulen: It returns the stored value of the regex test already (correctEmail)

src/vue/src/api/auth.js
Larspolo: We however use the example I gave in other files

src/vue/src/utils/status_codes.js
Larspolo: okay

PR #282, opened 2018-08-23
Fixes #218 for now.

### LINKED ISSUE #218: GDPR compliance
It should be clear what our software has to conform to, and properly test security to be entrusted with student details.

mjvkeulen:
## Description
A user can now download all of their respective user data, including files in one neat archive.

## Summary of changes
- Download of user data archive
- File structure changed to MEDIA_ROOT/uid/... instead of MEDIA_ROOT/files/uid/... This because the file directory was not needed, for the MEDIA_ROOT contains user files only.

image

### DIFF
--- a/src/django/VLE/views/get.py
+++ b/src/django/VLE/views/get.py
@@ -16,6 +16,7 @@ import jwt
 import VLE.lti_launch as lti
 import VLE.edag as edag
 import VLE.utils.generic_utils as utils
+import VLE.utils.file_handling as file_handling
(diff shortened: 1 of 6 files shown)


src/django/VLE/views/get.py
```diff
@@ -746,7 +747,8 @@ def get_entrycomments(request, eID):
 def get_user_data(request, uID):
     """Get the user data of the given user.
 
-    Get his/her profile data and posted entries with the titles of the journals of the user based on the uID.
+    Get his/her profile data and posted entries with the titles of the journals of the user based on the uID. As well
```
engelhamer: 'his/her' could be left out.

PR #283, opened 2018-08-23
Feature/sort group

JowehL:
## Description
Back-end and front-end implementation of group(except for deleting on the front-end)


src/django/VLE/factory.py
```diff
@@ -108,6 +108,21 @@ def make_assignment(name, description, author=None, format=None, lti_id=None,
...
+def make_course_group(name, course):
+    """Make a new course group.
+
+    Arguments:
+    name -- name of course group
+    course -- course the group belongs to
+
+    On success, returns a new course group.
+    """
+    course_group = Group(name=name, course=course)
```
mjvkeulen: return Group.objects.create(name = name, course = course) 

Is my preferred style, result is the same nonetheless.

src/django/VLE/models.py
```diff
@@ -190,11 +207,15 @@ class Participation(models.Model):
         on_delete=models.CASCADE,
         related_name='role',
     )
+    group = models.TextField(
+        default='',
+        null=False
```
mjvkeulen: Why was this solution decided opposed to null=True, default=None?

src/django/VLE/models.py
```diff
@@ -190,11 +207,15 @@ class Participation(models.Model):
...
+    )
 
     class Meta:
         """Meta data for the model: unique_together."""
 
-        unique_together = ('user', 'course',)
+        unique_together = ('user', 'course', 'group')
```
mjvkeulen: Would this not allow a user to be enrolled to the same course twice? But then under a different group.

src/django/VLE/models.py
JowehL: When filtering on the front-end, I am using the contains function which searches through a String. So instead of having another check function for the null value, I thought it would be better to have the group data type to be consistent

src/django/VLE/views/create.py
```diff
@@ -59,6 +59,44 @@ def create_new_course(request):
...
+    On success, returns a json string containing the course group.
+    """
+    user = request.user
+    if not user.is_authenticated:
+        return responses.unauthorized()
+
+    perm = permissions.get_permissions(user)
+
+    if not perm["can_add_course"]:
+        return responses.forbidden("You have no permissions to create a course.")
```
mjvkeulen: Incorrect error description, can add course user group deserves its own permission type as well.

src/django/VLE/models.py
JowehL: I forgot to undo this change, since this was done in the beginning

src/django/VLE/views/create.py
```diff
@@ -59,6 +59,44 @@ def create_new_course(request):
...
+        name -- name of the course group
+        cID -- course ID of the course
+
+    On success, returns a json string containing the course group.
+    """
+    user = request.user
+    if not user.is_authenticated:
+        return responses.unauthorized()
+
+    perm = permissions.get_permissions(user)
```
mjvkeulen: The permission for the creation of a user group is course level based, and no course id is passed. As such this will check if the user is a super user, and if not will return false for all permissions.

src/django/VLE/views/create.py
```diff
@@ -59,6 +59,44 @@ def create_new_course(request):
...
+
+    if not perm["can_add_course"]:
+        return responses.forbidden("You have no permissions to create a course.")
+
+    try:
+        name, cID = utils.required_params(request.data, "name", "cID")
+    except KeyError:
+        return responses.keyerror("name", "cID")
```
mjvkeulen: Please note that the keyerror will return a bad_request response as follows:

Fields ('name', 'cID') are required but one or more are missing. This error will be displayed as a toast front end, as such in the case of not easily read keys consider translating them to more human readable text.

src/django/VLE/views/create.py
```diff
@@ -59,6 +59,44 @@ def create_new_course(request):
...
+
+    try:
+        course = Course.objects.get(pk=cID)
+    except Course.DoesNotExist:
+        return responses.not_found('Course does not exist.')
+
+    if len(Group.objects.filter(name=name, course=course)):
+        course_group = factory.make_course_group(name, course=course)
+    else:
+        return responses.response('409', 'Course group already exists', 'Course group already exists')
```
mjvkeulen: 409 Conflict
Indicates that the request could not be processed because of conflict in the current state of the resource, such as an edit conflict between multiple simultaneous updates.

Please use an already defined response, not_found with an appropriate description would be suitable here.

src/django/VLE/views/delete.py
```diff
@@ -44,6 +44,39 @@ def delete_course(request):
...
+    try:
+        name, cID = utils.required_params(request.data, 'name', 'cID')
+    except KeyError:
+        return responses.keyerror('name', 'cID')
+
+    role = permissions.get_role(user, cID)
+    if role is None:
+        return responses.unauthorized(description="You have no access to this course")
+    elif not role.can_delete_course:
+        return responses.forbidden(description="You have no permissions to delete a course.")
```
mjvkeulen: Same here, the response requires an appropriate description and we require a more fine grained permission for this functionality.

src/django/VLE/views/delete.py
```diff
@@ -44,6 +44,39 @@ def delete_course(request):
...
+
+    course = Course.objects.get(pk=cID)
+    course_group = Group.objects.filter(name=name, course=course)
```
mjvkeulen: This wont work, whereas by now it is known that the given cID exists (otherwise role would have been none). The given course group name is not checked yet. 
As filter returns a query set this can be empty, and should be checked first. If not empty something other than success should be returned

src/django/VLE/models.py
```diff
@@ -125,6 +125,23 @@ def __str__(self):
...
+    A group
+    - name: name of the group
+    """
+
+    name = models.TextField()
+    course = models.ForeignKey(
+        Course,
+        on_delete=models.CASCADE
+    )
+
```
mjvkeulen: I feel a 

`    class Meta:
        """Meta data for the model: unique_together."""

        unique_together = ('name', 'course')`

Seems appropriate

src/django/VLE/views/update.py
```diff
@@ -435,6 +435,42 @@ def update_user_role_course(request):
...
+    except KeyError:
+        return responses.keyerror("group", "uID", "cID")
+
+    try:
+        course = Course.objects.get(pk=cID)
+        participation = Participation.objects.get(user=uID, course=cID)
+    except (Participation.DoesNotExist, Course.DoesNotExist):
+        return responses.not_found('Participation or Course does not exist.')
+
+    participation.group = ''
```
mjvkeulen: Again am a big fan of None

src/vue/src/views/CourseEdit.vue
```diff
@@ -135,7 +154,8 @@ export default {
     created () {
         courseApi.get_course_data(this.cID)
             .then(response => {
```
mjvkeulen: You renamed this, this should no longer be called response but data (as you named it)

src/vue/src/api/course.js
```diff
@@ -92,6 +108,15 @@ export default {
         }).then(response => response.data.result)
     },
 
+    /* Updates the role of a student linked to a course. */
+    update_user_course_group (uID, cID, group) {
+        return auth.authenticatedPost('/update_user_course_group/', {
+            uID: uID,
+            cID: cID,
+            group: group
+        }).then(response => response.data.result)
```
mjvkeulen: You return this in update_user_course_group:

return responses.success(payload={'new_group': participation.group})

As such response.data.result does not exist!, response.data.new_group does.

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+    <div>
+        <b-btn class="change-button flex-grow-1 multi-form" v-b-modal="'myModal'">Manage Course Groups</b-btn>
+
+        <b-modal id="myModal"
```
mjvkeulen: An id reflecting the purpose of the modal would be preferred

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+                 title="Manage Course Groups"
+                 hide-footer
+                 >
+
+            <b-form @submit.prevent="onSubmit" @reset.prevent="onReset">
+                <b-input class="mb-2 mr-sm-2 mb-sm-0 multi-form theme-input" v-model="form.groupName" placeholder="Group name" required/>
```
mjvkeulen: theme-input multi-form would suffice, placeholder Desired group name would have my preference

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+    <div>
+        <b-btn class="change-button flex-grow-1 multi-form" v-b-modal="'myModal'">Manage Course Groups</b-btn>
+
+        <b-modal id="myModal"
+                 title="Manage Course Groups"
+                 hide-footer
+                 >
+
+            <b-form @submit.prevent="onSubmit" @reset.prevent="onReset">
```
mjvkeulen: Strongly prefer verbose names, so on submit createUserGroup for example

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+            form: {
+                groupName: '',
+                groupTA: ''
+            }
+        }
+    },
+    methods: {
+        onSubmit () {
+            courseApi.create_new_course_group(this.form.groupName, this.cID)
+                .then(response => {
```
mjvkeulen: Again this is not the response object, but the response.data object. Please let the name reflect so.

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+                    this.$emit('create-group', this.form.groupName)
+                    this.$toasted.success('Group creation succes.')
+                    this.onReset(undefined)
+                })
+                .catch(response => {
```
mjvkeulen: This is an error object, accessing the description would be, error.response.data.description

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    },
+    methods: {
+        onSubmit () {
+            courseApi.create_new_course_group(this.form.groupName, this.cID)
+                .then(response => {
+                    this.$emit('create-group', this.form.groupName)
+                    this.$toasted.success('Group creation succes.')
+                    this.onReset(undefined)
```
mjvkeulen: Why is this needed?

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+                })
+                .catch(response => {
+                    this.$toasted.error(response.data.description)
+                })
+        },
+        onReset (evt) {
```
mjvkeulen: This function seems weird, are you sure this needs to be done in such a manner?

`resetCourseGroupInput () {
                this.form.groupName = Null
                this.form.groupTA = Null
}`

Should do the trick being reactive and all?

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+            if (evt !== undefined) {
```
mjvkeulen: This could be removed

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
...
+                evt.preventDefault()
+            }
+
+            /* Reset our form values */
+            this.form.groupName = ''
+            this.form.groupTA = ''
+
+            /* Trick to reset/clear native browser form validation state */
```
mjvkeulen: This too

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -20,6 +20,16 @@
                         </option>
                     </b-form-select>
                 </div>
+                <div class="shadow">
+                    <b-form-select v-if="this.$root.canEditCourseRoles"
```
mjvkeulen: I would move the if check to the actual container

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -20,6 +20,16 @@
...
+                                   v-model="selectedGroup"
+                                   :select-size="1">
+                        <option :value="''">No group</option>
```
mjvkeulen: Should the default not actually be retrieved rather than hard coded?

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -58,11 +68,18 @@ export default {
...
+            required: true
+        },
+        groups: {
+            required: true
         }
     },
     data () {
         return {
             selectedRole: '',
+            selectedGroup: '',
```
mjvkeulen: Initialization with Null has the preference (admittedly its a mess currently) but think twice when you use '' instead of Null

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+    <div>
+        <b-btn class="change-button flex-grow-1 multi-form" v-b-modal="'myModal'">Manage Course Groups</b-btn>
```
mjvkeulen: I believe a Manage Course Groups button is indeed required, currently this functionality is simply add course group. 
We need a way to list all current course groups and rename/delete them. Right idea, some more functionality required.

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -110,10 +127,29 @@ export default {
...
+                this.init = false
+            } else {
+                this.selectedGroup = val
+                this.$emit('update:group', val)
+                courseApi.update_user_course_group(
+                    this.uID,
+                    this.cID,
+                    this.selectedGroup)
+                    .then(_ => {
+                        this.checkPermission()
```
mjvkeulen: I dont believe a permission update check is required for the assignment of a user group

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -110,10 +127,29 @@ export default {
...
+                    })
+                    .catch(_ => {
+                        this.$toasted.error('Error while changing user group.')
```
mjvkeulen: Please catch the error inline with all other error handling.

src/vue/src/views/CourseEdit.vue
```diff
@@ -36,6 +36,10 @@
                             <icon name="users"/>
                             Manage Roles and Permissions
                         </b-button>
+                        <group-modal v-if="this.$root.canEditCourseRoles()"
+                                     :cID="this.cID"
+                                     @create-group="createGroup">
```
mjvkeulen: The emitting could be avoided by simple binding group object and updating it. This operation is reactive.

src/vue/src/views/CourseEdit.vue
```diff
@@ -51,14 +55,24 @@
...
                                <option :value="null">Sort by ...</option>
                                <option value="sortFullName">Sort by name</option>
                                <option value="sortUsername">Sort by username</option>
                             </b-form-select>
                         </b-col>
-                        <b-col sm="6" class="d-flex flex-wrap">
+                        <b-col sm="4" class="d-flex flex-wrap">
+                            <b-form-select v-model="selectedFilterGroupOption"
+                                           :select-size="1">
+                                <option :value="''">Filter group by ...</option>
```
mjvkeulen: null vs '' not that it makes a difference here

src/vue/src/views/CourseEdit.vue
```diff
@@ -241,6 +266,14 @@ export default {
                 }
             }
 
+            function checkGroup (user) {
+                if (user.group.includes(self.selectedFilterGroupOption)) {
```
mjvkeulen: Is this not a boolean check in itself? (And could thus be returned)

src/vue/src/views/CourseEdit.vue
```diff
@@ -254,20 +287,21 @@ export default {
...
+            return viewList.filter(checkFilter).filter(checkGroup)
         }
     },
     components: {
         'add-user-card': addUsersToCourseCard,
         'bread-crumb': breadCrumb,
         'content-single-column': contentSingleColumn,
         'course-participant-card': courseParticipantCard,
-        'icon': icon
+        'icon': icon,
```
mjvkeulen: We switched from 'icon': icon, to icon where possible.

src/django/VLE/views/create.py
Larspolo: I dont agree. Maybe this then has to change in the frontend, but as a debugging purpose, its really nice to immediately see which keys are missing. Also, it our platform is bug free, they will never see this in the first place. Keyerror is really mostly for developers

src/django/VLE/models.py
```diff
@@ -53,18 +47,16 @@ class User(AbstractUser):
 
     User is an entity in the database with the following features:
     - email: email of the user.
-    - verified_email: Boolean to indicate if the user has validated their email adress.
```
Larspolo: Why is this removed again?

src/django/VLE/models.py
```diff
@@ -510,16 +532,12 @@ class Field(models.Model):
...
-    RICH_TEXT = 'rt'
     IMG = 'i'
     FILE = 'f'
     VIDEO = 'v'
-    PDF = 'p'
     TYPES = (
         (TEXT, 'text'),
-        (RICH_TEXT, 'rich text'),
         (IMG, 'img'),
-        (PDF, 'pdf'),
```
Larspolo: And why are all these removed as well?

src/django/VLE/views/create.py
```diff
@@ -271,10 +307,10 @@ def create_lti_user(request):
     """
     if request.data['jwt_params'] is not '':
         lti_params = jwt.decode(request.data['jwt_params'], settings.LTI_SECRET, algorithms=['HS256'])
-        lti_id, user_image = lti_params['user_id'], lti_params['user_image']
+        user_id, user_image = lti_params['user_id'], lti_params['user_image']
```
Larspolo: Why did you change this to user_id? I think lti_id or lti_user_id are better in this case as user_id is used for the django id, not the lti id

src/django/VLE/views/create.py
```diff
@@ -288,23 +324,22 @@ def create_lti_user(request):
...
-    try:
-        validators.validate_password(password)
-    except ValidationError as e:
-        return responses.bad_request(e.args[0])
+    if len(password) < 8:
+        return responses.bad_request('Password needs to contain at least 8 characters.')
+    if password == password.lower():
+        return responses.bad_request('Password needs to contain at least 1 capital letter.')
+    if re.match(r'^\w+$', password):
+        return responses.bad_request('Password needs to contain a special character.')
```
Larspolo: Seems like you reverse a lot of things. How did all of this happen?

src/django/VLE/views/delete.py
```diff
@@ -71,18 +108,19 @@ def delete_assignment(request):
...
     assignment = Assignment.objects.get(pk=aID)
     assignment.courses.remove(course)
     assignment.save()
+    response['removed_from_course'] = True
     if assignment.courses.count() == 0:
         assignment.delete()
         response['removed_completely'] = True
 
-    return responses.success(description='Succesfully deleted the assignment.', payload=response)
+    return responses.success(message='Succesfully deleted assignment', payload=response)
```
Larspolo: Should be description

src/django/VLE/views/delete.py
```diff
@@ -41,7 +41,44 @@ def delete_course(request):
...
+
+    course = Course.objects.get(pk=cID)
+    course_group = Group.objects.get(name=name, course=course)
+
+    if course_group:
+        course_group.delete()
+    else:
+        return responses.not_found('Course group does not exist.')
+
+    return responses.success(message='Succesfully deleted course group')
```
Larspolo: Should be description=

src/django/VLE/views/get.py
```diff
@@ -81,7 +81,12 @@ def get_course_data(request, cID):
 
     course = serialize.course_to_dict(q_course)
 
-    return responses.success(payload={'course': course})
+    course_groups = Group.objects.filter(course=q_course)
+    course_groups = [serialize.group_to_dict(course_group)
+                     for course_group in course_groups]
+
+    return responses.success(payload={'course': course,
+                                      'course_groups': course_groups})
```
Larspolo: Can be on 1 line prob

src/django/VLE/views/get.py
```diff
@@ -81,7 +81,12 @@ def get_course_data(request, cID):
 
     course = serialize.course_to_dict(q_course)
 
-    return responses.success(payload={'course': course})
+    course_groups = Group.objects.filter(course=q_course)
+    course_groups = [serialize.group_to_dict(course_group)
+                     for course_group in course_groups]
```
Larspolo: Same here (but will also change with the new rest-api)

Larspolo:
Closing this because there is a new feature/sort-group-rest-api branch

PR #291, opened 2018-08-28
Feature/vuex

### LINKED ISSUE #231: Vuex & Localstorage for global statemanagement client side

### LINKED ISSUE #265: URL template field
# Feature request

## Description
Add a field to the template editor that handles URLs.

## User story
A teacher should be able to add a URL field to an entry template.

### LINKED ISSUE #287: Auto select first ungraded entry when navigating as grader
# Feature request

## Description
Auto select first ungraded entry when navigating as grader to make the graders flow more intuitive.

mjvkeulen:
## Description
Implements vuex as local store for generic user data and all permissions associated with the respective user.

## Fixes
- Fixed #231
- Fixed #287
- Fixed #265


src/django/VLE/views/get.py
```diff
@@ -377,17 +358,11 @@ def create_teacher_assignment_deadline(course, assignment):
...
-        return {}
-
-    deadline = format['presets'][0]['deadline']
-
     return {'name': serialize.assignment_to_dict(assignment)['name'],
             'courseAbbr': course.abbreviation,
             'cID': course.id,
             'aID': assignment.id,
-            'deadline': deadline,
+            'deadline': datetime.datetime.now(),  # TODO placeholder untill the assignment end dates are set.
```
Larspolo: Why not use the upcomming deadline? (like the one that was used)

src/django/VLE/views/get.py
```diff
@@ -732,27 +691,21 @@ def get_entrycomments(request, eID):
...
-    permission = permissions.get_permissions(user, cID=-1)
-    if not (permission['is_superuser'] or request.user.id == uID):
-        return responses.forbidden('You cannot view this users data.')
-
     profile = serialize.user_to_dict(user)
     # Don't send the user id with it.
     del profile['uID']
 
-    journals = Journal.objects.filter(user=uID)
+    journals = Journal.objects.filter(user=user)
```
Larspolo: superusers arent allowed to download everything from someone anymore?

src/vue/src/components/assets/TextEditor.vue
```diff
@@ -115,14 +107,14 @@ export default {
...
-                toolbar1: 'bold italic underline alignleft aligncenter alignright alignjustify | forecolor backcolor restoredraft | formatselect | bullist numlist | image media table | removeformat fullscreentoggle',
+                toolbar1: 'bold italic underline alignleft aligncenter alignright alignjustify | forecolor backcolor restoredraft | formatselect | bullist numlist | image media table | removeformat fullscreentoggle fullscreen',
                 plugins: [
                     'autoresize paste textcolor image lists wordcount autolink autosave',
                     'table media fullscreen'
                 ]
             },
             extensiveConfig: {
-                toolbar1: 'bold italic underline alignleft aligncenter alignright alignjustify | forecolor backcolor | formatselect | bullist numlist | image media table | removeformat fullscreentoggle',
+                toolbar1: 'bold italic underline alignleft aligncenter alignright alignjustify | forecolor backcolor | formatselect | bullist numlist | image media table | removeformat fullscreentoggle fullscreen',
```
Larspolo: extensiveConfig has less confiruration then basic config? Why?

mjvkeulen: This is just the specific toolbar1 slot, there is also the menu bar which is cramped with additional buttons

src/django/VLE/permissions.py
```diff
@@ -99,7 +99,10 @@ def get_permissions(user, cID=-1):
         if not role:
             return {}
 
-        roleDict = model_to_dict(role)
+        roleDict = model_to_dict(role)  # TODO This even adds keys such as 'id', 'name' and 'course', prob not wanted?
+        del roleDict['id']
+        del roleDict['name']
+        del roleDict['course']
```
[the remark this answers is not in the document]
mjvkeulen: Nothing will break if we do not delete these keys, yet I think its strange the permission specific function would return these. (Simply because we use the built in model_to_dict function. Why do you need the name to permission relation? Surely if you have the user this relation is not required?

src/django/VLE/views/get.py
mjvkeulen: I believe it was a dud, sadly the entire deadlines will require a rework.

src/django/VLE/views/get.py
mjvkeulen: Well superusers have database access right?

PR #292, opened 2018-08-31
Feature/rest api

### LINKED ISSUE #222: More properly use Django REST
Django REST has an amazing feature set that we do not use properly.

Larspolo:
## Description
Large rework in how the backend is structured.
- The api has moved from a semi-CRUD structure, to a full on restful-api
- Serializers are now done by using the django viewset serializers

## Summary of changes
- Fixes #222


src/django/VLE/views/course.py
```diff
@@ -0,0 +1,230 @@
...
+course.py.
+
+In this file are all the course api requests.
+"""
+from rest_framework import viewsets
+from rest_framework.decorators import action
+
+import VLE.views.responses as response
+import VLE.serializers as serialize
+# from VLE.serializers import CourseSerializer
```
mjvkeulen: Uncommented imports

src/django/VLE/serializers.py
```diff
@@ -3,288 +3,401 @@
...
+import statistics as st
+
+
+class UserSerializer(serializers.ModelSerializer):
+    name = serializers.SerializerMethodField()
+    role = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        fields = ('last_login', 'username', 'first_name', 'last_name', 'is_active', 'email', 'name',
```
mjvkeulen: Missing field verified_email

Larspolo: Fixed

PR #293, opened 2018-09-01
Enhancement/template preview

mjvkeulen:
## Description
Addition of a template preview as well as automated LTI flow if the user has no associated courses.

image

### DIFF
--- a/src/vue/src/components/lti/LtiCreateConnectUser.vue
+++ b/src/vue/src/components/lti/LtiCreateConnectUser.vue
@@ -1,23 +1,25 @@
 <template>
     <div>
-        <p class="lti-intro-text">Welcome to eJournal! Would you like to couple with a previously registered account or
-            would you like us to create an account for you?</p>
+        <span class="multi-form">If you dont have an account on eJournal, you can create an account below. This account will be linked to your
+        canvas account from then on.</span>
         <b-row align-h="center">
-            <b-button class="lti-button-option" @click="showModal('createUserRef')">
+            <b-button class="lti-button-option multi-form" @click="showModal('createUserRef')">
                 <icon name="user-plus" scale="1.8"/>
                 <h2 class="lti-button-text">Create account</h2>
             </b-button>
         </b-row>
+        <span class="multi-form">If you already have an account on eJournal, and would like to link this account to your
+        canvas account. Please press the button below.</span>
         <b-row  align-h="center">
             <b-button class="lti-button-option" @click="showModal('connectUserRef')">
                 <icon name="link" scale="1.8"/>
-                <h2 class="lti-button-text">Couple account</h2>
+                <h2 class="lti-button-text">Link with existing <br/> eJournal account</h2>
             </b-button>
         </b-row>
 
         <b-modal
             ref="createUserRef"
-            title="New User"
+            title="Create eJournal account that will be linked with your canvas account."
             size="lg"
             hide-footer>
                 <register-user @handleAction="handleRegistered" :lti="lti"/>
@@ -25,7 +27,7 @@
 
         <b-modal
             ref="connectUserRef"
-            title="Connect User"
+            title="Login with your eJournal account to link with your canvas acccount."
             size="lg"
             hide-footer>
                 <login-form @handleAction="handleConnected"/>
--- a/src/vue/src/utils/generic_utils.js
+++ b/src/vue/src/utils/generic_utils.js
@@ -1,0 +1,9 @@
+export default {
+    yearOffset (startDate) {
+        let split = startDate.split('-')
+        let yearOff = parseInt(split[0]) + 1
+
+        split[0] = String(yearOff)
+        return split.join('-')
+    }
+}
--- a/src/vue/src/views/LtiLaunch.vue
+++ b/src/vue/src/views/LtiLaunch.vue
@@ -2,9 +2,14 @@
     <content-single-columns>
         <h1 class="mb-2">{{ currentPage }}</h1>
         <b-card class="no-hover" :class="this.$root.colors[1]">
-            <lti-create-connect-course v-if="handleCourseChoice" @handleAction="handleActions" :lti="lti"/>
+
+            <lti-create-connect-course v-if="handleCourseChoice" @handleAction="handleActions" :lti="lti" :courses="courses"/>
             <lti-create-connect-assignment v-else-if="handleAssignmentChoice" @handleAction="handleActions" :lti="lti" :page="page"/>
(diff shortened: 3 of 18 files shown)


src/vue/src/components/lti/LtiCreateConnectUser.vue
```diff
@@ -1,31 +1,33 @@
 <template>
     <div>
-        <p class="lti-intro-text">Welcome to eJournal! Would you like to couple with a previously registered account or
-            would you like us to create an account for you?</p>
+        <span class="multi-form">If you dont have an account on eJournal, you can create an account below. This account will be linked to your
+        canvas account from then on.</span>
```
engelhamer: Generic LTI (learning environment) instead of Canvas?

src/vue/src/components/lti/LtiCreateConnectUser.vue
```diff
@@ -1,31 +1,33 @@
...
                 <icon name="link" scale="1.8"/>
-                <h2 class="lti-button-text">Couple account</h2>
+                <h2 class="lti-button-text">Link with existing <br/> eJournal account</h2>
             </b-button>
         </b-row>
 
         <b-modal
             ref="createUserRef"
-            title="New User"
+            title="Create eJournal account that will be linked with your canvas account."
```
engelhamer: '...that will be linked to your Canvas account' is excessive and not generic enough. 'Create eJournal account' should be fine.

src/vue/src/components/lti/LtiCreateConnectUser.vue
```diff
@@ -1,31 +1,33 @@
...
             size="lg"
             hide-footer>
                 <register-user @handleAction="handleRegistered" :lti="lti"/>
         </b-modal>
 
         <b-modal
             ref="connectUserRef"
-            title="Connect User"
+            title="Login with your eJournal account to link with your canvas acccount."
```
engelhamer: Again, Canvas should be replaced.

src/vue/src/utils/generic_utils.js
```diff
@@ -0,0 +1,9 @@
+export default {
+    yearOffset (startDate) {
```
engelhamer: This will return an invalid date when the 29th of february is used as a startDate.

src/vue/src/views/LtiLaunch.vue
```diff
@@ -2,9 +2,14 @@
     <content-single-columns>
         <h1 class="mb-2">{{ currentPage }}</h1>
         <b-card class="no-hover" :class="this.$root.colors[1]">
-            <lti-create-connect-course v-if="handleCourseChoice" @handleAction="handleActions" :lti="lti"/>
+
```
engelhamer: Excessive whitespace.

src/vue/src/views/LtiLaunch.vue
```diff
@@ -139,10 +157,9 @@ export default {
                 this.createAssignment = true
                 break
             case this.states.check_assign:
-                assignApi.get_assignment_by_lti_id(this.lti.ltiAssignID)
+                assignmentApi.get_assignment_by_lti_id(this.lti.ltiAssignID)
                     .then(response => {
                         if (response === undefined) {
-                            console.log('UNDEFINED RESPONSE JE MOEDER')
```
engelhamer: How did this ever end up here 😅

src/vue/src/utils/generic_utils.js
mjvkeulen: This expects a datetime format inline with the db, eg "2018-02-29". Not sure how this would break? Will add some documentation and its definitely not pretty.

src/vue/src/components/lti/LtiCreateConnectUser.vue
engelhamer: But why is it necessary to state that in the modal title?

mjvkeulen: Ah very true, not required here. Point still holds for the other references of canvas vs LTI.

src/vue/src/utils/generic_utils.js
engelhamer: Yes, but the 29th of february is intercalary day.

src/vue/src/components/lti/LtiCreateConnectUser.vue
engelhamer: Sure, but I would still prefer 'learning environment' over the name of a specific VLE in our code.

mjvkeulen: I think that works

src/vue/src/utils/generic_utils.js
mjvkeulen: Slipped my mind, very true good spot.

src/django/VLE/management/commands/preset_db.py
```diff
@@ -21,42 +21,44 @@ def gen_users(self):
         """Generate users with password 'pass'."""
         users_examples = [
             {
-                "username": "22222222",
+                "username": "Student",
                 "first_name": "Lars",
                 "last_name": "van Hijfte",
+                "email": "larsderpolo@gmail.com",
```
Larspolo: Nice addition of making an account without a verified_email. But maybe we need to change the email to another email that everyone can access. Because i dont want to be spammed with verification emails

src/vue/src/views/LtiLaunch.vue
Larspolo: GIT BLAME

src/vue/src/views/LtiLogin.vue
```diff
@@ -81,6 +81,7 @@ export default {
 
             /* The LTI parameters are verified in our backend, however there is no corresponding user yet. We must create/connect one. */
             } else if (this.$route.query.state === this.states.no_user) {
+                this.$store.commit('user/LOGOUT') // Ensures no old user is loaded from local storage.
```
Larspolo: Why not? If you want to link a lot of courses/assignments its nice to stay logged in

Larspolo:
Also just a comment, this is more then just the `Enhancement/template preview`, maybe you want to change the description of the other things you did as well on this branch

src/vue/src/views/LtiLogin.vue
mjvkeulen: Agreed, however this state is only reached upon login and verification of the LTI jwt. If there is no user matching in our system for the given (valid) jwt, we need to create one. In this scenario the store is always outdated, and this ensures the store is purged before we begin.

engelhamer:
This will come with #292.

PR #297, opened 2018-09-06
Feature/optional template field

### LINKED ISSUE #267: Non-mandatory template fields
# Feature request

## Description
Currently, all fields in an entry template are mandatory. Sometimes this is not desirable.

## User story
As a teacher I want to specify whether a template field is mandatory.

## User interface
A simple tickbox in the template editor should be sufficient.
If a template field is not mandatory and not filled in by the student it should be hidden when viewing the entry.

### LINKED ISSUE #276: Sticky grading section
# Feature request

## Description
For long journal entries it can be a hassle to scroll all the way back to the top in order to grant points in the grading section. If we make this section stick to the top of the window after scrolling teachers can grade quicker.

## User story
As a teacher I want to always have the grading of an entry in view, unless that entry is out of sight.

## User interface
The grading section shall move within an entry to remain in the window for as long as possible.

Airgetfrog:
## Description
Added sticky grading and required/optional toggle on template fields

## Summary of changes
- Fixes #267 
- Fixes #276


src/vue/src/components/entry/EntryNode.vue
```diff
@@ -24,7 +24,7 @@
             -->
             <div v-for="(field, i) in entryNode.entry.template.fields" :key="field.eID">
                 <div v-if="field.title">
-                    <b>{{ field.title }}</b>
+                    <b>{{ field.title }}</b> <b style="color: red" v-if="field.required">*</b>
```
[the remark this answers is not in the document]
engelhamer: We do not use inline style anywhere. I suggest making a new red-text class (please use our theme red color by the way!) in the global sass file.

Larspolo:
Fixed the login issue by merging it with develop. Also changed the height to 10px iso 84.

Larspolo:
unknown

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -33,6 +41,7 @@
 import ContentSingleColumn from '@/components/columns/ContentSingleColumn.vue'
 import icon from 'vue-awesome/components/Icon'
 import draggable from 'vuedraggable'
+import Switch from '@/components/assets/SwitchComponent.vue'
```
engelhamer: Imported but never used?

src/vue/src/components/entry/EntryNonStudentPreview.vue
```diff
@@ -178,3 +178,9 @@ export default {
     }
 }
 </script>
+
+<style lang="sass">
+.sticky
```
engelhamer: This should be moved to the sass file for entries, nested within the grade-section class. That way we can define a sticky class for other elements if we want to do so later on, but won't have to take on ugly names with suffixes in them.

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -91,6 +102,18 @@ export default {
 @import '~sass/modules/colors.sass'
 @import '~sass/modules/breakpoints.sass'
 
+.optional
```
engelhamer: We might want to use these later on, maybe move them to the button sass file?

PR #298, opened 2018-09-06
After summer, before rest-api

Larspolo:
## Description
This is all the code that doesnt have to reviewd as it is already done before it was pushed to develop.


engelhamer:
Looks all good to me!

PR #302, opened 2018-09-07
Added and fixed the representation of progresses of journals

### LINKED ISSUE #209: Visual representation of journal progress compared to average
See title.

### LINKED ISSUE #228: 'Needs marking' statistics are incorrectly computed
The 'Needs marking' count for assignment statistics is incremented for every entry in a journal for which a grade has not yet been published. This is no correct behaviour for this function.

There shall be a different function which computes the 'Unpublished' count. 'Needs marking' should only cover entries that do not even have an unpublished mark.

### LINKED ISSUE #250: Remove assignment from course without deleting assignment from all courses
It is currently impossible to remove an assignment from a course without deleting it entirely. Possible fixes might be adding a remove button on assignments cards in courses (and deleting automatically if it was the only course using that assignment), or a similar solution in the course editor screen.

Larspolo:
## Description
Added a visual representation of the progress compared to the average when viewing a journal.
Furthermore the needs_marking is correctly calculated, and also the unpublished stat is added.

## Summary of changes
- Fixes #209 
- Fixes #228 
- Fixed #250 (this was already implemented in the backend, the only thing I did was visually change that you now 'remove an assignment from course' iso 'delete assignment')
Teacher view
Student view


src/django/VLE/utils/generic_utils.py
```diff
@@ -84,6 +84,16 @@ def get_graded_count(entries):
 
     - entries: the entries to count with.
 
+    Returns the graded entry count.
+    """
+    return entries.exclude(grade=None).count()
+
+
+def get_published_count(entries):
+    """Count the number of graded entries.
```
Airgetfrog: Shouldn't the docstring be changed to match?

src/vue/src/components/assets/ProgressBar.vue
```diff
@@ -3,17 +3,71 @@
...
+        'totalPoints': {
+            required: true
+        },
+        'comparePoints': {
+            default: -1
+        }
+    },
+    data () {
+        return {
+            aheadClass: 'ahead'
```
Airgetfrog: Vestigial code?

Larspolo:
image

PR #303, opened 2018-09-07
Added asc and desc option in AssignmentOverview and Assignment

### LINKED ISSUE #183: Choose whether to sort journals ascending or descending
Note that this should be changed on both the assignment page and journalNonStudent page.

Larspolo:
## Description
Added a order type button and moved some buttons around to make it look all nice and pretty again :wink: 
image

## Summary of changes
- Fixes #183


src/vue/src/views/Assignment.vue
```diff
@@ -4,20 +4,30 @@
         <bread-crumb slot="main-content-column" @eye-click="customisePage" @edit-click="handleEdit()"/>
         <b-card slot="main-content-column" class="no-hover settings-card">
             <b-row>
-                <b-col sm="6">
-                    <b-form-select class="multi-form" v-model="selectedSortOption" :select-size="1">
+                <b-col sm="12">
+                    <input class="theme-input full-width multi-form" type="text" v-model="searchVariable" placeholder="Search..."/>
+                </b-col>
+                <b-col sm="8">
+                    <b-form-select class="multi-form btn" v-model="selectedSortOption" :select-size="1">
```
Airgetfrog: These probably shouldn't be classed as `btn`'s, it causes the entire dropdown to have a blue background while active. Ditto for the one below

src/vue/src/views/AssignmentsOverview.vue
```diff
@@ -4,16 +4,27 @@
 
         <b-card class="no-hover">
                 <b-row>
-                    <b-col sm="6">
-                        <b-form-select v-model="selectedSortOption" :select-size="1">
+                    <b-col sm="12">
+                        <input class="theme-input full-width multi-form" type="text" v-model="searchVariable" placeholder="Search..."/>
+                    </b-col>
+                    <b-col sm="8">
+                        <b-form-select class="btn" v-model="selectedSortOption" :select-size="1">
```
Airgetfrog: Same as above

Airgetfrog:
Probably good to add this to "manage course members" too: 

image

Also, I'd prefer using a simple/more generic "ascending/descending" over the current "A-Z/1-9".

src/vue/src/views/Assignment.vue
engelhamer: I guess we should copy the required styling from the btn class (in buttons.sass) to the select tag (in select.sass) and indeed leave the changing background behind.

Larspolo:
Fixed manage course members and implemented the other requested changes
image

Larspolo:
image
image
Added icons

PR #307, opened 2018-09-09
Bugfix/open bugfixes

### LINKED ISSUE #236: Decimal or empty point input results in a 500 error
When assigning a decimal grade to an entry the server throws a 500 because it is not an integer. Support for decimal grades would be nice.

Edit: This also happens when no grade is given at all (i.e. the input field is empty)
Edit: This also applies to the point maximum in the assignment format

### LINKED ISSUE #306: Course creation throws an error
# Bug report

## Expected Behavior
A course should be created with relevant values and no error.

## Actual Behavior
An error is thrown but the course is still created with the correct values.
image

1. Attempt to create a course

Larspolo:
## Description
Some (minor) bug fixes to improve the quality of the website, and make the demo more user friendly.

## Summary of changes
- Fixes #236 (also it wont crash when inserting text)
- Fixes #306 (also the permissions will load properly)
- and other unlisted bugs

### DIFF
--- a/src/vue/src/components/entry/EntryNonStudentPreview.vue
+++ b/src/vue/src/components/entry/EntryNonStudentPreview.vue
@@ -7,7 +7,7 @@
     <div v-if="entryNode.entry !== null">
         <b-card class="entry-card no-hover entry-card-teacher" :class="$root.getBorderClass($route.params.cID)">
             <div v-if="$hasPermission('can_grade_journal')" class="grade-section shadow sticky">
-                <b-form-input class="theme-input" type="number" size="2" v-model="grade" placeholder="0" min=0></b-form-input>
+                <b-form-input class="theme-input" type="number" step="0.01" size="2" v-model="grade" autofocus placeholder=".." min="0.0"></b-form-input>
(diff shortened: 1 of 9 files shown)


src/vue/src/components/entry/EntryNonStudentPreview.vue
```diff
@@ -7,7 +7,7 @@
     <div v-if="entryNode.entry !== null">
         <b-card class="entry-card no-hover entry-card-teacher" :class="$root.getBorderClass($route.params.cID)">
             <div v-if="$hasPermission('can_grade_journal')" class="grade-section shadow sticky">
-                <b-form-input class="theme-input" type="number" size="2" v-model="grade" placeholder="0" min=0></b-form-input>
+                <b-form-input class="theme-input" step="0.01" size="2" v-model="grade" autofocus placeholder=".." min="0.0"></b-form-input>
```
mjvkeulen: Looks good apart from the placeholder, something else to specify that this field is the grade of the entry is required.

Larspolo: Fixed

PR #318, opened 2018-09-12
Feature/default format

### LINKED ISSUE #169: Provide a default format upon creation of an assignment
Ideally we should create assignments with a format including a simple default template with a single textbox

Rickyboy320:
## Description
Provide a default format upon assignment creation (#169)

- More verbose error message in the store,
- Fixes assignment creation permission issues.

## Summary of changes
- Fixes #169


src/vue/src/store/modules/permissions.js
```diff
@@ -12,7 +12,7 @@ const getters = {
         const routeParams = router.currentRoute.params
 
         if (givenKeyLevel && id) {
-            if (permissions[givenKeyLevel + id] === undefined) { throw Error('Permission input error, no match for the given custom key this should never occur!') }
+            if (permissions[givenKeyLevel + id] === undefined) { throw Error('Permission input error, no match for the given key:' + givenKeyLevel + id + '.') }
```
Larspolo: Does this work? or do you need `+ (givenKeyLevel + id) +`?

engelhamer: Don't know about that but I do notice now that there should be a space after ':' in the error.

Rickyboy320: Yeah works fine, givenKeyLevel is a string.

PR #328, opened 2018-09-15
Feature/easy developer feedback

### LINKED ISSUE #275: Easy developer feedback
# Feature request

## Description
In order to improve the application users should be able to contact the developers via email in a quick and easy way. Follow up emails can be sent directly to the email address stored in the user's account.

## User story
As a user I want to be able to send technical feedback to the developers in a simple way. Technical details like the browser or specific page should be attached automatically.

## User interface
There shall be an easy to find button on every page that prompts a user to type some text and then immediately send an email to the developer team.

cloudzi1:
## Description
Fixed the easy developer feedback issue. Added a page where a form can be filled in with multiple attached files possible. After someone presses send an email will be send to the developers which is for now ejourn.al.trial@gmail.com. Only users with an account can send feedback.

## Summary of changes
- Fixes #275

### DIFF
--- a/src/django/VLE/utils/email_handling.py
+++ b/src/django/VLE/utils/email_handling.py
@@ -2,6 +2,10 @@ from django.conf import settings
 from django.contrib.auth.tokens import PasswordResetTokenGenerator
 from django.core.mail import EmailMessage
 
+from VLE.models import Role
+
+DEV_EMAIL = 'zllong.zhu@gmail.com'
+
 
 def send_email_verification_link(user):
     """Sends an email verification link to the users email adress."""
@@ -35,3 +39,17 @@ If you did make the request please visit the link below and set a new password:
 '''.format(recovery_link=recovery_link)
 
     EmailMessage('eJourn.al password recovery', email_body, to=[user.email]).send()
+
+
+def send_email_feedback(user, topic, type, feedback, browser):
+    """Sends the feedback of an user to the developers."""
+    subject = "[Feedback] {}".format(topic)
+    body = "TYPE: {}\n\n".format(type)
+    body += "FEEDBACK BY: {}\n".format(user.username)
+    body += "EMAIL: {}\n".format("" if user.email is None else user.email)
+    body += "TEACHER: {}\n".format(user.is_teacher)
+    body += "ROLES: {}\n".format(Role.objects.filter(role__user=user).values('name'))
+    body += "USERAGENT: {}\n\n".format(browser)
+    body += "THE FEEDBACK:\n{}".format(feedback)
+
+    EmailMessage(subject, body, to=[DEV_EMAIL]).send()
--- a/src/django/VLE/utils/email_handling.py
+++ b/src/django/VLE/utils/email_handling.py
@@ -2,9 +2,8 @@ from django.conf import settings
 from django.contrib.auth.tokens import PasswordResetTokenGenerator
 from django.core.mail import EmailMessage
 
+from VLE.settings.development import EMAIL_HOST_USER
 from VLE.models import Role
-
-DEV_EMAIL = 'zllong.zhu@gmail.com'
 
 
 def send_email_verification_link(user):
@@ -56,4 +55,4 @@ def send_email_feedback(user, files, topic, ftype, feedback, user_agent):
     for file in files:
         attachs.append((file.name, file.read(), file.content_type))
 
-    EmailMessage(subject, body, attachments=attachs, to=[DEV_EMAIL]).send()
+    EmailMessage(subject, body, attachments=attachs, to=[EMAIL_HOST_USER]).send()
--- a/src/django/VLE/utils/email_handling.py
+++ b/src/django/VLE/utils/email_handling.py
@@ -41,15 +41,19 @@ If you did make the request please visit the link below and set a new password:
     EmailMessage('eJourn.al password recovery', email_body, to=[user.email]).send()
 
 
-def send_email_feedback(user, topic, type, feedback, browser):
+def send_email_feedback(user, files, topic, ftype, feedback, user_agent):
     """Sends the feedback of an user to the developers."""
-    subject = "[Feedback] {}".format(topic)
-    body = "TYPE: {}\n\n".format(type)
+    subject = "[Feedback] {}".format(topic[0])
+    body = "TYPE: {}\n\n".format(ftype[0])
     body += "FEEDBACK BY: {}\n".format(user.username)
     body += "EMAIL: {}\n".format("" if user.email is None else user.email)
     body += "TEACHER: {}\n".format(user.is_teacher)
     body += "ROLES: {}\n".format(Role.objects.filter(role__user=user).values('name'))
-    body += "USERAGENT: {}\n\n".format(browser)
-    body += "THE FEEDBACK:\n{}".format(feedback)
+    body += "USER-AGENT: {}\n\n".format(user_agent[0])
+    body += "THE FEEDBACK:\n{}".format(feedback[0])
 
-    EmailMessage(subject, body, to=[DEV_EMAIL]).send()
+    attachs = []
+    for file in files:
+        attachs.append((file.name, file.read(), file.content_type))
+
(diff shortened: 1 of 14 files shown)


engelhamer:
I'll move the feedback button to a more appropriate position. Especially on mobile devices this looks very peculiar.

feedbacu

src/django/VLE/utils/email_handling.py
```diff
@@ -35,3 +38,21 @@ def send_password_recovery_link(user):
...
+
+
+def send_email_feedback(user, files, topic, ftype, feedback, user_agent):
+    """Sends the feedback of an user to the developers."""
+    subject = "[Feedback] {}".format(topic[0])
+    body = "TYPE: {}\n\n".format(ftype[0])
+    body += "FEEDBACK BY: {}\n".format(user.username)
+    body += "EMAIL: {}\n".format("" if user.email is None else user.email)
+    body += "TEACHER: {}\n".format(user.is_teacher)
+    body += "ROLES: {}\n".format(Role.objects.filter(role__user=user).values('name'))
```
engelhamer: This does not put the actual permissions in the email, which might be more helpful.

src/django/VLE/utils/email_handling.py
```diff
@@ -35,3 +38,21 @@ def send_password_recovery_link(user):
...
+    body += "USER-AGENT: {}\n\n".format(user_agent[0])
+    body += "THE FEEDBACK:\n{}".format(feedback[0])
+
+    attachs = []
```
engelhamer: Pretty weird variable name. Why not just finish it?

src/django/VLE/utils/email_handling.py
```diff
@@ -35,3 +38,21 @@ def send_password_recovery_link(user):
...
+    for file in files:
+        attachs.append((file.name, file.read(), file.content_type))
+
+    EmailMessage(subject, body, attachments=attachs, to=[EMAIL_HOST_USER]).send()
```
engelhamer: We now only send an email to the developers, but maybe it is better to also send an email to the user itself. This should not contain the ALL CAPS: fields, but simply their written message as well as some confirmation that we have received their feedback. It is also possible to include the name of the user in this.

We could then reply to that email in case we want to get in touch with the user, and the conversation as a whole will be more organized.

src/django/VLE/views/email.py
```diff
@@ -125,3 +132,38 @@ def request_email_verification(request):
...
+@api_view(['POST'])
+def send_feedback(request):
+    """Send an email with feedback to the developers.
+
+    Arguments:
+    request -- the request that was sent
+        topic -- the topic of the feedback
+        type -- the type of feedback
+        feedback -- the actual feedback
+        browser -- the browser of the user who sends the feedback.
```
engelhamer: Why do these last two arguments end with a period but the previous do not?

src/django/VLE/views/email.py
```diff
@@ -125,3 +132,38 @@ def request_email_verification(request):
...
+        On failure:
+            bad request -- when required keys are missing or file sizes too big.
+            unauthorized -- when the user is not logged in
+        On success:
+            success -- with a description
+    """
+    if not request.user.is_authenticated:
+        return response.unauthorized()
+
+    if not all(x in request.POST for x in ['topic', 'feedback', 'ftype', 'user_agent']):
```
engelhamer: Don't we have a function (check params) for this?

src/vue/src/Header.vue
```diff
@@ -20,6 +20,7 @@
         </b-collapse>
 
         <b-navbar-nav class="ml-auto">
+            <b-nav-item :to="{ name : 'Feedback' }">Feedback</b-nav-item>
```
engelhamer: This does not look very good.

src/vue/src/api/feedback.js
```diff
@@ -0,0 +1,8 @@
+import auth from '@/api/auth'
+
+export default {
+
```
engelhamer: Useless whitespace.

src/vue/src/views/Feedback.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+    <content-single-column>
+        <bread-crumb>&nbsp;</bread-crumb>
+        <b-card class="no-hover blue-border">
+            <h2>Send the developers your feedback</h2>
+            Hi! If you have any suggestions for improvements or found
```
engelhamer: You could personalize this by adding the user's first name to the greeting.

engelhamer: Additionally, I would replace 'found' with 'encountered'.

src/vue/src/views/Feedback.vue
```diff
@@ -0,0 +1,128 @@
...
+                        :state="Boolean(files)"
+                        class="fileinput"
+                        multiple
+                        ref="fileinput"
+                        @change="filesHandler"
+                        placeholder="Choose a file...">
+                    </b-form-file>
+                </b-col>
+                <b-col>
+                    <b-button class="add-button float-right mt-2" @click="sendFeedback">
```
engelhamer: This should not have a top margin. The send button is now misaligned with the file input. Besides, why put this in a row/column structure? I would suggest getting rid of that and simply follow the style of forms we use throughout the entire application, which means a send button on a separate line (so simply after the file upload).

This currently does not look good on mobile devices.

screenshot from 2018-09-15 16-49-14

src/django/VLE/views/email.py
```diff
@@ -125,3 +132,38 @@ def request_email_verification(request):
...
+        feedback -- the actual feedback
+        browser -- the browser of the user who sends the feedback.
+        files -- potential files as attachments.
+
+        Returns:
+        On failure:
+            bad request -- when required keys are missing or file sizes too big.
+            unauthorized -- when the user is not logged in
+        On success:
+            success -- with a description
```
Larspolo: tab to the left

src/django/VLE/views/email.py
```diff
@@ -125,3 +132,38 @@ def request_email_verification(request):
...
+        return response.bad_request('Required feedback field missing.')
+
+    files = request.FILES.getlist('files')
+    try:
+        validators.validate_email_files(files)
+    except ValidationError:
+        return response.bad_request('The selected files exceeds the total file size limit.')
+
+    email_handling.send_email_feedback(request.user, files, **request.POST)
+    return response.success(description='Feedback received')
```
Larspolo: recieved.

src/vue/src/views/Feedback.vue
```diff
@@ -0,0 +1,128 @@
...
+            } else if (this.feedback === '') {
+                this.$toasted.error('Feedback required')
+            } else {
+                var data = new FormData()
+                data.append('topic', this.topic)
+                data.append('ftype', this.type)
+                data.append('feedback', this.feedback)
+                data.append('user_agent', navigator.userAgent)
+                if (this.files != null) {
+                    for (var i = 0; i < this.files.length; i++) {
```
Larspolo: `var i in this.files`

src/vue/src/views/Feedback.vue
```diff
@@ -0,0 +1,128 @@
...
+        'text-editor': textEditor,
+        icon
+    },
+    methods: {
+        filesHandler (e) {
+            let files = e.target.files
+            if (files.length < 1) { return }
+
+            var uploadSize = 0
+            for (var i = 0; i < files.length; i++) {
```
Larspolo: `var i in files`

src/django/VLE/views/email.py
engelhamer: Nope, received is correct.

src/vue/src/api/feedback.js
cloudzi1: Most api files in vue have this whitespace. I just followed that

engelhamer:
Placed a button at the lower right corner of every screen, and moved the feedback form to a modal that can be launched from there.

screenshot from 2018-09-15 18-42-25
screenshot from 2018-09-15 18-44-01

engelhamer:
@cloudzi1 maybe you can change the reply address? Spoofing the sender doesn't sound very neat anyway.

PR #337, opened 2018-09-16
EntryFields dedicated component

mjvkeulen:
## Description
Code duplication removed of all entry field type checks. Some bug fixes

## Summary of changes
- Entry nodes now correctly swap components if required (no more lazy loading).
- Teacher can once again download user files from the journal view


src/vue/src/components/template/TemplatePreview.vue
```diff
@@ -26,7 +26,7 @@
                 :aID="$route.params.aID"
             />
             <b-input
-                v-else-if="field.type=='v'"
+                v-else-if="field.type== 'v'"
```
Larspolo: Then also add a space before the `==`?

engelhamer:
Why did you close this @mjvkeulen?

PR #338, opened 2018-09-16
Bugfix/monkeytesting

### LINKED ISSUE #325: Confirmation for unsaved changes
# Feature request
A confirmation window for every edit location, if you try to leave the page while there are still unsave changes available.

## Description
A confirmation window should pop-up when you try to leave a page when there is still unsaved data, while editing the information.

## User story
As a user I want a confirmation before going to another page if there are unsaved changes.

### LINKED ISSUE #333: Edit mode stays on when switching entries
# Bug report
When in edit mode for an editable entry and you switch to a different entry is stays on edit mode. This also works when switching to an entry that is not editable anymore and saving will still update the noneditable entry.

## Expected Behavior
An entry should only be in edit mode when the edit button of that entry is clicked.

## Actual Behavior
All entries will switch to edit mode if you switch to edit mode in one of the entries and then click on another entry.

1. Go to an editable entry.
2. Go into edit mode.
3. Click on any other entry (even noneditable ones) and it will be in edit mode and be editable.

Archifer:
## Description
A lot of bugfixxes and enhancements. 

Added the enhancement that for every location where you can edit if you have unsaved changes and you try to leave the page, entry or something that a confirmation wil be asked. Locations of implementation: Course-edit, Assignment-edit, Permission-edit, entry-edit and entry-grading.

The bug of non updating rich texteditors when switching between entries in edit modus is now fixt and will update properly. 

The bug of staying in edit mode when switching between entries is now solved by forcing the user in the view mode instead of possibly staying in edit mode.

The bug of the undo changes button not working properly is now fixt in the rolconfirmation window.

## Summary of changes
- Fixes #325 
- Fixes #333


engelhamer:
#333 actually also requires an added back end check, which I don't see implemented in this pull request. Please make sure to add that as well.

engelhamer:
Good!

PR #340, opened 2018-09-16
Feature/lti many to one

BigDGW:
## Description
With this pull request it will be made possible for multiple courses or assignment on the VLE to be coupled with one course or assignment eJournal

## Summary of changes
- New model for the one to many relationship between course/assignment and lti_id
- Rewrite the get by lti_id and update lti_id for the new system


src/django/VLE/lti_launch.py
```diff
@@ -113,10 +114,10 @@ def create_lti_query_link(names, values):
 def check_course_lti(request, user, role):
     """Check is an course with the lti_id exists"""
     course_id = request['custom_course_id']
-    courses = Course.objects.filter(lti_id=course_id)
+    courses = Lti_ids.objects.filter(lti_id=course_id, for_model='Course')
```
Rickyboy320: Shouldn't this be named something like 'lti_ids = ...'

src/django/VLE/lti_launch.py
```diff
@@ -126,9 +127,9 @@ def check_course_lti(request, user, role):
 def check_assignment_lti(request):
     """Check is an assignment with the lti_id exists"""
     assign_id = request['custom_assignment_id']
-    assignments = Assignment.objects.filter(lti_id=assign_id)
+    assignments = Lti_ids.objects.filter(lti_id=assign_id, for_model='Assignment')
```
Rickyboy320: Same here, assignments[0].assignment below is a little weird, lti_ids = ... should make more sense.

src/django/VLE/models.py
```diff
@@ -668,3 +659,35 @@ class Comment(models.Model):
...
+    )
+
+    course = models.ForeignKey(
+        'Course',
+        on_delete=models.CASCADE,
+        null=True
+    )
+
+    lti_id = models.TextField()
+    for_model = models.TextField()
```
Rickyboy320: Maybe a choice field would be nicer here?

src/django/VLE/views/assignment.py
```diff
@@ -69,7 +69,10 @@ def list(self, request):
         else:
             return self.upcoming()
 
-        return response.success({'assignments': serializer.data})
+        data = serializer.data
+        for i, assignment in enumerate(data):
+            data[i]['lti_couples'] = len(Lti_ids.objects.filter(assignment=assignment['id']))
```
Rickyboy320: Maybe embed this in the serializer? (Could be that it's not possible, I'm still not well versed on the serializers)

src/django/VLE/views/course.py
```diff
@@ -200,10 +204,12 @@ def linkable(self, request):
...
 
         unlinked_courses = Course.objects.filter(participation__user=request.user.id,
-                                                 participation__role__can_edit_course_details=True,
-                                                 lti_id=None)
+                                                 participation__role__can_edit_course_details=True)
         serializer = serialize.CourseSerializer(unlinked_courses, many=True)
-        return response.success({'courses': serializer.data})
+        data = serializer.data
+        for i, course in enumerate(data):
+            data[i]['lti_couples'] = len(Lti_ids.objects.filter(course=course['id']))
```
Rickyboy320: See above comment on serializer

PR #360, opened 2018-09-18
Bugfix/todo redirect

JowehL:
## Description
Fixed permission checks for redirects and spelling errors


src/vue/src/components/assets/TodoCard.vue
```diff
@@ -40,13 +40,13 @@ export default {
             delta -= minutes * 60
 
             if (days) {
-                return days + ' days(s)'
+                return days + ' day(s)'
```
Larspolo: This can be even further split into "day" and "days" with an `days > 1 ? days + 'days' : '1 day'`

PR #361, opened 2018-09-18
Use npm ci to force same versions for everything

Rickyboy320:
## Description
npm install always forced the package-lock.json file to reconfigure. Because of this that file was worthless, because it was never even used. npm ci builds from the package-lock.json file instead of the package.json, forcing the exact same dependencies on all machines. This is useful, because then we use the exact same versions of packages that the eventual production will also use. Any updating of packages will also be reflected across the board, instead of npm 'guessing' what version would be most suitable. (Generally newest, but as we have seen, the package-lock was updated constantly). In favor of more stable builds, the package-lock should be used in this manner.

(Thanks to Hendrik for pointing this out)

I do have to note that since the package-lock.json is still not synchronized between branches, any postive effects are not immediately visible. Once everyone/all branches use the updated makefile (that is, uses npm ci instead of npm install), we can finally lock the dependencies in place.

## Summary of changes
- npm ci instead of npm install
- Added update-dependencies to fullfill the actual role of 'updating'; This only performs 'minor' updates (bugfix releases), as a full update is never safe (breaking changes).
- I have looked into properly updating our dependencies (we are severely outdated with some of our core dependencies), but this is an epic on its own. I do not suggest doing this anytime soon, as the benefit will be minimal.
- Fixed a deploy issue with a file trying to access development settings.


engelhamer:
Accordding to [CVE-2018-14041](https://nvd.nist.gov/vuln/detail/CVE-2018-14041) there is a security vulnerability in our current Bootstrap version. Maybe this is a good moment to update?

Rickyboy320:
> According to [CVE-2018-14041](https://nvd.nist.gov/vuln/detail/CVE-2018-14041) there is a (moderate) security vulnerability in our current Bootstrap version. Maybe this is a good moment to update?

Good catch, updated bootstrap as well.

PR #362, opened 2018-09-18
Added LTI badge

### LINKED ISSUE #357: Visual indication that an assignment/course is linked via LTI
# Feature request

## Description
Add a visual indication that an assignment/course is linked via LTI.
image
A link icon on the top left of each linked item would be nice

engelhamer:
## Description
Added a badge stating that an assignment has been linked via LTI.

lti

## Summary of changes
- Fixes #357


engelhamer:
I do not think that this would make much sense for courses, since there are
cases in which they are not linked via LTI but some assignments are.

>

PR #367, opened 2018-09-18
TextEditor: listen to givenContent for content updates. 

### LINKED ISSUE #344: Description of assignment not loading
# Bug report

## Expected Behavior
When going to the AssignmentEdit page, you want to see the description of the assignment.

## Actual Behavior
The description of the assignment is empty, only appears after a reload/move back and forth from 2 nodes.

1. login
2. https://ejourn.al/Home/Course/5/Assignment/5/Format

Rickyboy320:
## Description
Listen to givenInput on the text editor. This is necessary because the data is not always present upon loading the text-editor; it is loaded with a slight delay. The loading succeeds correctly in this manner, but one possible concern here could be the altering of the data after the user has already typed something. While this is unlikely to happen, it could introduce another bug. Maybe some function could be added that updates the text-editor with a new value, or a boolean could be tracked to see if the user already typed something, and then do not replace. Regardless, I know too little about vue so I need some input here.

## Summary of changes
- Fixes #344


libre-man:
You are kind of reinventing the wheel here as you're recreating the [`v-model` directive](https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components). I think makes way more sense that you shouldn't load a component if the initial data of that component is not present, as this could be very confusing for a user.

Rickyboy320:
This has been updated, v-models are not possible since the editor does apparently not work that way.
@Airgetfrog

Airgetfrog:
I wonder why we didn't use [this](https://github.com/tinymce/tinymce-vue)...

Airgetfrog:
@mjvkeulen above

engelhamer:
So are we going to implement the Vue version or merge this pull request?

Airgetfrog:
Was there a particular purpose to creating our own wrapper or would it make sense to replace it with the official Vue wrapper in that repo? There are apparently more issues found by Xav that could be solved by using it.

mjvkeulen:
I will try it out today, the official component also receives constant updates so if I can't figure out why I decided against it (early July?) I will swiftly make a switch.

Rickyboy320:
Using the component will probably fix the issue, as then the content can be bound to the content field, after which vue can automatically update it all.

Definitely think it’s worth trying.

engelhamer:
The TinyMCE Vue component might work out better over the long haul (#524). For now this will do.

PR #371, opened 2018-09-19
Make sure grade is between 0 and 10

### LINKED ISSUE #365: A negative grade is a valid grade
# Bug report

## Expected Behavior
Negative grades should be rejected (both front-sided and back-sided)

## Actual Behavior
They are happily accepted, and work properly with all calculations.

1. Attempt to edit someone's grade
2. Input a negative grade
3. Save

This may be an intended feature (to deduct points for whatever reason), but it looks unintended (as confirmed by Engel).

libre-man:
## Description
Make sure that the grade is always between 0 and 10 inclusive.

## Summary of changes
- Fixes #365


libre-man:
Oh wait, we need another check first, one second.

engelhamer:
The grade for an indiviual entry does not need limitation. Only the total grade of a journal (which can be computed based on the amount of possible points for the assignment) should be capped at 10.

It is a good Idea to prevent giving entries a grade lower than zero though!

src/django/VLE/views/entry.py
```diff
@@ -149,6 +149,8 @@ def partial_update(self, request, *args, **kwargs):
         if grade and \
            not permissions.has_assignment_permission(request.user, journal.assignment, 'can_grade'):
             return response.forbidden('You cannot grade or publish entries.')
+        elif isinstance(grade, (int, float)) and grade < 0:
+            return response.bad_request('Grade must be between zero and ten.')
```
Rickyboy320: This should now become 'Grade should be positive'

PR #373, opened 2018-09-20
Implemented group filter on assignment page and renamed functions

### LINKED ISSUE #332: Filter by groups on the assignment page
# Feature request

## Description
Users who have rights to grade journals should be able to filter by groups on the assignment.vue page. This should make it easier for teachers/TA's to search for students.

## User story
As a teacher/TA(or people with grading rights), I want to be able to filter by groups on the assignment page.

JowehL:
## Description
Implemented filter functionality on the assignment page to filter journals by group.

## Summary of changes
- Fixes #332

### DIFF
--- a/src/django/VLE/serializers.py
+++ b/src/django/VLE/serializers.py
@@ -122,8 +122,7 @@ class AssignmentSerializer(serializers.ModelSerializer):
     def get_journals(self, assignment):
         if 'journals' in self.context and self.context['journals']:
             return JournalSerializer(
-                Journal.objects.filter(
-                assignment=assignment),
+                Journal.objects.filter(assignment=assignment),
                 many=True,
                 context=self.context).data
         else:
--- a/src/django/VLE/serializers.py
+++ b/src/django/VLE/serializers.py
@@ -121,10 +121,11 @@ class AssignmentSerializer(serializers.ModelSerializer):
 
     def get_journals(self, assignment):
         if 'journals' in self.context and self.context['journals']:
-            return JournalSerializer(Journal.objects.filter(
-                   assignment=assignment),
-                   many=True,
-                   context=self.context).data
+            return JournalSerializer(
+                Journal.objects.filter(
+                assignment=assignment),
+                many=True,
+                context=self.context).data
         else:
             return None
 
--- a/src/django/VLE/serializers.py
+++ b/src/django/VLE/serializers.py
@@ -122,9 +122,9 @@ class AssignmentSerializer(serializers.ModelSerializer):
     def get_journals(self, assignment):
         if 'journals' in self.context and self.context['journals']:
             return JournalSerializer(Journal.objects.filter(
-                    assignment=assignment),
-                    many=True,
-                    context=self.context).data
+                   assignment=assignment),
+                   many=True,
+                   context=self.context).data
         else:
             return None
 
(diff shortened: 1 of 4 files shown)


src/vue/src/views/CourseEdit.vue
```diff
@@ -378,7 +378,7 @@ export default {
                 viewList = viewList.sort(compareUsername)
             }
 
-            return viewList.filter(searchFilter).filter(checkGroup)
+            return viewList.filter(searchFilter).filter(groupFIlter)
```
Larspolo: *groupFilter

engelhamer:
Can you please add a UI screenshot? @JowehL

JowehL:
journals

groupfilter

@engelhamer @Larspolo

src/django/VLE/serializers.py
```diff
@@ -121,7 +121,10 @@ def get_journal(self, assignment):
 
     def get_journals(self, assignment):
         if 'journals' in self.context and self.context['journals']:
-            return JournalSerializer(Journal.objects.filter(assignment=assignment), many=True).data
+            return JournalSerializer(Journal.objects.filter(
+                    assignment=assignment),
```
engelhamer: You went one space too far.

src/django/VLE/serializers.py
```diff
@@ -121,7 +121,10 @@ def get_journal(self, assignment):
...
+                    many=True,
```
engelhamer: And again.

src/django/VLE/serializers.py
```diff
@@ -121,7 +121,10 @@ def get_journal(self, assignment):
...
+                    context=self.context).data
```
engelhamer: And again.

Larspolo: I would also suggest changing the styling here. This way it seems that assignment is a parameter of JournalSerializer, but its not, its from the query. Maybe change it to something like this:
```
return JournalSerializer(
    Journal.objects.filter(assignment=assignment),
    many=True,
    context=self.context).data
```

JowehL:
screenshot from 2018-09-20 11-47-40

src/vue/src/components/assignment/StudentCard.vue
```diff
@@ -6,7 +6,8 @@
             </b-col>
             <b-col order="2" cols="9" class="pl-3">
                 <todo-square v-if="numMarkingNeeded > 0 && !hideTodo" class="float-right" :num="numMarkingNeeded"/>
-                <b>{{ student.first_name + ' ' + student.last_name }}</b><br/>
+                <template v-if="student.group"><b>{{ student.first_name + ' ' + student.last_name }}</b> ({{ student.group }})<br/></template>
```
engelhamer: This seems ineffecient andere inappropriate. Please reduce these two lines to just one, nu wrapping the group part in a span and applying a v-if only there.

src/vue/src/views/Assignment.vue
```diff
@@ -5,6 +5,13 @@
         <b-card slot="main-content-column" class="no-hover settings-card">
             <input class="theme-input full-width multi-form" type="text" v-model="searchVariable" placeholder="Search..."/>
             <div class="d-flex">
+                <b-form-select class="multi-form mr-2" v-model="selectedFilterGroupOption"
+                               :select-size="1">
+                    <option :value="null">Filter group by ...</option>
```
engelhamer: What is the point of binding null?

JowehL: You need a value to get back all the journals when you don't want to filter on groups anymore

engelhamer: Ah I see.

PR #385, opened 2018-09-26
Bugfix/journal scroll down

Jeroentetje3:
## Description
Added a scroll to top function in order to negate the problem of automatically scrolling down in a journal.
This isn't the complete fix for the problem that rich text editors (tinyMCE) autofocus.
## Summary of changes
- Fixes Auto scroll down of journal


src/vue/src/components/entry/EntryFields.vue
```diff
@@ -126,6 +126,9 @@ export default {
         'file-download-button': fileDownloadButton,
         'image-file-display': imageFileDisplay
     },
+    created () {
```
libre-man: This function is called before you component is added to the DOM, are you sure that is what you want?

engelhamer:
I have found a more appropriate solution and will share it shortly. Closing this now.

PR #389, opened 2018-09-26
Bugfix image shrink

### LINKED ISSUE #387: Email verification token link has 1 '/' too many because of the settings.BASELINK has an extra /
# Bug report

## Expected Behavior
Go to profile to verify email, get a mail, click on the link to verify, no 404 error and everything works
## Actual Behavior
404 error when trying to verify email

Go to profile to verify email, get a mail, click on the link to verify, 404 error

Mail:
image

Archifer:
## Description
Fixed the bug from hiding an picture again that would leave to much white space (removed the unneeded css from this since it uses a transition now)

## Summary of changes
- Fixes #386


src/django/VLE/utils/email_handling.py
```diff
@@ -10,7 +10,7 @@ def send_email_verification_link(user):
     token_generator = PasswordResetTokenGenerator()
     token = token_generator.make_token(user)
 
-    recovery_link = '%s/EmailVerification/%s' % (settings.BASELINK, token)
+    recovery_link = '%sEmailVerification/%s' % (settings.BASELINK, token)
```
Larspolo: This is not yet the full implementation I hoped. I would like for it to check whether there is a / at the end of baselink, and if not add a /. Also there is the same problem in passwordrecovery, lti_grade_passback and lti_launch that isnt fixed, and possible others as well, but not sure

engelhamer: Then the issue should be elaborated as well.

Larspolo: Realized it only after I made the issue, improved it now

PR #393, opened 2018-09-26
FAB's, better input and no more confusing points

### LINKED ISSUE #272: Shortcut for entry creation on mobile devices
# Feature request

## Description
On mobile devices the timeline is always collapsed and needs to be opened by clicking a button before a user has acces to the add node. In order to make adding an entry easier and quicker we should add a shortcut.

## User story
As a student I want to be able to add a new entry to a journal with just one click on a mobile device.

## User interface
We will have to add floating add button (like in material design) to the bottom right of the screen on mobile devices. Toasts should be shifted up accordingly.

### LINKED ISSUE #327: Merge parallel point goals
# Feature request

## Description
There currently exist two types of point limitations for assignments. One called 'points_possible' (stored in the Assignment, gathered through LTI) and another called 'max_points' (stored in the Format and used for computing progress). These two should be merged, ideally by getting rid of max_points and only using points_possible.

### LINKED ISSUE #343: Redesign of the add node and save buttons in the assignment edit page
# Feature request

## Description
Redesign of the add node and save buttons in the assignment edit page

## User story
As a teacher you want to add a node by pressing a button in the timeline, not at the bottom in the main-content part

## User interface
image
underneath this image, add an add node that adds a new node

image
Change this save format button to a FAB that only appears when there are changes to be saved

### LINKED ISSUE #345: Redesign of the "stackable templates" feature
# Feature request

## Description
Redesign of the "stackable templates" feature. The current implementation is unclear.

## User story
 Currently to enable a student to submit multiple versions of a template, as teacher you have to put the templated on "enabled" status.

Suggested improvement:
Add another node to the timeline where you can select the "stackable templates"

## User interface
Screenshot or description of user interface changes
image
These "enabled" buttons are unclear

### LINKED ISSUE #350: Automatically scrolls down on large entries
# Bug report

## Expected Behavior
To not scroll down when viewing an entry

## Actual Behavior
Scrolls down when viewing an entry (caused by the autofocus on the rich text editor)

1. go to an entry that is larger then your screen

### LINKED ISSUE #397: Due date not used for deadline(due date) note.
# Bug report

## Expected Behavior
default deadline should default to the given due date if unset. set on the same date between unlock and lock date

## Actual Behavior
default deadline outside assignment period

create assignment through lti

engelhamer:
## Description
Added FAB's for saving a format and adding a new entry to a journal, replaced the peach color with a more vibrant yellow and fixed some issues regarding input fields.

home

formatedit

journal

## Summary of changes
- Renamed 'edag' to timeline
- Fixed #272 
- Fixed #327 
- Fixed #343 
- Fixed #350 
- Fixed #397


src/django/test/test_get_apis.py
```diff
@@ -331,7 +331,7 @@ def test_get_comments(self):
     #     """Test get assignment by lti id function."""
     #     course = factory.make_course('Portfolio', 'PAV', author=self.rein)
     #     template = factory.make_entry_template('template')
-    #     format = factory.make_format([template], 10)
+    #     format = factory.make_format([template])
     #     factory.make_assignment('Colloq', 'description1', format=format, courses=[course], lti_id='12xy')
     #
     #     login = test.logging_in(self, self.rein_user, self.rein_pass)
```
Larspolo: Can this test be removed or implemented again?

engelhamer: What test? I have only updated the make format call since it now doesn't require this second argument.

engelhamer:
Input inconsistencies have been fixed, but I do not understand what is wrong with the rich text editor.

src/django/test/test_get_apis.py
Larspolo: True, but I only now see this test being commented out, maybe you can uncomment it, see if it works?

engelhamer: Ah, I didn't even notice that this part was disabled 😅

Larspolo:
The save button in now a little inconsistent with the other pages like CourseEdit (kinda understandable) and UserRoleConfiguration (would be nice if this was the same)

engelhamer:
In what sense? FAB's never contain text.

src/django/test/test_get_apis.py
libre-man: But why is are they disabled? Are they not useful in any way?

Archifer:
The FAB save button on the format editing might be a bit confusing for an user (for me and Okke atleast, since just the icon can be a bit confusing  about what functionality it has when you are not familiar with the environment) that it is the save button for the format. Maybe change the icon for the word 'save' (or add the word). I am not extremely sure what would be a nice way to repressent it, but for now I feel it would be confusing for users.

src/vue/src/components/journal/JournalStartCard.vue
```diff
@@ -1,11 +1,11 @@
...
     <b-card class="no-hover" :class="$root.getBorderClass($route.params.cID)">
         <h2>{{ assignment.name }}</h2>
         <div v-html="assignment.description"/>
-        <h6 v-if="assignment.unlock_date">Unlock date</h6>
-        <p v-if="assignment.unlock_date">{{ $root.beautifyDate(assignment.unlock_date) }}</p>
-        <b v-if="student && assignment.due_date && new Date(assignment.due_date) < new Date()">The deadline for this assignment has passed.</b>
-        <b v-if="assignment.unlock_date && new Date(assignment.unlock_date) > new Date()">This assignment is locked and will be made available later.</b>
+        <h2 class="field-heading" v-if="assignment.unlock_date">Unlock date</h2>
+        <span v-if="assignment.unlock_date">{{ $root.beautifyDate(assignment.unlock_date) }}</span>
+        <h2 class="field-heading" v-if="assignment.unlock_date && new Date(assignment.unlock_date) > new Date()">This assignment is locked and will be made available later.</h2>
```
mjvkeulen: Lock date is not private, as such displaying it here would be more intuitive

PR #394, opened 2018-09-27
Feature/auto filter

### LINKED ISSUE #351: Automatically filter on group as TA
# Feature request

## Description
If a TA is linked to a group, auto filter on that group.

## User story
Go to an assignment as TA to view only your students.

JowehL:
## Description
Changed permission needed to see the groups of a course.
Implemented group auto filter functionality on the assignment page. 
Journals are also filtered on students only(Should this later be more than only students?)

## Summary of changes
- Fixes #351


src/vue/src/views/Assignment.vue
```diff
@@ -226,6 +237,11 @@ export default {
                 return true
             }
 
+            /* TODO should work on users with no view journal permissions. */
+            function studentFilter (assignment) {
+                return assignment.student.role === 'Student'
```
Airgetfrog: This should filter on the role having `can_have_journal`, due to custom roles possibly having journals.

src/django/VLE/views/participation.py
```diff
@@ -49,6 +49,35 @@ def list(self, request):
         users = UserSerializer(course.users, context={'course': course}, many=True).data
         return response.success({'participants': users})
 
+    def retrieve(self, request, pk=None):
```
Airgetfrog: I don't know much about RESTful api, but shouldn't retrieve be a getter for a specific participation, or does that not matter?

Larspolo: Agreed, however these calls are more about the participation iso the participant. Thats why we just use the UserSerializer iso a PerticipationSerializer. It however is a nice find and should be changed to a ParticipationSerializer. @JowehL will you do that in this PR? Is not, create an issue and someone will fix it later on

JowehL: @Larspolo At the moment I have implemented a ParticipationSerializer for only the retrieve. Do you want all the UserSerializers changed to ParticipationSerializers or only the retrieve.

src/django/VLE/serializers.py
```diff
@@ -14,12 +14,13 @@
...
     name = serializers.SerializerMethodField()
     role = serializers.SerializerMethodField()
+    can_have_journal_permission = serializers.SerializerMethodField()
     group = serializers.SerializerMethodField()
 
     class Meta:
         model = User
         fields = ('username', 'first_name', 'last_name', 'name', 'profile_picture', 'is_teacher', 'lti_id', 'id',
-                  'role', 'group')
+                  'role', 'can_have_journal_permission', 'group')
```
Airgetfrog: I feel like this is something that should be done on the frontend? Or at least not in the userserializer. But I suppose it should work and I'm not sure if I'm correct anyway.

src/django/VLE/views/group.py
```diff
@@ -47,9 +47,12 @@ def list(self, request):
 
         role = permissions.get_role(request.user, course)
         if role is None:
-            return response.forbidden('You are not in this course.')
-        if not role.can_view_course_users:
-            return response.forbidden('You are not allowed to view the groups.')
+            return response.forbidden('You are not a participant of this course.')
+        if not (role.can_view_course_users or
```
Airgetfrog: Can't this be simplified in the new permission structure? Not sure.

JowehL: Maarten changed it to this in the develop branch

src/django/VLE/serializers.py
JowehL: I have no idea how to create a good function to check for permissions in the frontend, so this could be a temporary fix in the backend

engelhamer: I don't think this is a good idea. You can simply use the `$hasPermission` function to check this on the front-end.

src/django/VLE/serializers.py
```diff
@@ -31,6 +32,12 @@ def get_role(self, user):
 
         return permissions.get_role(user, self.context['course']).name
 
+    def get_can_have_journal_permission(self, user):
```
engelhamer: The back-end codebase would turn into a total mess if we were to do this for all permissions.

src/django/VLE/serializers.py
Airgetfrog: `$hasPermission` works on the current user, what we need here is a way to check the permission for a specific role.

engelhamer: Isn't it a better idea to check (front-end) if there exists a journal for that user in the set of assignmentJournals then? That way this would not have to be added to the user serializer, which is more desirable since this is a very particular implementation.

src/django/VLE/views/group.py
```diff
@@ -48,7 +48,10 @@ def list(self, request):
         role = permissions.get_role(request.user, course)
         if role is None:
             return response.forbidden('You are not a participant of this course.')
-        if not (role.can_edit_course_user_group or role.can_add_course_user_group or role.can_delete_course_user_group):
+        if not (role.can_view_course_users or
+                role.can_edit_course_user_group or
+                role.can_add_course_user_group or
+                role.can_delete_course_user_group):
             return response.forbidden('You are not allowed to manage the user groups of this course.')
```
mjvkeulen: As the groups become available for users with the role.can_view_course_users, the response should be updated accordingly.

src/vue/src/views/Assignment.vue
```diff
@@ -109,7 +110,7 @@ export default {
     created () {
         // TODO Should be moved to the breadcrumb, ensuring there is no more natural flow left that can get you to this
         // page without manipulating the url manually. If someone does this, simply let the error be thrown (no checks required)
-        if (!this.$hasPermission('can_view_assignment_journals')) {
+        if (!this.$hasPermission('can_view_assignment_journals', 'assignment', String(this.aID))) {
```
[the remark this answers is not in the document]
JowehL: No it will not work when routing with the breadcrumb with two courses with the same assignment as a TA.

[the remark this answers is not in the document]
engelhamer: Please do not refer to roles but rather to permissions @JowehL.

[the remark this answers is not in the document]
engelhamer: Otherwise the source of the problem (if any) doesn't become clear.

Issue #375, opened 2018-09-20
asynchronous profile pictures


Larspolo:
# Feature request

## Description
The profile pictures need to load asynchronous to improve speed on especially the assignment overview page where there is a large list of students that all load in with their profile picture

Archifer:
What do you mean exactly by loading asynchronous? Do you only mean it for the front-end?, for example show only the first n students with then a button of load n more, or how do you exactly see this enhancement?

engelhamer:
Current alle pictures are loaded along with the page, which introduces a
delay. Loading the pictures asynchronously (independent of the page itself)
this should make the page feel quicker. This does not require any
alterations to the amount of cards shown.

>

Archifer:
Ooh oke I was thinking in another way, but this makes way more sense thanks!

Airgetfrog:
Is this being handled? If not, could someone give a clear overview of the issue (and relevant code)? From what I've seen it doesn't look to be an issue (profile pictures are loaded from url when student cards render, student cards render once filteredjournals exists, filteredjournals only exists when assignmentAPI get returns, i.e. the rest of the page data already exists)?

Larspolo:
![image](https://user-images.githubusercontent.com/11822478/46581231-64526900-ca35-11e8-98cd-f5fd6f520ba8.png)
The profile picture is only a url when its: the default profile picture, or connected to a canvas user. Else you get this b64 string back that takes can easily be half a MB. As a solution I would suggest to store the profilepictures in the user folder iso as a b64 string in the db

Larspolo:
Closed in file rework

PR #399, opened 2018-09-29
Enhancement: Added async loading for the student cards on the assignment page

Archifer:
## Description
On the Assignment page a  query is ran to get all the information of the students of a specific assignment to render all the student cards. This query takes a long time, one of the reasons are already mentioned in issue: #375, but due to this the rest of the page won't be rendered until this query is finished.

## Summary of changes
This enhancement lets the page render the student cards asynchronous so that the rest of the page can be rendered until this query is finished and the student cards can also be rendered.

### DIFF
--- a/src/vue/src/views/Assignment.vue
+++ b/src/vue/src/views/Assignment.vue
@@ -35,20 +35,18 @@
             </b-button>
         </b-card>
 
+        <!--
+            The student-cards will be loaded asynchronous once the
+            heavy querie that loads in all the student data is finished.
+        -->
         <div v-if="filteredJournals" v-for="journal in filteredJournals" :key="journal.student.id" slot="main-content-column">
-            <b-link tag="b-button" :to="{ name: 'Journal',
-                                          params: {
-                                              cID: cID,
-                                              aID: aID,
-                                              jID: journal.id
-                                          }, query: query
-                                        }">
-
-                <student-card
-                    :student="journal.student"
-                    :stats="journal.stats">
-                </student-card>
-            </b-link>
+            <student-Linkcard
+                :cID="cID"
+                :aID="aID"
+                :jID="journal.id"
+                :query="query"
+                :jStudent="journal.student"
+                :jStats="journal.stats"/>
         </div>
         <main-card v-if="assignmentJournals.length === 0" slot="main-content-column" class="no-hover" :line1="'No participants with a journal'"/>
         <main-card v-else-if="filteredJournals.length === 0" slot="main-content-column" class="no-hover" :line1="'No journals found'"/>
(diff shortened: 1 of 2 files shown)


src/vue/src/views/Assignment.vue
```diff
@@ -35,20 +35,18 @@
...
-                                              jID: journal.id
-                                          }, query: query
-                                        }">
-
-                <student-card
-                    :student="journal.student"
-                    :stats="journal.stats">
-                </student-card>
-            </b-link>
+            <student-Linkcard
```
Airgetfrog: `student-linkcard` shouldn't be capitalized.

src/vue/src/views/Assignment.vue
```diff
@@ -99,7 +96,7 @@ export default {
     },
     components: {
         'content-columns': contentColumns,
-        'student-card': studentCard,
+        'student-Linkcard': () => import('@/components/assignment/StudentLinkCard.vue'),
```
Airgetfrog: dont do this

src/vue/src/views/Assignment.vue
```diff
@@ -35,20 +35,18 @@
             </b-button>
         </b-card>
 
+        <!--
+            The student-cards will be loaded asynchronous once the
+            heavy querie that loads in all the student data is finished.
```
Airgetfrog: Asynchronous should be asynchronously, querie should be query.

src/vue/src/views/Assignment.vue
```diff
@@ -35,20 +35,18 @@
...
+        -->
         <div v-if="filteredJournals" v-for="journal in filteredJournals" :key="journal.student.id" slot="main-content-column">
```
Airgetfrog: Div could optionally be merged with component inside.

src/vue/src/views/Assignment.vue
Archifer: Previously you asked where the logic wise change was, it is here. In the description I may have written it a bit unclear, but this let's the component not be loaded until it's called, this is also the reason why I separated the 'student card' in a component. More information about this can be found here https://vuejs.org/v2/guide/components-dynamic-async.html. Since the loading/calculations of this component is rather heavy (calculating everything inside) this method lets the page load all the other components first and then load this components when the calculations are ready. (I am not exactly 100% sure how this works and how to explain it in more depth, but while testing it results in the right behavior)

src/vue/src/views/Assignment.vue
Archifer: Not sure how to reply directly to the general requested changes message, but you asked about "However I'm not too sure what the original issue was. Is it sufficient to make sure the list is only rendered after the rest of the page (if that isn't already done), or should the profile pictures only start loading once the entire list is rendered without pictures (if that isn't already done)? Is someone able to elaborate?" 

What happened was that I was originally trying to fix issue #375 but what I later heard from Lars that the issue was to resolve the problem that the pictures are now being send in the in a b64 string and not the url. What he wanted was that an url in the first query would be send and then in a async call the pictures were retrieved later on. While I thought it was meant front-end only. He agreed that this was a good temporary solution/ small enhancement, but not the fix he wanted. I also previously suggested the re-rendering of the student cards but we thought first that this would't matter that much. Now thinking it over again I believe it is indeed better to re-render once the implementation is in place of first sending the url and then retrieving the profile pictures. So in short I am closing the pull request for adaptation.

PR #401, opened 2018-09-30
Dublicate removed

BigDGW:
## Description
Dublicate line removed


src/django/VLE/lti_launch.py
```diff
@@ -31,13 +31,8 @@ def parse_request(self, request):
         Parses een django request om de method, url header en post data terug
         te geven.
```
Rickyboy320: Should be English.

BigDGW: i didnt change that
so a little late

Issue #402, opened 2018-09-30
Stop using first and last name for users


libre-man:
# Bug report
The names of users are split up into first and last name, however this is not a valid assumption for many people.

## Expected Behavior
There should be only one name field, something like `full_name`.

## Actual Behavior
The name of the user is split up.

engelhamer:
Do you mean that we exclude people with only a first name by this (or special cases like 'van Keulen' as a last name)?

libre-man:
The distinction simply doesn't make sense. There are people without a last/last name, you can read more about it [here](https://www.w3.org/International/questions/qa-personal-names). I don't see the reason for the distinction, so I think it should simply be removed.

PR #404, opened 2018-09-30
User first name, last name and image are overwritten with lti params …

BigDGW:
…when user is linked with a VLE user.

## Description
User meta data is change to lti user meta data when account is koppeld with lti

## Summary of changes
- Fixes user_image lti param changed to custom_user_image (forgotten in automatic setup change).
- Fixes user full and lastname of lti is set as the values if account linked via LTI


src/django/VLE/views/user.py
```diff
@@ -196,12 +196,22 @@ def partial_update(self, request, *args, **kwargs):
...
         else:
-            lti_id, user_image, is_teacher = None, None, False
+            lti_id, user_email, user_full_name, user_image, is_teacher = None, None, None, None, False
         if user_image is not None:
             user.profile_picture = user_image
+        if user_email is not None:
+            user.email = user_email
+            user.verified_email = True
+        if user_full_name is not None:
+            splitname = user_full_name.split(' ')
```
engelhamer: This will have to be changed if we are going to proceed with #402, I must admit it actually seems a bit weird to split the name in this way.

BigDGW: i know but that seemed like a different issue to me then this bug fix

PR #408, opened 2018-10-01
Fixed removed student journal error

### LINKED ISSUE #390: Internal server error on assignment page due to removed student's journals
# Bug report

## Expected Behavior
Deletion of users should not influence availability of assignment page.

## Actual Behavior
As soon as a user who has a journal is removed from a course, the assignment page becomes unusable and will immediately encounter an internal server error. This is because the journal of the removed student still exists, but their role cannot be found since their participation has disappeared.

1. Remove a student (that has a journal) from a course
2. Go to the assignment page

Archifer:
## Description
The error occurred due the fact that on an assignment in a course it would load all the journals, without looking if the journal corresponded to the correct course. We expanded the get journals function so that when a course is given in the context it will only respond with the journals that are in the course, otherwise it will still respond with all the journals for a specific assignment.

## Summary of changes
- Fixes #390

### DIFF
--- a/src/django/VLE/serializers.py
+++ b/src/django/VLE/serializers.py
@@ -139,7 +139,8 @@ class AssignmentSerializer(serializers.ModelSerializer):
             journals = Journal.objects.filter(assignment=assignment)
 
             if 'course' in self.context:
-                journals = [journ for journ in journals if Participation.objects.filter(user=journ.user, course=self.context['course']).count() > 0]
+                journals = [journ for journ in journals
+                            if Participation.objects.filter(user=journ.user, course=self.context['course']).count() > 0]
 
             return JournalSerializer(
                 journals,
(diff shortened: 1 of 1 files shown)


src/django/VLE/serializers.py
```diff
@@ -131,10 +136,16 @@ def get_journal(self, assignment):
 
     def get_journals(self, assignment):
         if 'journals' in self.context and self.context['journals']:
+            journals = Journal.objects.filter(assignment=assignment)
+
+            if 'course' in self.context:
+                journals = [journ for journ in journals if Participation.objects.filter(user=journ.user, course=self.context['course']).count() > 0]
```
engelhamer: This line is too long. Please shorten it tot fit 120 columns.

PR #412, opened 2018-10-02
lti generator, BUGFIX token has no .encode, lti uses correct secret.

BigDGW:
## Summary of changes
- Fixes token has no encode when entering to lti a second time.
- Fixes lti jwt encode uses our jwt secret
- .ADDED lti key secret couple generator

### DIFF
--- a/lti/lti_generator.py
+++ b/lti/lti_generator.py
@@ -1,6 +1,11 @@
 import os
-print('key')
+
+# Generates a new key and secret couple for the lti Oauth communication te key
+# and secrect are generated according to the guidelines given in
+# https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/
+
+print('LTI_KEY:')
 print(os.urandom(16).hex())
 print()
-print('secret')
+print('LTI_SECRET:')
 print(os.urandom(32).hex())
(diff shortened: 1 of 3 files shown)


lti/lti_generator.py
```diff
@@ -0,0 +1,6 @@
+import os
+print('key')
+print(os.urandom(16).hex())
```
libre-man: You can use the builtin `secrets` module to do this.

lti/lti_generator.py
```diff
@@ -0,0 +1,11 @@
+import secrets
+
+# Generates a new key and secret couple for the lti Oauth communication te key
+# and secrect are generated according to the guidelines given in
+# https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/
```
libre-man: This is for oath2 servers, however LTI uses oath 1.

libre-man: Not that it is incorrect or anything, just a bit weird.

BigDGW: The key and secret specifications hasnt changed between Oauth1 and Oauth2. And thus on the offical site the link to the specifications for oauth1 key and secret also links to this link.

PR #417, opened 2018-10-02
Bugfix/deployment issues

Larspolo:
## Description
LTI now get auto deployed, and fixed some deployment issues


settings/database.conf
```diff
@@ -3,4 +3,4 @@ DATABASE_NAME='ejourn'
 DATABASE_HOST='localhost'
 DATABASE_PORT='5432'
 DATABASE_USER='master'
-DATABASE_PASSWORD='django123'
+DATABASE_PASSWORD='django123' # try to exluce '~' from the password, else deploy.sh has to be edited
```
mjvkeulen: I would suggest using a similar style for the config files. My preference is that of the email.conf, where the fields are blank with a suggestion as comment.

lti/lti_setup.xml
```diff
@@ -9,9 +9,9 @@
     http://www.imsglobal.org/xsd/imslticm_v1p0 http://www.imsglobal.org/xsd/lti/ltiv1p0/imslticm_v1p0.xsd
     http://www.imsglobal.org/xsd/imslticp_v1p0 http://www.imsglobal.org/xsd/lti/ltiv1p0/imslticp_v1p0.xsd">
     <blti:title>eJournal</blti:title>
-    <blti:description>Placeholder!!!</blti:description>
-    <blti:icon>https://ejourn.al/static/favicon/favicon-16x16.png</blti:icon>
-    <blti:launch_url>http://127.0.0.1:8000/lti/launch</blti:launch_url>
+    <blti:description>Incremental Long-Term Assignments for Canvas VLE</blti:description>
```
mjvkeulen: for Canvas virtual learning environment -> for THE Canvas VLE?

settings/database.conf
engelhamer: Agreed. Furthermore 'excluce' should be 'exclude'.

lti/lti_setup.xml
engelhamer: Why even mention Canvas? I would suggest leaving it out alltogether.

settings/email.conf
```diff
@@ -1,4 +1,4 @@
 SMTP_HOST='smtp.gmail.com' # example: smtp.gmail.com
 SMTP_LOGIN_MAIL='' # example: ejournal@company.com
-SMTP_LOGIN_PASSWORD='' # password used to login with smtp
+SMTP_LOGIN_PASSWORD='' # password used to login with smtp, try to exluce '~' from the password, else deploy.sh has to be edited
```
engelhamer: Again 'exluce' is incorrect.

lti/lti_setup.xml
Larspolo: This is the same as the nodejs description, will change them both

PR #427, opened 2018-10-03
Added loading icon when there are open API requests

### LINKED ISSUE #416: Visible loading when waiting for API call to get response
# Feature request

## Description
Visible loading when waiting for API call to get response. Mostly important when a mail is being send, this take longer then expected.

Larspolo:
## Description
Added loading icon when there are open API requests. When reviewing this PR, please also look for yourself, I cant show how it looks in a single image.

## Summary of changes
- Fixes #416 
- Now users wont register if email server is incorrectly confirgured
- Removed old test file
- Reduced duplicate code in auth.js


engelhamer:
Don't like the current positioning. Would suggest moving the spinner (without any border) to the bottom left corner of the screen.

engelhamer:
Also, it would be nice to use a material spinner.

Airgetfrog:
Shouldn't the spinner ideally be positioned in the container that has loading data, or is that too large to implement?

engelhamer:
That would look pretty crazy with spinners alle over the place.

src/vue/src/api/auth.js
```diff
@@ -2,6 +2,7 @@ import connection from '@/api/connection'
 import statuses from '@/utils/constants/status_codes.js'
 import router from '@/router'
 import store from '@/store'
+import Store from '@/Store.vue'
```
mjvkeulen: Two stores?

src/vue/src/Header.vue
```diff
@@ -1,6 +1,9 @@
 <template>
     <!-- Section visible if user logged in -->
     <b-navbar v-if="loggedIn" id="header" class="shadow" toggleable="md" type="dark" fixed=top>
+        <div class="spinner small-shadow" id="spinner-loading-api-request" v-if="store.loadingApiRequests > 0">
```
engelhamer: Can you wrap this in a fade transition? I think that would look quite fancy.

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
...
+            )
+    ).catch(error => {
+        Store.loadingApiRequests--
+        return handleError(error, noRedirect)
+    })
+}
+
+function unvalidatedSend (func, url, data = null, noRedirect = true) {
+    Store.loadingApiRequests++
+    console.log(Store.loadingApiRequests)
```
engelhamer: What are we logging this for?

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
...
+    return func(url, data).then(
+        resp => {
+            Store.loadingApiRequests--
+            console.log(Store.loadingApiRequests)
```
engelhamer: Same

src/vue/src/Header.vue
```diff
@@ -59,16 +65,19 @@
 </template>
 
 <script>
+import store from '@/Store.vue'
```
mjvkeulen: Not needed due to mapgetters

src/vue/src/Header.vue
```diff
@@ -59,16 +65,19 @@
...
 import icon from 'vue-awesome/components/Icon'
 import { mapGetters } from 'vuex'
 
 export default {
     components: {
+        store,
```
mjvkeulen: not needed

src/vue/src/Header.vue
```diff
@@ -59,16 +65,19 @@
...
         icon
     },
     data () {
         return {
-            defaultProfileImg: '/static/unknown-profile.png'
+            defaultProfileImg: '/static/unknown-profile.png',
+            store: store
```
mjvkeulen: etc

src/vue/src/api/auth.js
```diff
@@ -63,85 +109,36 @@ export default {
...
     },
 
     /* Recover password */
     recoverPassword (username, recoveryToken, newPassword) {
-        return connection.conn.post('/recover_password/', {username: username, recovery_token: recoveryToken, new_password: newPassword})
+        return unvalidatedSend(connection.conn.post, improveUrl('recover_password'), {username: username, recovery_token: recoveryToken, new_password: newPassword})
     },
 
     get (url, data = null, noRedirect = false) {
-        if (url[0] !== '/') url = '/' + url
```
engelhamer: That code duplication though.

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -71,7 +71,7 @@ export default {
         return {
             selectedRole: '',
             selectedGroup: '',
-            init: true
+            init: 2
```
engelhamer: Could you explain what's with the numbers?

src/vue/src/Header.vue
mjvkeulen: With map getters

v-if="loadingApiRequests > 0"

More clean would be a boolean getter that does this check directly.

src/django/VLE/views/user.py
```diff
@@ -150,7 +150,10 @@ def create(self, request):
                                  verified_email=True if lti_id else False)
 
         if lti_id is None:
-            email_handling.send_email_verification_link(user)
+            if not email_handling.send_email_verification_link(user):
```
mjvkeulen: How would you feel about raising a custom error on send instead of a boolean? If raised delete etc

src/vue/src/Header.vue
```diff
@@ -1,6 +1,9 @@
 <template>
     <!-- Section visible if user logged in -->
     <b-navbar v-if="loggedIn" id="header" class="shadow" toggleable="md" type="dark" fixed=top>
+        <div class="spinner small-shadow" id="spinner-loading-api-request" v-if="store.loadingApiRequests > 0">
+            <icon name="spinner" pulse scale='1.3'/>
```
engelhamer: Maybe a `circle-notch` that will `spin` instead of this pulsating spinner will match our style better.

src/django/VLE/views/email.py
```diff
@@ -165,5 +170,8 @@ def send_feedback(request):
     except ValidationError as e:
         return response.bad_request(e.args[0])
 
-    email_handling.send_email_feedback(request.user, files, **request.POST)
-    return response.success(description='Feedback was succesfully received, thank you!')
+    if email_handling.send_email_feedback(request.user, files, **request.POST):
+        return response.success(description='Feedback was succesfully received, thank you!')
+    else:
+        return response.bad_request(
```
mjvkeulen: The request was correct, this is an internal server error. I believe the response should be more accurate and as its repeated could even be used as a dedicated response in response.py

src/django/VLE/views/email.py
```diff
@@ -33,6 +33,7 @@ def forgot_password(request):
 
     Generates a recovery token if a matching user can be found by either the prodived username or email.
     """
+    print('asd')
```
mjvkeulen: Oversight

src/django/VLE/utils/email_handling.py
```diff
@@ -97,6 +105,9 @@ def send_email_feedback(user, files, topic, ftype, feedback, user_agent, url):
...
         reply_to=[user.email]
     )
-
-    reply.send()
-    forward.send()
+    try:
+        reply.send()
+        forward.send()
+        return True
+    except SMTPAuthenticationError:
```
mjvkeulen: Can we not catch this one level higher and then return a dedicated response?

src/vue/src/components/course/CourseParticipantCard.vue
Larspolo: two watches both dont have to activate on the initialization. To make this happen, but functions subtract 1 from the init if init is larger then 0, and if its 0 (false) they will activate and run normally

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
     throw error
 }
 
+function validatedSend (func, url, data = null, noRedirect = false) {
+    Store.loadingApiRequests++
```
mjvkeulen: Do not manipulate store state directly. Please use mutations.

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
...
+    return func(url, data).then(
+        resp => {
+            Store.loadingApiRequests--
```
mjvkeulen: Do not manipulate store state directly. Please use mutations.

src/vue/src/Header.vue
Larspolo: This is the same as the Canvas spinner, I will change both then

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
...
+            return resp
+        }, error =>
+            store.dispatch('user/validateToken', error).then(_ =>
+                func(url, data).then(resp => {
+                    Store.loadingApiRequests--
```
mjvkeulen: etc

src/vue/src/api/auth.js
mjvkeulen: Please remove

src/vue/src/api/auth.js
mjvkeulen: Please remove

src/vue/src/api/auth.js
```diff
@@ -37,14 +38,59 @@ function handleError (error, noRedirect = false) {
...
+    Store.loadingApiRequests++
+    console.log(Store.loadingApiRequests)
+    return func(url, data).then(
+        resp => {
+            Store.loadingApiRequests--
+            console.log(Store.loadingApiRequests)
+            return resp
+        }, error => {
+            Store.loadingApiRequests--
+            console.log(Store.loadingApiRequests)
```
mjvkeulen: Please remove

src/vue/src/Header.vue
Larspolo: Not sure how that works, can you do this? Just adding 
`transition: all 0.6s cubic-bezier(.25,.8,.25,1) !important`
didnt work

engelhamer: Sure!

src/vue/src/views/Assignment.vue
```diff
@@ -50,8 +50,8 @@
                 </student-card>
             </b-link>
         </div>
-        <main-card v-if="assignmentJournals.length === 0" slot="main-content-column" class="no-hover" :line1="'No participants with a journal'"/>
-        <main-card v-else-if="filteredJournals.length === 0" slot="main-content-column" class="no-hover" :line1="'No journals found'"/>
```
engelhamer: There should still be something for when filteredjournals is empty. This would mean that there exist journals for this assignment, yet the current filter/search query doesn't match any.

src/vue/src/api/auth.js
engelhamer: @mjvkeulen we current seem to have two stores indeed. One from vuex and the other existed before but hasn't been ported. Nonetheless we do not required both here.

src/vue/src/Header.vue
Larspolo: Cant seem to import circle-notch. Maybe you want to take a look at this as well

engelhamer: Will do it right away.

engelhamer:
Added a slight delay to decreasing the running API call count, in order to prevent the spinner from flashing and provide an overall more consistent loading impression.

src/django/VLE/utils/email_handling.py
Larspolo: I find this solution a cleaner one.

src/django/VLE/views/user.py
Larspolo: I prefer catching it in the email function

mjvkeulen: try:
    send_email_feedback(request.user, files, **request.POST)
    return response.success(description='Feedback was succesfully received, thank you!')
except SMTPAuthenticationError:
  return response.internal_server_error(description='Mailserver is not configured correctly,...') 

This approach looks inline with the rest and straightforward compared to the added layer in the email handling?

src/vue/src/Header.vue
```diff
@@ -1,6 +1,11 @@
 <template>
     <!-- Section visible if user logged in -->
     <b-navbar v-if="loggedIn" id="header" class="shadow" toggleable="md" type="dark" fixed=top>
+        <transition name="fade">
+            <div class="spinner small-shadow" v-if="$store.getters['connection/checkOpenApiCalls']">
```
mjvkeulen: You can use map getters for this, see line 84. These are computed properties and have a nice shorthand.

src/vue/src/Header.vue
```diff
@@ -38,6 +43,11 @@
 
     <!-- Section visible if user logged out -->
     <b-navbar v-else id="header" class="shadow" toggleable="md" type="dark" fixed=top>
+        <transition name="fade">
+            <div class="spinner small-shadow" v-if="$store.getters['connection/checkOpenApiCalls']">
```
mjvkeulen: Same

src/vue/src/Store.vue
```diff
@@ -1,5 +1,6 @@
 <script>
 export default {
+    loadingApiRequests: 0,
```
mjvkeulen: No longer required

src/vue/src/api/auth.js
```diff
@@ -63,85 +115,36 @@ export default {
...
      * Checks if a user is known by the given email or username. Sends an email with a link to reset the password. */
     forgotPassword (username, email) {
-        return connection.conn.post('/forgot_password/', {username: username, email: email})
+        return unvalidatedSend(connection.conn.post, improveUrl('forgot_password'), {username: username, email: email})
     },
 
     /* Recover password */
     recoverPassword (username, recoveryToken, newPassword) {
-        return connection.conn.post('/recover_password/', {username: username, recovery_token: recoveryToken, new_password: newPassword})
+        return unvalidatedSend(connection.conn.post, improveUrl('recover_password'), {username: username, recovery_token: recoveryToken, new_password: newPassword})
```
mjvkeulen: We should really discuss the preferred style for js objects.

Larspolo: Not for this PR, but I agree

PR #428, opened 2018-10-04
Fix readonly input field appearance in profile

### LINKED ISSUE #407: Readonly text fields look the same as non readonly
# Bug report

## Expected Behavior
An indication that an input field is readonly

## Actual Behavior
No difference between the two
image
second input is readonly

JowehL:
## Description
Readonly field in profile page now looks different than normal input fields

## Summary of changes
- Fixes #407 

image


JowehL:
Or do you prefer the email verification to be not greyed out

image
@Larspolo

Larspolo:
This only needs to be readonly when its a user linked to Canvas, else it should be editable. Can that be fixed in this PR?

PR #429, opened 2018-10-05
Add selection field feature

### LINKED ISSUE #212: Selection template fields
There should be a template field that allows the user to choose one of multiple predefined options, for example using dropdowns and radio buttons.

Airgetfrog:
## Description
- Add 'options' field to Field model, intended for field types taking options.
- Added selection to backend structure and frontend structure.

The options input is a textfield with placeholder text used as instructions and generated through a function. How the data for options is handled is entirely up to the frontend implementation. It is simply a textfield, but it's accessible from the field object when needed so it should work fine. Possible issue is that there is no indication in grade view or overview that a field was a selection field (it's displayed in the same way as regular text inputs are).

image
image
image

## Summary of changes
- Fixes #212


src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -117,6 +119,14 @@ export default {
         },
         onUpdate () {
             this.updateLocations()
+        },
+        fieldOptionsPlaceholder (type) {
+            switch (type) {
+            case 's':
```
mjvkeulen: Cases are nested and should be indented accordingly

engelhamer:
@mjvkeulen adding even more inputs to a template field in the editor will make them enormous and difficult to handle (imagine dragging a field that covers the entire modal..)

Larspolo:
> And format these options with ';' behind the scenes?

Then all the semicolons in the options need to be encoded so that a ; inside an option, doesnt create two options

> @mjvkeulen adding even more inputs to a template field in the editor will make them enormous and difficult to handle (imagine dragging a field that covers the entire modal..)

A solution could be to minimize a field (maybe only showing the title) and only when clicked on, it expends.

engelhamer:
> A solution could be to minimize a field (maybe only showing the title) and only when clicked on, it expends.
But won't that make things even less intuitive than simply separating with ; ?

src/vue/src/components/template/TemplateEdit.vue
Airgetfrog: This is actually enforced by ESlint.

mjvkeulen: Fair enough, thanks!

engelhamer:
What does the UI look like now?

Airgetfrog:
Attempted to fix requested changes. It's best to take a look at the new UI and behavior yourself. The list is what I would call 'usable' but doesn't seem particularly user friendly (appearance can still be upgraded). Main drawback is not being able to see options without selecting the input. Another option could be a preview select but that seems somehow worse due to requiring an extra click even when using the text input. On the whole I'd say it's about equivalent to the previous solution.

engelhamer:
To be honest I like the original delimiter idea better.

Airgetfrog:
Seems like a fine idea Lars, I'll attempt to implement that. It might be unviable if we take into account responsiveness and such though. Engel what's your take on it?

engelhamer:
Lars' solution will not work well for options longer than a few words, and the amount of different buttons and inputs in one field just doesn't look very organized to me.

mjvkeulen:
Are we sure Lars' solution will not work? We could remove the crosses, and simply use backspace on an empty field as the key to delete them. This way the options will display just fine in our own style. They should also be able to scale just fine for any length.

src/django/VLE/migrations/0005_auto_20181005_2212.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('VLE', '0004_participation_role_required'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='field',
+            name='options',
+            field=models.TextField(null=True),
```
libre-man: This violates the first normal form as it seems you are saving more than a single value in this field.

Larspolo:
@Airgetfrog 
>It might be unviable if we take into account responsiveness and such though.
On mobile every option can be the full width

@engelhamer 
>Lars' solution will not work well for options longer than a few words
It will, you just have to scroll a bit. This was also an issue in the first example where there are multiple options in 1 input box, there you didnt have a clear overview on all the options if it was longer then the input box itself. Now the overview is there, but it is indeed a bit harder to read the options if they are too long

> and the amount of different buttons and inputs in one field just doesn't look very organized to me.
True, maybe we can add a card underneath the field type option where all of this is stored? This way there is a clearer differentiation between configuring the field type and the field itself

@mjvkeulen 
>We could remove the crosses, and simply use backspace on an empty field as the key to delete them.
I favor the crosses in this case, what I also imagined was that the crosses change to checkmarks when the option input has changed. I think that is a more intuitive solution for non tech-geeks then by removing when they are empty

libre-man:
You can also take a look at [vue-multiselect](https://vue-multiselect.js.org/), that is a vue library (so it is webscale) that enables you to easily create multi select dropdowns.

Airgetfrog:
Creating a full list might cause problems with reordering on mobile with longer selects though. Vue-multiselect doesn't seem to be a solution for our particular problem (configuration ui for a selection)? Since I'm not sure about db normal forms and we can't seem to find a solution I'd appreciate it if someone could help contribute to this branch.

libre-man:
Order van be achieved quite easily, simply add order numbers when inserting the subitems in to the database or always sort then bij name (seems the most intuitive option). Multiselect can be used when creating but also when selecting and it would directly support selecting multiple items.

libre-man:
Vue multiselect also support adding labels. At the demo page see the tagging demo.

mjvkeulen:
What is the current situation with this PR?

engelhamer:
We still haven't come up with a solution.

Airgetfrog:
I'm personally in favor of simple delimiting, but it might also be possible to complete the current wip. I was planning on taking a look at it over the next few days.

engelhamer:
It would be great if this makes it to develop by the end of the week.

Airgetfrog:
image
Looks like this!

mjvkeulen:
Current implementation looks good, much has changed since however. Possibly avoid updating develop until someone goes over the changes again?

src/django/VLE/validators.py
```diff
@@ -60,5 +64,18 @@ def validate_entry_content(data, field):
...
-        url_validate(data)
+        try:
+            url_validate = URLValidator(schemes=('http', 'https', 'ftp', 'ftps'))
+            url_validate(data)
+        except ValidationError as e:
+            raise e
+
+    if field.type == SELECTION:
+        if data not in json.loads(field.options):
+            raise ValidationError("Selection not in options")
```
Larspolo: A more descriptive error would be nice `Selected option is not in the given options`?

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -36,6 +36,29 @@
                                     Required
                                 </b-button>
                             </div>
+
+                            <!-- Field Options -->
+                            <div v-if="field.type == 's'">
+                                <!-- Event targeting allows us to access the input value -->
+                                <div class="d-flex">
+                                    <b-input class="multi-form mr-2 theme-input" placeholder="Enter an option" @keyup.enter.native="addSelectionOption($event.target, field)" @focus.native="selectedLocation = field.location" @focusout.native="selectedLocation = null"/>
```
Larspolo: I know we dont have a character limit, but this stretches it a bit far imo, 270 characters on 1 line

engelhamer:
Sorry.

mjvkeulen:
As this branch is rather behind, you might want to let someone have another look after updating with base.

Airgetfrog:
We should handle Lars' comments before merging. I'll get on it on Wednesday if nobody else does it first.

src/django/VLE/validators.py
```diff
@@ -60,5 +64,18 @@ def validate_entry_content(data, field):
         return
 
     if field.type == URL:
-        url_validate = URLValidator(schemes=('http', 'https', 'ftp', 'ftps'))
-        url_validate(data)
+        try:
+            url_validate = URLValidator(schemes=('http', 'https', 'ftp', 'ftps'))
+            url_validate(data)
+        except ValidationError as e:
+            raise e
```
Rickyboy320: This seems a bit unnecessary

src/vue/src/components/template/TemplatePreview.vue
```diff
@@ -72,6 +78,16 @@ export default {
...
+    },
+    methods: {
+        parseSelectionOptions (fieldOptions) {
+            if (!fieldOptions) {
+                return [{ value: null, text: 'Please select an option' }]
+            }
+            var options = JSON.parse(fieldOptions).filter(e => e).map(x => { return { value: x, text: x } })
+            options.unshift({ value: null, text: 'Please select an option' })
+            return options
+        }
```
Rickyboy320: Duplicated code block from above (not sure how we treat this in frontend specifically, but maybe move this somewhere where both files can access it?)

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -152,6 +158,14 @@ export default {
                 return null
             }
         },
+        parseSelectionOptions (fieldOptions) {
+            if (!fieldOptions) {
+                return [{ value: null, text: 'Please select an option' }]
+            }
+            var options = JSON.parse(fieldOptions).filter(e => e).map(x => { return { value: x, text: x } })
```
Rickyboy320: Interesting filter operation, what does it do exactly? (I might be missing some javascript magic here)

src/django/VLE/migrations/0008_can_view_assignment_journals_rename.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('VLE', '0005_presetnode_description'),
+        ('VLE', '0007_add_selection_field'),
```
Rickyboy320: By changing this file to 0008, you're essentially changing the already established history.
This file should remain 0006, the files depending on this should become 0007 and 0008 respectively (here 'instance' should come first, and then 'assignment_published' as per the established 0008 merge), then the 0008 merge can be removed as you did, and any changes you make on this branch should come after that.

src/django/VLE/migrations/0007_add_selection_field.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.1.2 on 2018-10-16 14:24
```
Rickyboy320: This file should become 0009_...

src/django/VLE/migrations/0006_assignment_published.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('VLE', '0006_can_view_assignment_journals_rename'),
+        ('VLE', '0005_presetnode_description'),
```
Rickyboy320: This should continue to depend on 0006, and file should be renamed to 0008 (as per established merge history)

src/django/VLE/migrations/0009_instance.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('VLE', '0006_can_view_assignment_journals_rename'),
+        ('VLE', '0008_can_view_assignment_journals_rename'),
```
Rickyboy320: this file should continue to depend on 0006_can_view_assignment_journals_rename, and should itself stay 0007.

src/vue/src/components/entry/EntryFields.vue
Airgetfrog: Filters out empty strings. Filter is not a map, it expects a function returning a boolean.

Rickyboy320: Ah right, had a gut feeling something like that was at play...
thanks for clearing that up!

PR #433, opened 2018-10-06
Added post date and edited add.

### LINKED ISSUE #426: Add post date to entry
# Feature request

## Description
Entries should have a date stamp and be sorted by post date.

## User story
As a user I want to see when an entry has been posted.

## User interface
A date in the timeline like presets currently have, yet a distinction in styling should be made between deadlines and post dates (i.e. underline).

Larspolo:
## Description
Added submitted on and edited at to an entry.
Submitted:
image

Edited:
image

In timeline (always submission date)
image

## Summary of changes
- Fixes #426


src/vue/src/components/timeline/TimelineNodeDate.vue
```diff
@@ -12,7 +12,9 @@ export default {
     props: ['date', 'selected', 'deadline'],
     computed: {
         dateClass () {
-            return this.deadline ? 'deadline-class' : (this.selected ? 'date-selected' : 'date-unselected')
+            var classes = this.deadline ? 'deadline-class' : ''
```
libre-man: Don't use `var`, use `let` or `const`.

src/vue/src/components/timeline/TimelineNode.vue
```diff
@@ -8,8 +8,8 @@
 <template>
     <b-row class="node-container">
         <b-col cols="4" sm="1"/>
-        <b-col cols="4" sm="5" class="d-flex h-100 align-items-center justify-content-center">
-            <timeline-node-date :date="node.deadline" :selected="selected"/>
+        <b-col cols="4" sm="5" class="d-flex h-100 align-items-center">
+            <timeline-node-date :date="node.deadline ? node.deadline : (node.entry ? node.entry.createdate : null)" :selected="selected" :deadline="node.deadline != null"/>
```
Airgetfrog: Why not just pass date and deadline and let timeline-node-date check for the deadline prop instead of passing both in date and a boolean in deadline? Seems a bit cleaner.

src/vue/src/components/timeline/TimelineNodeDate.vue
```diff
@@ -4,15 +4,17 @@
...
 
 <script>
 export default {
-    props: ['date', 'selected'],
+    props: ['date', 'selected', 'deadline'],
     computed: {
         dateClass () {
-            return (this.selected) ? 'date-selected' : 'date-unselected'
+            var classes = this.deadline ? 'deadline-class' : ''
+            classes += this.selected ? ' date-selected' : ' date-unselected'
```
Airgetfrog: Don't use strings for date binding, instead use [vue dictionary syntax](https://vuejs.org/v2/guide/class-and-style.html#ad).

src/vue/src/main.js
```diff
@@ -121,16 +121,25 @@ new Vue({
         getBorderClass (cID) {
             return this.colors[cID % this.colors.length]
         },
-        beautifyDate (date) {
+        beautifyDate (date, displayDate = true, displayTime = true) {
```
Airgetfrog: We should probably start using javascript Date objects here.

Airgetfrog:
Looks like we'll have to get rid of `var`s throughout our project :^)

libre-man:
Oh sorry I didn't know you were using `vars` everywhere. The `var` keyword is simply very strange, it binds to the enclosing function block, not any block, and there is not really a reason to still use it. Simply ignore my comment for this PR :)

src/vue/src/main.js
Larspolo: Not for this PR

Larspolo:
~150 vars in our project, lets do that another time. But I fixed the other issues

src/vue/src/components/timeline/TimelineNode.vue
Larspolo: otherwise I would have to deal with 2 variables where at least 1 would be not null. This is worse then doing it like this imo.

engelhamer:
How about this? The calendar will show a tooltip on hover, and will disappear once the deadline has passed.

deadline

src/vue/src/components/timeline/TimelineNodeDate.vue
```diff
@@ -4,40 +4,62 @@
...
+
 .date-unselected
     opacity: 0.35
-
-.deadline-class
-    opacity: 1
-    color: $theme-blue
-    font-weight: bold
-
+    cursor: pointer
```
Airgetfrog: Shouldn't there be a newline here

src/vue/src/components/timeline/TimelineNodeDate.vue
```diff
@@ -4,40 +4,62 @@
 -->
 
 <template>
-    <span v-if="date" class="date-text" :class="dateClass">{{ $root.beautifyDate(date, true, this.isDeadline) }}</span>
+    <span v-if="date" class="date-text" :class="dateClass">
+        <icon
```
Airgetfrog: Why not just v-if this instead of using css?

Airgetfrog: I don't see where the code is duplicated? It's less code to insert a single v-if on rendering than a conditional class with visibility css

src/vue/src/components/timeline/TimelineNodeDate.vue
Airgetfrog: That's what I'm saying, there should be one.

Larspolo: oww true, forgot

PR #434, opened 2018-10-07
Enhancement/image cropper

### LINKED ISSUE #384: Cropping tool baked into the user profile picture upload
# Feature request

## Description
We only allow square profile images due responsiveness handling. We require a cropping tool so the user doesn't have to square his desired profiled image manually.

Archifer:
## Description
Added the functionality of uploading any sized picture which will then be cropped and previewed to be the new profile picture, which can then be saved. The uploading and editing will be handled in a modal. The cropping can be done by using the mouse by moving and scrolling and a mobile user can just use his fingers and multi touch.

Here are pictures to show how the interface will look, 
image

image

image

If you save a new profile picture successfully a toast will appear.
image

Edit edit  (imagine all the pictures are centered like this one)
image

## Summary of changes
- Fixes #384


engelhamer:
Please center the picture in the modal.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,116 @@
...
+                canvas-color="transparent"
+                :show-remove-button="false"
+                :show-loading="true"
+                :loading-size="50"
+                accept="image/*"
+                :file-size-limit="this.$root.maxFileSizeBytes"
+                @file-type-mismatch="onFileTypeMismatch"
+                @file-size-exceed="onFileSizeExceed">
+            <img :src="$store.getters['user/profilePicture']"
+                slot="initial">
```
Larspolo: this should have to be on the next row
and the image tag should be indented one more to the right

engelhamer:
It looks like the picture is not really centered, there appears to be some relative positioning going on.

Larspolo:
Profile pic jumps when trying to move it again after pressed "Save"

Archifer:
I wasn't able to reproduce this  "Picture teleports a bit when trying to move it again after "Save" is pressed.", since for me it is working smoothly (maybe it was fixt with the previous fix of disappearing pictures).

Larspolo:
image

On default profile picture, it is zoomed out a bit. When uploading another picture or modifying the default a bit, it doesnt happen.

Archifer:
It wasn't intentional, since this screen should be a preview of how your profile picture would look like, and since it loads in your profile picture it should be displayed the way its displays the profile picture on the rest of the site.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,114 @@
...
+                :loading-size="50"
+                accept="image/*"
+                :file-size-limit="this.$root.maxFileSizeBytes"
+                @file-type-mismatch="onFileTypeMismatch"
+                @file-size-exceed="onFileSizeExceed"
+                :initial-image="pictureUrl"
+                initial-size="natural"
+                initial-position="center"/>
+        </div>
+        <br/><br/>
```
mjvkeulen: Wouldn't margins would be desired over breaks?

src/vue/src/main.js
```diff
@@ -7,6 +7,7 @@ import BootstrapVue from 'bootstrap-vue'
 import '../node_modules/bootstrap/dist/css/bootstrap.css'
 import '../node_modules/bootstrap-vue/dist/bootstrap-vue.css'
 import '../node_modules/flatpickr/dist/flatpickr.css'
+import '../node_modules/vue-croppa/dist/vue-croppa.css'
```
mjvkeulen: Croppa is only used in the profile section, and then on profile edit. Would moving this to the profile page or dedicated component not make more sense?

src/vue/src/main.js
```diff
@@ -60,11 +61,13 @@ import 'vue-awesome/icons/github'
 
 import Toasted from 'vue-toasted'
 import flatPickr from 'vue-flatpickr-component'
+import Croppa from 'vue-croppa'
 
 Vue.config.productionTip = false
 Vue.use(Toasted, { position: 'top-center', duration: 4000 })
 Vue.use(BootstrapVue)
 Vue.use(flatPickr)
+Vue.use(Croppa)
```
mjvkeulen: This component does not have to be globally available.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,114 @@
...
+        </b-button>
+        <b-button class="add-button float-right" @click="savePicture()">
+            <icon name="save"/>
+            Save
+        </b-button>
+    </b-card>
+</template>
+
+<script>
+import icon from 'vue-awesome/components/Icon'
```
engelhamer: I would move the 'Croppa' import to here as well.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,114 @@
...
+    .btn
+        position: absolute
+        width: 100%
+        height: 25%
+        bottom: -25%
+        opacity: 0
+    &:hover
+        .btn
+            bottom: 0px
+            opacity: 1
```
engelhamer: There should be a newline after this one.

src/vue/src/components/profile/ProfileData.vue
```diff
@@ -1,20 +1,18 @@
 <template>
     <b-row>
         <b-col md="5" sm="12" class="text-center">
+            <b-modal
+                ref="cropperModal"
+                title="Edit your profile picture"
```
engelhamer: I would suggest omitting 'your', so just 'Edit profile picture'.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,114 @@
...
+    display: block
+    position: relative
+    width: 100%
+    max-width: 250px
+    max-height: 250px
+    margin-bottom: 20px
+    border-radius: 50% !important
+    overflow: hidden
+    margin: 0 auto
+    croppa
```
engelhamer: Since croppa is not a valid HTML tag I would rather use a class here. Even though sass seems to still work it's magic on this.

src/vue/src/main.js
```diff
@@ -60,11 +61,13 @@ import 'vue-awesome/icons/github'
 
 import Toasted from 'vue-toasted'
 import flatPickr from 'vue-flatpickr-component'
+import Croppa from 'vue-croppa'
```
engelhamer: Please move this  to the image cropper component itself.

engelhamer:
By the way: is it not possible to make the image cover the preview? E.g. black bars will never appear.

Archifer:
> By the way: is it not possible to make the image cover the cropper (i.e. no black bars will ever appear)?

How the cropper now works is that it will load in the picture in the center, so if the selected picture is large enough it will cover the whole cropper. The only way the black 'bars' will appear is when the selected picture is too small or when you zoom out to much (it becomes too small) or by moving the picture to an edge, which is unavoidable for this cropper plug in.

Airgetfrog:
>By the way: is it not possible to make the image cover the cropper (i.e. no black bars will ever appear)?

I believe this is actually possible by accessing [croppa's own state data](https://github.com/zhanziyang/vue-croppa#-state-data) inside a watcher and using that to clamp the values to a custom range (s.t. data.imgdata.width <= props.width, data.imgdata.width - props.width < startx < 0). I'll leave it up to you whether that is desirable though.

src/vue/src/components/assets/ImageCropper.vue
```diff
@@ -0,0 +1,106 @@
...
+            this.croppa.refresh()
+        }
+    }
+}
+</script>
+
+<style lang="sass">
+@import '~sass/modules/breakpoints.sass'
+
+.profile-portrait-edit
```
engelhamer: You are really only copying all this for 4 lines. I will fix this now.

src/vue/src/sass/partials/modals.sass
```diff
@@ -19,3 +19,16 @@
             margin: 0px
             .card
                 margin-bottom: 8px
+
+.card
```
engelhamer: Defining a class in two distinct locations is not a good idea.

engelhamer:
On a side note: It would be nice if the original picture would remain on the server, so that a user can make subsequent alterations to the cropped picture without losing quality or adding black bars. Currently, once you click save all but what's in the view is lost. A user would have to upload a picture again if they desire another crop.

PR #436, opened 2018-10-07
Enhancement/file handling

### LINKED ISSUE #381: Non automatic file upload handling
# Feature request

## Description
A file submitted to an entry field should not actually be uploaded until the user posts the entire entry content. The solution can work with temporary files until submit is pressed or simply upload all files once the submit is pressed.

### LINKED ISSUE #383: Logical user file deletion that allow for duplicately named files in an assignment.
# Feature request

## Description
User files should be deleted if the content they are attached to is deleted. E.g. assignment/entry, but possibly also the content itself/node.

It should be possible for a user to enter files with a similar name but different content to the same assignment and entry without overwriting one another.

mjvkeulen:
## Description
UserFiles are now simply stored under MEDIA_ROOT/uID/aID/<file>
UserFiles are considered temporary on upload, and made permanent on post or save edit. Any temporary files are removed after the creation or patch of an entry.
The partial update of an entry now actually PATCHES instead of simply creating new content.

## Summary of changes
- Fixes #383 
- Fixes #381


src/django/VLE/views/entry.py
```diff
@@ -30,6 +32,8 @@ class EntryView(viewsets.ViewSet):
     def create(self, request):
         """Create a new entry.
 
+        Deletes remaining temporary user files if successfull.
```
Larspolo: successful

src/django/VLE/views/entry.py
```diff
@@ -196,23 +208,29 @@ def partial_update(self, request, *args, **kwargs):
...
+                    old_content = entry.content_set.get(pk=content_id)
+                    validators.validate_entry_content(data, field)
+                except KeyError:
+                    return response.keyerror('content.id', 'content.data', 'content.contentID')
+                except (Field.DoesNotExist, Content.DoesNotExist):
+                    return response.not_found('Field or content does not exist.')
+                except ValidationError as e:
+                    return response.bad_request(e.args[0])
+
+                if old_content.field.pk != int(field_id):
```
Larspolo: Shouldnt it already be an int? as it is used on line 214

src/django/VLE/views/responses.py
```diff
@@ -6,9 +6,9 @@
 to receive the appropriate error code.
 """
 from django.http import JsonResponse, HttpResponse, FileResponse
-
-import os
 from django.conf import settings
+from VLE.models import UserFile
```
Larspolo: VLE imports are preferably separated from the other imports

src/django/VLE/views/user.py
```diff
@@ -365,82 +365,80 @@ def download(self, request, pk):
...
 
-        return response.file(os.path.join(settings.MEDIA_ROOT, user_file.file.name))
+        return response.file(user_file)
 
-    # TODO P Test changes
     @action(methods=['post'], detail=False)
     def upload(self, request):
         """Update user profile picture.
 
         No validation is performed beyond a size check of the file and the available space for the user.
```
Larspolo: This line can be removed right?

src/django/VLE/views/responses.py
libre-man: Simply use isort, that solves all issues with import ordering.

src/django/VLE/views/entry.py
libre-man: No because file_id is used, not `field.pk`.

mjvkeulen: Well the query on 214 performs the int conversion, for the comparison it is required.

mjvkeulen: I do feel I explicit casting is desired

PR #439, opened 2018-10-08
Added global error catching

Larspolo:
## Description
- ObjectDoesNotExist
- ValidationError
- KeyError
- ValueError
- SMTPAuthenticationError

will now be caught in a global exception handler, this reduces a lot of error handling

## Summary of changes
- Fixes "This check occurs many times in the code, why not register a global exception handler. The custom error messages aren't really any better than generic ones (course does not exist)."


src/django/test/test_create_apis.py
```diff
@@ -45,21 +45,6 @@ def test_create_new_assignment(self):
         test.api_post_call(self, '/assignments/', params=create_assign_dict, login=login, status=201)
         self.assertEquals(Lti_ids.objects.get(lti_id=lti_id).assignment.name, 'SIFT')
 
-    def test_create_journal(self):
```
libre-man: It seems that this test should just be removed, is in't it useful?

Larspolo: No, we never use the create journal api call, because of what you said:
> Why can users create journals? They seem to be created automatically, so why do they also need to be able to create them later?

So I removed it from the viewset and in the tests

libre-man: Ah, but that should be a different PR right? Futhermore, some sort of test for creating journals seem to useful, even if it just tests the factory function.

Larspolo: True, should have done it on another branch.

This test didnt test anything about the creation of journals through other calls as well, so I left it. We will rework all the tests after the security issues are all fixed, with that rework I will also implement testing the creation of journals through other calls

src/django/VLE/utils/error_handling.py
```diff
@@ -12,7 +13,13 @@ def __call__(self, request):
...
         if isinstance(exception, ObjectDoesNotExist):
-            DoesNotExist = re.search(
-                r'([A-z .]*)matching query does not exist([A-z .]*[\']+)',
-                getattr(exception, 'message', repr(exception)))
-            return response.not_found(f'{DoesNotExist.group(1)} does not exist.')
+            return response.not_found(f'{str(exception).split()[0]} does not exist.')
+        elif isinstance(exception, ValidationError):
+            return response.bad_request(exception.args[0])
+        elif isinstance(exception, KeyError):
+            return response.key_error(str(exception))
```
libre-man: You should probably use some more specific class for this, as you will now catch all KeyErrors, even those because of bugs. The same for ValueErorrs. You can create something like `VLEMissingRequiredKey` and `VLEParamWrongType`.

src/django/VLE/views/responses.py
```diff
@@ -128,12 +128,20 @@ def json_response(payload={}, description='', status=None, reason=None, charset=
...
+        return bad_request(description='Field {0} is required but is missing.'.format(keys[0]))
     else:
-        return bad_request(description='Fields {0} are required but one or more are missing.'.format(keys))
+        return bad_request(description='Fields {0} are required but one or more are missing.'.format(', '.join(keys)))
+
+
+def value_error(message=None):
+    """Generate a bad request response with each given key formatted in the description."""
+    if message:
+        return bad_request(description=f'One or more field are invalid: {message}')
```
mjvkeulen: fields

src/django/VLE/settings/base.py
```diff
@@ -56,6 +56,8 @@
 }
 
 MIDDLEWARE = [
+    'VLE.utils.error_handling.ErrorMiddleware',
+
```
mjvkeulen: I understand the order/grouping, but a newline for this constructioN?

PR #440, opened 2018-10-08
Added automated import sorting

Larspolo:
## Description
Imports will now be ordered automatically. This  will improve consistency and now developers dont have to choose where to put the imports. It will order itself every time `make setup` is called.


libre-man:
You should probably also add this the the automated tests

Rickyboy320:
Make setup seems to be a weird time to cleanup the imports... in the code it seems it’s done in make reset which is a little more viable, but we shouldn’t expect make reset to be ran (if not needed).

I’ll approve after testing locally.
Libre’s suggestion for make test seems viable as well, though theoretically not the best spot either (though in practice it should always be ran before PRing so viable enough).

libre-man:
Well my suggestion was more that testing should check if imports are sorted, it shouldn't sort them.

PR #441, opened 2018-10-09
Add preset description

### LINKED ISSUE #419: Description field for preset entries
# Feature request

## Description
Predefined entries should have a description field. This makes it possible to reuse a template (such as a PDF template that is generic for 'uploading pdfs'), while keeping the freedom of explaining the reasoning for the entry. Currently this is only achievable by making a template for each entry.

A concrete example is the following:
An assignment (portfolio) requires us to upload multiple PDFs, each with different their own reasoning and deadline. Currently we cannot make one template that fits all these scenarios, because the first deadline needs a different description than the others, instead we have to make a template that embeds the new description.

## User story
As an user, I want to add a description to a preset-entry (entrydeadline) node such that I do not have to redefine templates for minor description changes.

Airgetfrog:
## Description
Adds a description to presets and progress nodes, allowing reuse of templates with different descriptions per preset as well as description of deadlines in general.

image
image

## Summary of changes
- Fixes #419


engelhamer:
Can you make it rich text and move the input to under the deadline selector?

Airgetfrog:
The input is currently the same as the one in the template editor and I feel that's appropriate, since I feel it's not intended for extensive descriptions like the assignment description, but it can be done of course. Shouldn't the description be put before the date as in course/assignment creation?

engelhamer:
Ah I see, rich text might not be desired then.

engelhamer:
@Airgetfrog migrations will need to be merged first.

PR #442, opened 2018-10-09
Hotfixed important issues

Larspolo:
## Description
Fixed some deployment issues

## Summary of changes
- Fixes migration folder
- Fixes login
- Fixes python3.6 -> python3


src/django/VLE/views/user.py
```diff
@@ -63,7 +63,9 @@ def retrieve(self, request, pk):
         """
         if not request.user.is_authenticated:
             return response.unauthorized()
-        if pk == 0:
+        print(pk)
+        print(type(pk))
```
libre-man: Shouldn't be here.

src/django/VLE/views/user.py
```diff
@@ -63,7 +63,9 @@ def retrieve(self, request, pk):
...
+        if pk == '0':
```
libre-man: Maybe simply convert it to an integer?

src/django/VLE/views/role.py
```diff
@@ -65,7 +65,10 @@ def retrieve(self, request, pk=0):
         if not request.user.is_authenticated:
             return response.unauthorized()
 
-        user = request.user if pk == 0 else User.objects.get(pk)
+        if pk == '0':
+            pk = request.user.id
+
+        user = User.objects.get(pk=pk)
```
Jeroentetje3: Is it supposed to check on the string '0' or just 0?

Larspolo: image
No

PR #444, opened 2018-10-10
Bugfix/security issues

Larspolo:
## Description
Fixed some security issues that were requested by Thomas

## Summary of changes
- assignment.py:62:0: Wrong permission is checked here. It should be `can_view_assignment_journals`.
- user.py:411:0: This seems like it should be extracted to a function.
- user.py:373:0: Old style formatting used here
- user.py:368:0: This should be done after the permission check
- user.py:218:0: This can be done in one query.
- user.py:250:0: You should not be able to delete the last superuser account.
- user.py:294:0: This should be implemented, or simply only allow this endpoint for superusers that will handle all (n=0 ;)) GDPR reqeuests.
- group.py:172:0: The util function should be used here.
- group.py:79:0: This should not be in the try block.
- group.py:38:0: The utils function should be used here.


src/django/VLE/views/user.py
```diff
@@ -294,8 +297,9 @@ def GDPR(self, request, pk):
         user = User.objects.get(pk=pk)
 
         # Check the right permissions to get this users data, either be the user of the data or be an admin.
-        if not (user.is_superuser or request.user.id == pk):
-            return response.forbidden('You are not allowed to view this user\'s data.')
+        if not user.is_superuser:
```
mjvkeulen: I believe this functionality is still desired and should not be locked to all but superusers?

src/django/VLE/views/user.py
```diff
@@ -237,6 +237,9 @@ def destroy(self, request, pk):
 
         user = User.objects.get(pk=pk)
 
+        if len(User.objects.filter(is_superuser=True)) == 1:
+            return response.bad_request('There is only 1 superuser left and therefor cannot be deleted')
```
mjvkeulen: Whilst not entirely incorrect therefore, is much more commonplace.

src/django/VLE/views/user.py
Larspolo: Else everyone can download a lot, and overload the server easily. Its still AVG compliant if they can access it by mailing to a superuser, and the superuser giving it to them

mjvkeulen: Whereas this call is costly, so are others. For example, send feedback mail with max size attachment. Are we sure we need to lock this call in this manner?

I believe adding a counter which tracks expensive calls to our User model would be a more proper solution. The counter can time out after XX minutes when not set.

As all expensive calls require an authenticated user, and these calls be short-circuited with a simple check. This should come along way to fixing this issue?

src/vue/src/components/profile/ProfileData.vue
```diff
@@ -109,10 +109,10 @@ export default {
                     document.body.appendChild(link)
                     link.click()
                     link.remove()
-                }, error => {
-                    this.$toasted.error(error.response.data.description)
+                }, _ => {
+                    this.$toasted.error('If you want to download userdata, please contact your supervisor.')
```
engelhamer: Would be nice to show a modal from which an email can be sent directly. (Just some explanatory text with a mailto button).

libre-man: When not using arguments in javascript you can simply leave them out, so this function should be something like `() => { this.$toasted.error(...); }`

Larspolo: This is not necessary anymore after the changes Maarten suggested

src/django/VLE/migrations/0006_user_last_gdpr.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0005_presetnode_description'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='last_gdpr',
```
mjvkeulen: I believe this implementation is not generic enough. The problem we want to address is limiting the rate expensive calls be accessed. A quick look around gave this Django solution, allows limiting user accesses on IP or authenticated basis Again on a view or viewset basis.

engelhamer: IP based doesn't sound very suitable given that users will often be on the same network.

libre-man: Another problem with that is for such expensive calls that you want to limit them across different machines/processes, so you still need some sort of backing store to do this and this not mentioned anywhere on that page.

mjvkeulen: > IP based doesn't sound very suitable given that users will often be on the same network.

As mentioned, it also allows for tracking on authentication basis, which would be very suitable here as all expensive calls require authorization.

src/django/VLE/utils/throttle.py
```diff
@@ -0,0 +1,26 @@
+
+from rest_framework.settings import api_settings
+from rest_framework.throttling import SimpleRateThrottle
+
+
+class GDPRThrottle(SimpleRateThrottle):
```
libre-man: Do realize that this will create a throttle per process, as the cache is stored in local memory, which of course is different for different processes.

Larspolo: Price to pay for the throttle class, but I guess its fine for now, just good to know in the future

Rickyboy320:
@mjvkeulen I believe the requested changes have been made, mind re-reviewing when you get the chance (or dismissing the review so someone else could)?

src/django/VLE/settings/base.py
```diff
@@ -44,6 +44,12 @@
     'DEFAULT_AUTHENTICATION_CLASSES': (
         'rest_framework_simplejwt.authentication.JWTAuthentication',
     ),
+    'DEFAULT_THROTTLE_CLASSES': (
+        'VLE.utils.throttle.GDPRThrottle',
+    ),
+    'DEFAULT_THROTTLE_RATES': {
+        'gdpr': '1/day',
```
mjvkeulen: I suggest we bump this to 3, giving leeway cross devices.

src/django/VLE/views/user.py
```diff
@@ -137,9 +137,9 @@ def create(self, request):
         if lti_id is None:
             try:
                 email_handling.send_email_verification_link(user)
-            except SMTPAuthenticationError:
+            except SMTPAuthenticationError as err:
```
mjvkeulen: I suggest we decide on a single naming convention now. Dislike seeing e, err, error. I believe we have many occurences of e already?

src/django/VLE/views/user.py
```diff
@@ -295,7 +297,7 @@ def GDPR(self, request, pk):
 
         # Check the right permissions to get this users data, either be the user of the data or be an admin.
         if not (user.is_superuser or request.user.id == pk):
-            return response.forbidden('You are not allowed to view this user\'s data.')
+            return response.bad_request(description='You are not allowed to view this user\'s data.')
```
mjvkeulen: This should remain a forbidden

src/vue/src/components/profile/ProfileData.vue
```diff
@@ -106,9 +106,13 @@ export default {
                     link.click()
                     link.remove()
                 }, error => {
-                    this.$toasted.error(error.response.data.description)
+                    if (error.response.status === 429) {
+                        this.$toasted.error('You have to wait a day before you can get your user data again.')
```
mjvkeulen: As we work with a setting config in the backend for this, either add this setting front end to app. I believe this was done to MAX_USER_FILE_SIZE and stuff. Or make this message generic.

src/django/VLE/utils/throttle.py
mjvkeulen: Some comments would be welcome here

src/django/VLE/views/user.py
```diff
@@ -338,14 +340,15 @@ def download(self, request, pk):
             request.query_params, (str, 'file_name'), (int, 'entry_id'), (int, 'node_id'), (int, 'content_id'))
 
         try:
-            user_file = UserFile.objects.get(author=pk, file_name=file_name, entry=int(entry_id), node=int(node_id),
-                                             content=int(content_id))
+            user_file = UserFile.objects.get(author=pk, file_name=file_name, entry=entry_id, node=node_id,
+                                             content=content_id)
+
+            if user_file.author.id != request.user:
```
Rickyboy320: You are comparing the id with the user itself.

src/django/VLE/views/user.py
Larspolo: everything is not called err

src/django/test/test_get_apis.py
```diff
@@ -95,6 +95,36 @@ def test_get_unenrolled_users(self):
...
+    def test_GDPR(self):
+        # Test normal user
+        login = test.logging_in(self, self.username, self.password)
+        _, _, other_user = test.set_up_user_and_auth('teacher', 'pass', 'teach@teach.com')
+
+        # Other user
+        test.api_get_call(self, '/users/{0}/GDPR/'.format(other_user.pk), login, status=403)
+
+        # Multiple times its own
+        test.api_get_call(self, '/users/0/GDPR/', login)
```
mjvkeulen: Should for loop this in range based on the settings param.

src/django/VLE/views/user.py
mjvkeulen: I believe all validation errors are thrown as e for example.

src/django/VLE/serializers.py
```diff
@@ -23,7 +23,7 @@ class Meta:
         model = User
         fields = ('username', 'first_name', 'last_name', 'name', 'profile_picture', 'is_teacher', 'lti_id', 'id',
                   'role', 'group')
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'lti_id', 'is_teacher')
```
mjvkeulen: Would adding username to this group not be desired?

src/django/VLE/serializers.py
```diff
@@ -62,7 +62,7 @@ class Meta:
         fields = ('id', 'last_login', 'username', 'first_name', 'last_name', 'is_active', 'email', 'permissions',
                   'name', 'lti_id', 'profile_picture', 'is_teacher', 'grade_notifications', 'comment_notifications',
                   'verified_email')
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'last_login', 'is_active', 'permissions', 'lti_id', 'is_teacher', 'verified_email')
```
mjvkeulen: Username should be read_only here as well. Furthermore, is_active is not in use atm maybe we should remove this field from the serializer for now? 

Apart from this once the lti_id is set, we want nearly all fields to be read only. Currently even with the proposed changes, someone with lti_id set could change his username, first and last name, email, name, and verified email status. All of which are undesired. This will require a check within the serializer.

src/django/VLE/serializers.py
```diff
@@ -110,7 +110,7 @@ class GroupSerializer(serializers.ModelSerializer):
     class Meta:
         model = Group
         fields = '__all__'
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'course', 'lti_id')
```
mjvkeulen: In group patch, a new group.name is set to the group but not saved. This group is then passed to the serializer which apparently then saves the entire group object on save. This is a dangerous workaround and should be handled by the serializer itself.

src/django/VLE/serializers.py
```diff
@@ -219,7 +219,7 @@ class CommentSerializer(serializers.ModelSerializer):
     class Meta:
         model = Comment
         fields = '__all__'
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'entry', 'author')
```
mjvkeulen: In comment patch, the request.data last_edited key is simply overwritten. Could this be moved to the serializer? Making  last_edited a read-only field which is set by a function of the datetime.now()?
Furthermore, the timestamp field should be read-only, as its only initialized on the creation of the comment.

src/django/VLE/serializers.py
```diff
@@ -239,7 +239,7 @@ class JournalSerializer(serializers.ModelSerializer):
     class Meta:
         model = Journal
         fields = '__all__'
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'assignment', 'user')
```
mjvkeulen: Currently the grade_url can be changed by the serializer, and no checks are present in the Journal patch to delete said key.

src/django/VLE/serializers.py
```diff
@@ -310,7 +310,7 @@ class Meta:
         model = Entry
         fields = ('id', 'createdate', 'published', 'template', 'content',
                   'editable', 'grade', 'last_edited', 'comments')
-        read_only_fields = ('id', )
+        read_only_fields = ('id', 'template', 'createdate')
```
mjvkeulen: Here too, we should move last_edited update functionality to the serializer. All other keys are filtered in the patch path and should be secure.

src/django/VLE/views/user.py
Larspolo: image

src/django/VLE/serializers.py
Larspolo: > Could this be moved to the serializer?

Dont know how, want to rework the serializers some other time. This will be fixed this as well

src/django/VLE/serializers.py
Larspolo: >This will require a check within the serializer.

Also dont know how yet, will do it in the user.partial_update as that is the only place where a user can be updated

src/django/VLE/serializers.py
Larspolo: Added readonly to both

src/django/VLE/views/assignment.py
```diff
@@ -197,6 +197,8 @@ def partial_update(self, request, *args, **kwargs):
                 del req_data['published']
 
             data = request.data
+            if not request.user.is_superuser and 'author' in data:
+                del data['author']
```
mjvkeulen: What is happening here, req_data and data variables are set based on the request.data. However published is deleted from req_data after which that variable is ignored.

src/django/VLE/serializers.py
mjvkeulen: Currently the published and content field are set outside of the serializer context. Shall we make these fields read only for security?

PR #445, opened 2018-10-12
Feature/add student revamp

### LINKED ISSUE #355: Add users to a course needs to be changed
# Feature request

## Description
A teacher should be able to get the whole list of all the student, they should only be able to add them by student id/name

## User interface
image
This should be just an input field iso a list of all users

JowehL:
## Description
Adding students to a course is not done with a list of all unenrolled students, but with list of unenrolled student matching with a search query of at least 5 characters.

## Summary of changes
- Fixes #355 

screenshot from 2018-10-12 13-36-05
Searching without input

screenshot from 2018-10-12 13-36-12
Searching on 'Student'

screenshot from 2018-10-12 13-36-28
Searching on 'Student4'


engelhamer:
I have a few concerns:
- 'Numbers' should be 'characters'.
- Why do you need an extra search bar for this? I would suggest removing it and just keeping solely the old one.
- I don't think toasts are desirable here. A card stating one has to type at least N characters would be more appropriate.
- We should introduce a username constraint in terms of length (at least N characters), so that every username is searchable.
- I think the input should not disappear after pressing return.

Larspolo:
The code looks fine by me, but Engel has a few valid concerns, escepecially 
>We should introduce a username constraint in terms of length (at least N characters), so that every username is searchable.

needs to be implemented

src/vue/src/views/CourseEdit.vue
```diff
@@ -82,10 +82,24 @@
...
                         type="text"
                         v-model="searchVariable"
                         placeholder="Search..."/>
+                    <input
+                        v-if="!viewEnrolled"
+                        class="theme-input flex-grow-1 no-width multi-form mr-2"
+                        type="text"
+                        v-model="unenrolledQuery"
+                        placeholder="Username with at least 5 characters"
+                        v-on:keyup.enter="searchUnenrolled"/>
```
Airgetfrog: Use @

src/vue/src/views/CourseEdit.vue
```diff
@@ -121,6 +135,12 @@
                 </b-col>
             </b-card>
 
+            <b-card class="no-hover" v-if="!viewEnrolled && !this.unenrolledStudents.length">
+                <div class="float-left">
+                    <b>{{unenrolledQueryDescription}}</b>
```
Airgetfrog: Use whitespace in moustache syntax {{ like so }}

src/vue/src/views/CourseEdit.vue
```diff
@@ -195,7 +199,8 @@ export default {
             roles: [],
             viewEnrolled: true,
             order: false,
-            unenrolledQuery: ''
+            unenrolledQuery: '',
+            unenrolledQueryDescription: 'Search unenrolled users by using the search field in "Manage course members"'
```
Airgetfrog: Consider changing this to "[...] in the search field above."

JowehL:
image

JowehL:
image

image

image

JowehL:
image

image

@engelhamer, I prefer the first one

engelhamer:
Fine by me, as long as the button contains an icon.

src/django/VLE/views/participation.py
```diff
@@ -191,12 +192,26 @@ def unenrolled(self, request):
         if not request.user.is_authenticated:
             return response.unauthorized()
 
-        course_id, = utils.required_params(request.query_params, "course_id")
+        course_id, unenrolled_query = utils.required_params(request.query_params, 'course_id', 'unenrolled_query')
+        if ' ' in unenrolled_query:
+            first_name = unenrolled_query.split(' ', 1)[0]
```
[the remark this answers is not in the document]
Larspolo: first_name and last_name are default for django authentication

src/django/VLE/views/participation.py
```diff
@@ -191,12 +192,26 @@ def unenrolled(self, request):
...
+            last_name = unenrolled_query.split(' ', 1)[1]
+
+        if len(unenrolled_query) < 5:
```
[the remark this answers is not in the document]
Larspolo: I prefer to give suggestions, makes it easier for the teacher, gitlab does the same

src/django/VLE/views/participation.py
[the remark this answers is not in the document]
engelhamer: Well, this relates to search right? We currently just parse the full name given through LTI by splitting on spaces, then store the first word as the first name and what is left as the last name. This does not seem tot do all names justice and will therefore introduce cumbersome issues in cases like this.

src/django/VLE/management/commands/preset_db.py
```diff
@@ -53,7 +53,7 @@ def gen_users(self):
                 "is_superuser": False,
                 "is_teacher": False
             }, {
-                "username": "Student5",
+                "username": "zzhu",
```
Larspolo: Why this?

engelhamer: So we can easily test searching for <5 character usernames.

src/django/VLE/views/participation.py
```diff
@@ -169,12 +170,29 @@ def unenrolled(self, request):
...
-        return response.success({'participants': UserSerializer(users, many=True).data})
+
+        if len(unenrolled_query) < 5:
+            user = users.filter(username=unenrolled_query)
+            if user:
+                return response.success({'participants': UserSerializer(user, many=True).data})
+            else:
+                return response.success({'participants': []})
+
+        found_users = users.filter(username__contains=unenrolled_query) | \
```
mjvkeulen: As queries are evaluated lazily I believe this won't matter performance wise? But I think this is cleaner:

from django.db.models import Q

`found_users = users.filter(Q(username__contains=unenrolled_query) | Q(first_name__contains=unenrolled_query) | Q(last_name__contains=unenrolled_query))`

src/django/test/test_get_apis.py
```diff
@@ -58,9 +58,12 @@ def test_get_course_users(self):
 
         self.assertEquals(len(response.json()['participants']), 3)
 
-        response = test.api_get_call(self, '/participations/unenrolled/', login, params={'course_id': self.course.pk})
+        response = test.api_get_call(self,
```
mjvkeulen: In python do we not simply fit the line length before separating arguments?

Larspolo: We use a lot of different things, maybe an item to discuss this evening

src/django/VLE/views/participation.py
Larspolo: prob also faster

Larspolo:
Please also supply tests

Rickyboy320:
I believe all requested changes have been made, and this PR has been open for a while, mind rereviewing/dismissing @engelhamer?

PR #446, opened 2018-10-12
Feature/lti tests

BigDGW:
## Description
Added test for all lti functions with the exception of the gradepassback outgoing calls + bugfix lti multiple roles and lti unknown roles.


Makefile
```diff
@@ -12,6 +12,9 @@ test-front:
 	npm run lint --prefix ./src/vue
 	npm run test --prefix ./src/vue
 
+test-coverage:
+	bash -c "source ./venv/bin/activate && coverage run src/django/manage.py test src/django && coverage report -m && deactivate"
```
Larspolo: Inside the `make test-back` command you cna now just type `make test-coverage` to reuse code. However this line is slightly changed compared to whats in `make test-back`, why is there a `-m` in `coverage report -m`?

BigDGW: coverage report -m adds a extra collumn to the report that show the lines and branches that are not covered in each file. This is really nice when working on test, but when just normally testing it unneeded information that takes a lot of space in your terminal.

engelhamer:
I like seeing the coverage increase for once! 😉

PR #448, opened 2018-10-12
Better backend permissions

Rickyboy320:
## Description
Backend permissions now also (more strictly) follow the three level model.
Permissions are now explicitly separated into the three levels, and there is no longer an overlap.

Previously it was possible to check against permissions in multiple ways (getting role, using the helper function, etc...), this is now all (forcefully) merged into one function, which is accessible through the user model.

For general permissions, use: user.has_permission('some_general_permission')
For assignment permissions, use: user.has_permission('some_assignment_permission', assignment)
For course permissions, use: user.has_permission('some_course_permission', course),
where assignment and course are a valid assignment or course.

I also improved the error handling frontend and backend, no longer silently returning false when passed an incorrect permission. It now correctly throws an error, such that it is easier to catch issues.

The most important thing to keep track of is that due to the forced three-level split of the permissions, querying an assignment permission on a course is no longer possible. I have solved all issues this may have brought, both frontended and backended. (Surprisingly there were only about 4 instances of this happening). This also makes it more predictable what the behaviour will be, as assignment permissions can no longer be accidentally queried from a course, which could cause a desync between the actual value and the course value. Now regardless of what course is used to get to the assignment, the permissions are always the same.

Also fixed an issue where the assignment permission can_have_journal wasn't negated properly, as should be dictated by the permissions document.

I have tested all occurences of $hasPermission in the frontend, and all seem to work with the new system. Backend logic seems alright, but may have made a typo somewhere. I've read the diff multiple times, so I am fairly sure there are no issues.


src/django/VLE/views/responses.py
```diff
@@ -11,7 +11,7 @@
 from django.conf import settings
 from django.http import FileResponse, HttpResponse, JsonResponse
 
-from VLE.models import UserFile
+import VLE.models
```
Larspolo: Why this change?

src/vue/src/store/constants/permission-types.js
```diff
@@ -12,28 +12,25 @@ const COURSE_LEVEL_PERMISSIONS = new Set([
...
     'can_edit_assignment',
-    'can_delete_assignment',
     'can_view_assignment_journals',
     'can_grade',
-    'can_publish_grades'
-])
-
-const JOURNAL_LEVEL_PERMISSIONS = new Set([
+    'can_publish_grades',
+    // Journal level
```
Larspolo: ?

src/django/VLE/views/responses.py
Rickyboy320: This is to prevent circular imports. Importing from the VLE.models file would cause an Attribute error because this file is loaded during the loading of models (through the permissions class).

src/django/VLE/views/group.py
```diff
@@ -39,13 +38,10 @@ def list(self, request):
...
-        if role is None:
-            return response.forbidden('You are not a participant of this course.')
-        if not (role.can_view_course_users or
-                role.can_edit_course_user_group or
-                role.can_add_course_user_group or
-                role.can_delete_course_user_group):
+        if not (request.user.has_permission('can_view_course_users', course) or
+                request.user.has_permission('can_edit_course_user_group', course) or
+                request.user.has_permission('can_add_course_user_group', course) or
+                request.user.has_permission('can_delete_course_user_group', course)):
```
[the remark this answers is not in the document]
Rickyboy320: Can_add_course_user_group and can_delete_course_user_group are both not constrained under the permission 'can_view_course_users'. This means that if you only have that permission, and not can_view_course_users, you would be unable to view the groups in the course. It's a bit hard to work with groups if you cannot view what you are working with, so I am quite sure this permission check (albeit very verbose) is correct.

src/vue/src/store/constants/permission-types.js
Rickyboy320: I do not know why there were Journal level permissions in the first place, but I added this comment to remind what permissions were previously 'journal level'.

src/django/VLE/views/entry.py
```diff
@@ -147,22 +146,19 @@ def partial_update(self, request, *args, **kwargs):
 
         journal = entry.node.journal
```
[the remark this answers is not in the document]
Rickyboy320: :+1:

src/django/VLE/utils/error_handling.py
```diff
@@ -13,6 +13,10 @@ class VLEParamWrongType(ValueError):
     pass
 
 
+class VLEProgrammingError(Exception):
+    pass
```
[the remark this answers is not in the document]
Rickyboy320: :+1:

src/django/VLE/views/comment.py
```diff
@@ -90,13 +89,12 @@ def create(self, request):
         journal = Journal.objects.get(node__entry=entry)
         assignment = Assignment.objects.get(journal=journal)
 
-        if not permissions.has_assignment_permission(request.user, assignment, 'can_comment') or \
-           not (permissions.has_assignment_permission(request.user, assignment, 'can_view_assignment_journals') or
+        if not request.user.has_permission('can_comment', assignment) or \
+           not (request.user.has_permission('can_view_assignment_journals', assignment) or
```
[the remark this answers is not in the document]
Rickyboy320: I left the not's there, as it's clearer what the if statement does with them. I did remove the parentheses for consistency though.

src/django/VLE/views/comment.py
```diff
@@ -90,13 +89,12 @@ def create(self, request):
         journal = Journal.objects.get(node__entry=entry)
         assignment = Assignment.objects.get(journal=journal)
 
-        if not permissions.has_assignment_permission(request.user, assignment, 'can_comment') or \
-           not (permissions.has_assignment_permission(request.user, assignment, 'can_view_assignment_journals') or
-                journal.user == request.user):
+        if not request.user.has_permission('can_comment', assignment) or \
+           not request.user.has_permission('can_view_assignment_journals', assignment) or \
```
engelhamer: I think the parentheses are required here. This way you cannot ever comment unless you are the author of the journal *and* you can view assignment journals at the same time.

libre-man: Maybe this is a clear signal that the check should be extracted into a function...

src/django/VLE/views/participation.py
```diff
@@ -63,8 +59,8 @@ def retrieve(self, request, pk=None):
         course = Course.objects.get(pk=pk)
         participation = Participation.objects.get(user=request.user, course=course)
 
-        if not permissions.is_user_in_course(request.user, course):
-            return response.forbidden('You are not in this course.')
+        if not request.user.is_participant(course):
```
mjvkeulen: Appears like this should be combined with the two queries above, should be a single query. (Not your PR but just noticed this).

src/django/VLE/views/participation.py
```diff
@@ -148,10 +140,7 @@ def partial_update(self, request, pk):
         course = Course.objects.get(pk=pk)
         participation = Participation.objects.get(user=user, course=course)
 
-        role = permissions.get_role(request.user, course)
-        if role is None:
```
mjvkeulen: Again is a not_found desired over a forbidden?

src/django/VLE/permissions.py
```diff
@@ -3,272 +3,118 @@
...
-    permissions = get_assignment_permissions(user, assignment)
-    return permission in permissions and permissions[permission]
+    permissions = defaultdict(lambda: False)
+    for course in assignment.courses.all():
+        if user.is_participant(course):
+            role = VLE.models.Participation.objects.get(user=user, course=course).role
+            role_permissions = model_to_dict(role)
+            permissions = {key: (role_permissions[key] or permissions[key]) for key in ASSIGNMENT_PERMISSIONS}
 
+    # Apply negations
```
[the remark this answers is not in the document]
engelhamer: Can_grade depends on can_view_assignment_journals and therefore this should be sufficient.

src/django/VLE/permissions.py
```diff
@@ -3,272 +3,118 @@
...
-    for permission in permission_list:
-        if permission not in permissions or not permissions[permission]:
-            return False
-
-    return True
+    if permission not in COURSE_PERMISSIONS:
+        raise VLEProgrammingError("Permission " + permission + " is not a course level permission.")
 
+    if user.is_superuser:
+        return True
```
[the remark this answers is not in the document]
Rickyboy320: Having a journal is an assignment permission and is handled there correctly.

src/django/VLE/views/participation.py
Rickyboy320: In this particular case, yes, because we're updating the participation of the user (and if that does not exist, then that is a valid 'not found').
I agree that in the other cases a forbidden is more suitable.

src/django/VLE/views/group.py
```diff
@@ -115,10 +110,7 @@ def partial_update(self, request, *args, **kwargs):
         course = Course.objects.get(pk=course_id)
         group = Group.objects.get(name=old_group_name, course=course)
 
-        role = permissions.get_role(request.user, course)
-        if role is None:
```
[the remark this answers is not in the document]
Rickyboy320: I forgot, but I actually added this check in has_course_permission and has_assignment_permission in permissions.py. This means that if the user is not participating in the course, it always returns False for a permission request (which is probably intended).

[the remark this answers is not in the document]
Rickyboy320: I have now added check_participation that performs this check (throwing a VLEParticipationError if invalid). All has_permission checks will also perform this check, ensuring that the correct error is thrown in every situation.

This ParticipationError will result in a forbidden, which should resolve the issue if it wasn't already.

src/django/VLE/models.py
```diff
@@ -120,6 +123,41 @@ class User(AbstractUser):
...
+    def check_participation(self, obj):
+        if not self.is_participant(obj):
+            raise VLEParticipationError(obj)
+
+    def is_participant(self, obj):
+        if isinstance(obj, Course):
+            return Course.objects.filter(pk=obj.pk, users=self).exists()
+        if isinstance(obj, Assignment):
+            return Assignment.objects.filter(pk=obj.pk, courses__users=self).exists()
+        raise VLEProgrammingError("obj must be of type Course or Assignment.")
```
Larspolo: This is not consistant with `Permission object must be of type None, Course or Assignment.` either choose to just give the parameter name, or maybe change `obj` to `Participant object`

src/django/VLE/views/assignment.py
```diff
@@ -50,14 +49,16 @@ def list(self, request):
...
-        else:
-            queryset = Assignment.objects.filter(courses=course, journal__user=request.user)
+        # Consider all assignments that the user is in, or can grade.
+        ids = []
+        for assignment in course.assignment_set.all():
+            if request.user.has_permission('can_grade', assignment) or \
+               request.user.has_permission('can_have_journal', assignment):
+                ids.append(assignment.id)
+
+        queryset = Assignment.objects.filter(id__in=ids)
```
Larspolo: Cant you also just append the assignments, and use a list on the assignments in the assignmentserializer?

src/django/VLE/views/comment.py
```diff
@@ -90,13 +88,11 @@ def create(self, request):
         journal = Journal.objects.get(node__entry=entry)
```
Larspolo: journal = entry.node.journal
assignment = journal.assignment

src/django/VLE/views/comment.py
```diff
@@ -122,11 +118,10 @@ def retrieve(self, request, pk=None):
             return response.unauthorized()
 
         comment = Comment.objects.get(pk=pk)
+        assignment = comment.entry.node.journal.assignment
```
Larspolo:
```journal = entry.node.journal
assignment = journal.assignment```
to also change `comment.entry.node.journal.user` to `journal.user`

File: src/django/VLE/views/common.py [hunk 17]
```diff
@@ -32,19 +31,23 @@ def names(request, course_id, assignment_id, journal_id):
     result = {}
     if course_id:
         course = Course.objects.get(pk=course_id)
-        if not Participation.objects.filter(user=request.user, course=course).exists():
-            return response.forbidden('You are not a particpant of this course.')
+
```
Larspolo: too many empty lines imo

src/django/VLE/views/entry.py
```diff
@@ -49,21 +48,21 @@ def create(self, request):
...
 
             if not (node.preset and node.preset.forced_template == template):
                 return response.forbidden('Invalid template for preset node.')
-        elif not journal.assignment.format.available_templates.all().filter(pk=template.pk).exists():
+        elif not assignment.format.available_templates.all().filter(pk=template.pk).exists():
             return response.forbidden('Entry template is not available.')
 
-        if not permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal'):
-            return response.forbidden('You are not allowed to create an entry for this journal.')
+        request.user.check_permission('can_have_journal', assignment)
```
Larspolo: Permission check should be done first

src/django/VLE/views/entry.py
```diff
@@ -146,24 +145,22 @@ def partial_update(self, request, *args, **kwargs):
         grade, published, content_list = utils.optional_params(request.data, "grade", "published", "content")
 
         journal = entry.node.journal
+        assignment = journal.assignment
 
-        if grade and \
-           not permissions.has_assignment_permission(request.user, journal.assignment, 'can_grade'):
+        if grade and not request.user.has_permission('can_grade', assignment):
```
Larspolo:
```
if not grade:
    request.user.check_permission('can_grade', assignment)
```

src/django/VLE/views/entry.py
```diff
@@ -146,24 +145,22 @@ def partial_update(self, request, *args, **kwargs):
...
             return response.forbidden('You cannot grade or publish entries.')
         elif isinstance(grade, (int, float)) and grade < 0:
             return response.bad_request('Grade must be greater or equal to zero.')
         elif grade:
             entry.grade = grade
 
-        if published is not None and \
-           not permissions.has_assignment_permission(request.user, journal.assignment, 'can_publish_grades'):
+        if published is not None and not request.user.has_permission('can_publish_grades', assignment):
```
Larspolo:
```
if published:
```

src/django/VLE/views/entry.py
```diff
@@ -146,24 +145,22 @@ def partial_update(self, request, *args, **kwargs):
...
             return response.forbidden('You cannot publish entries.')
 
-        if ((journal.assignment.unlock_date and journal.assignment.unlock_date > datetime.now()) or
-            (journal.assignment.lock_date and journal.assignment.lock_date < datetime.now())) and \
-           not permissions.has_assignment_permission(request.user, journal.assignment,
-                                                     'can_view_assignment_journals'):
-            return response.bad_request('The assignment is locked and unavailable for students.')
+        if ((assignment.unlock_date and assignment.unlock_date > datetime.now()) or
+            (assignment.lock_date and assignment.lock_date < datetime.now())) and \
```
Larspolo: These two lines appear 3 times, maybe this can be extracted to another function? I think you can put it in the assignment model itself
`assignment.is_locked()`

src/django/VLE/views/entry.py
```diff
@@ -179,14 +176,13 @@ def partial_update(self, request, *args, **kwargs):
...
-            if not permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal'):
-                return response.forbidden('You are not allowed to have a journal.')
+            request.user.check_permission('can_have_journal', assignment)
 
             if entry.grade is not None:
                 return response.bad_request('You are not allowed to edit graded entries.')
 
             if entry.node.type == Node.ENTRYDEADLINE and entry.node.preset.deadline < now() or \
-               (journal.assignment.due_date and journal.assignment.due_date < datetime.now()):
+               (assignment.due_date and assignment.due_date < datetime.now()):
```
Larspolo: shouldnt due date also be checked in this code above:
```
        if ((assignment.unlock_date and assignment.unlock_date > datetime.now()) or
            (assignment.lock_date and assignment.lock_date < datetime.now())) and \
```

src/django/VLE/views/entry.py
```diff
@@ -255,18 +252,18 @@ def destroy(self, request, *args, **kwargs):
...
 
         if entry.node.type == Node.ENTRYDEADLINE and \
-           permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal'):
+           request.user.has_permission('can_have_journal', assignment):
             return response.forbidden('You cannot delete deadlines.')
 
-        if journal.assignment.due_date and journal.assignment.due_date < datetime.now() and \
-           permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal'):
-            return response.forbidden('You cannot delete an entry whose deadline deadline has already passed.')
+        if assignment.due_date and assignment.due_date < datetime.now() and \
```
Larspolo: `assignment.is_due()`?

src/django/VLE/views/entry.py
```diff
@@ -255,18 +252,18 @@ def destroy(self, request, *args, **kwargs):
...
+           request.user.has_permission('can_have_journal', assignment):
+            return response.forbidden('You cannot delete an entry for which the deadline has already passed.')
 
-        if (journal.assignment.unlock_date and journal.assignment.unlock_date > datetime.now()) or \
-           (journal.assignment.lock_date and journal.assignment.lock_date < datetime.now()) and \
-           permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal'):
+        if (assignment.unlock_date and assignment.unlock_date > datetime.now()) or \
+           (assignment.lock_date and assignment.lock_date < datetime.now()) and \
```
Larspolo: `assignment.is_locked()`

src/django/VLE/views/group.py
Larspolo: why not user `check_permission` for these?

src/django/VLE/views/journal.py
```diff
@@ -74,10 +72,10 @@ def retrieve(self, request, pk):
 
         journal = Journal.objects.get(pk=pk)
 
-        if not (journal.user == request.user and
-                permissions.has_assignment_permission(request.user, journal.assignment, 'can_have_journal')) and \
-           not permissions.has_assignment_permission(request.user, journal.assignment, 'can_view_assignment_journals'):
-            return response.forbidden('You are not allowed to view this journal.')
+        if journal.user != request.user:
+            if not request.user.has_permission('can_have_journal', journal.assignment) and \
+               not request.user.has_permission('can_view_assignment_journals', journal.assignment):
```
Larspolo: Maybe change it to:
```
            if not (request.user.has_permission('can_have_journal', journal.assignment) or
                      request.user.has_permission('can_view_assignment_journals', journal.assignment)):
```
To stay consistent with other mutliple permission checks

src/django/VLE/views/node.py
```diff
@@ -50,14 +49,12 @@ def list(self, request):
...
-            return response.forbidden('You are not allowed to view journals of other participants.')
+        if request.user != journal.user:
+            request.user.check_permission('can_view_assignment_journals', journal.assignment)
 
         if ((journal.assignment.unlock_date and journal.assignment.unlock_date > datetime.now()) or
             (journal.assignment.lock_date and journal.assignment.lock_date < datetime.now())) and \
-           not permissions.has_assignment_permission(request.user, journal.assignment,
-                                                     'can_view_assignment_journals'):
-            return response.bad_request('The assignment is locked and unavailable for students.')
+           not request.user.has_permission('can_view_assignment_journals', journal.assignment):
```
Larspolo:
```
if assignment.is_locked():
    request.user.check_permission('can_view_assignment_journals', assignment)

File: src/django/VLE/views/role.py [hunk 27]
```diff
@@ -74,18 +70,15 @@ def retrieve(self, request, pk=0):
...
+                return response.success({'role': permissions.serialize_course_permissions(request.user, course)})
         # Return assignment permissions if assignment_id is set
         except (VLEMissingRequiredKey, VLEParamWrongType):
             assignment_id, = utils.required_typed_params(request.query_params, (int, 'assignment_id'))
-            perms = permissions.get_assignment_id_permissions(request.user, assignment_id)
-            return response.success({'role': perms})
+            assignment = Assignment.objects.get(pk=assignment_id)
+
+            return response.success({'role': permissions.serialize_assignment_permissions(request.user, assignment)})
         # Returns keyerror if course_id nor assignment_id is set
```
Larspolo: can be removed, old comment

src/django/VLE/views/user.py
```diff
@@ -40,7 +40,7 @@ def list(self, request):
         if not request.user.is_authenticated:
             return response.unauthorized()
 
-        if not (permissions.can_add_users_to_a_course(request.user) or request.user.is_superuser):
+        if not (request.user.is_teacher or request.user.is_superuser):
             return response.forbidden(description="Only teachers and administrators are allowed to request all user \
                                        data.")
```
Larspolo: doesnt this create a lot of whitespace that is also send with? I would prefer concatenating the string (and also use single quotes)

src/django/test/test_utils.py
```diff
@@ -140,7 +140,7 @@ def logging_in(obj, username, password, status=200):
     result = obj.client.post(reverse('token_obtain_pair'),
                              json.dumps({'username': username, 'password': password}),
                              content_type='application/json')
-    obj.assertEquals(result.status_code, status)
+    obj.assertEquals(result.status_code, status, "Failed response was: " + str(result.json()))
```
Larspolo: single quotes are preferred

src/django/test/test_utils.py
```diff
@@ -156,7 +156,7 @@ def api_get_call(obj, url, login, status=200, params={}):
     """
     result = obj.client.get(url, params,
                             HTTP_AUTHORIZATION='Bearer {0}'.format(login.data['access']))
-    obj.assertEquals(result.status_code, status)
+    obj.assertEquals(result.status_code, status, "Failed response was: " + str(result.json()))
```
Larspolo: single quotes are preferred

src/vue/src/store/modules/permissions.js
```diff
@@ -12,16 +12,19 @@ const getters = {
...
+        } else if ('cID' in routeParams && types.COURSE_LEVEL_PERMISSIONS.has(permission)) {
             return permissions['course' + routeParams.cID][permission]
-        } else if ('aID' in routeParams) {
+        } else if ('aID' in routeParams && types.ASSIGNMENT_LEVEL_PERMISSIONS.has(permission)) {
             return permissions['assignment' + routeParams.aID][permission]
-        } else {
+        } else if (types.GENERAL_LEVEL_PERMISSIONS.has(permission)) {
             return permissions['general'][permission]
+        } else {
+            throw Error('Permission ' + permission + ' is not a valid permission for this context.')
```
Larspolo: "for this context"? For what context whould it have valid permissions? I think all are covered here right?

src/django/VLE/views/entry.py
Rickyboy320: This is not possible due to the succeeding elif statements.

src/django/VLE/views/role.py
Rickyboy320: It's still technically relevant. I'd love to see a redesign of this function though, it looks like a mess.

src/vue/src/store/modules/permissions.js
Rickyboy320: I guess the message isn't exactly descriptive.
The point of this error is that the permission is a permission that is not covered by the context. Imagine an assignment permission, but not having the aID in the route params. We cannot resolve this permission, so it will execute this clause, and throw the error.

src/django/VLE/views/entry.py
Rickyboy320: No, entries should still be able to be updated even after being due (by the teacher/TA), though _editing_ entries should not be allowed.

src/django/VLE/views/group.py
Rickyboy320: Because it's an _or_ statement. Using check_permission for each of them would cause it to be an 'and' statement.

src/django/VLE/views/journal.py
Rickyboy320: I feel this deducts a lot from readability of the check.
The user should _not_ have the permission can_have_journal and should _not_ have the permission can_view_assignment_journals for the if statement to execute. Using parentheses and _or_ obscures this goal.

src/django/test/test_get_apis.py
```diff
@@ -212,7 +212,7 @@ def test_get_nodes(self):
         login = test.logging_in(self, student_user, student_pass)
         response = test.api_get_call(self, '/nodes/', login, params={'journal_id': journal.pk})
         result = response.json()
-        self.assertEquals(len(result['nodes']), 4)
+        self.assertEquals(len(result['nodes']), 5)
```
Rickyboy320: You might think me mad for changing this. There actually was a bug with the previous way add nodes were handled. If an assignment didn't have a due_date (which is perfectly acceptable, since the field can be NULL), add nodes were never displayed. This is now fixed due to the improved assignment.is_due() function.

src/django/VLE/models.py
```diff
@@ -123,13 +123,13 @@ class User(AbstractUser):
...
 
-    def check_permission(self, permission, obj=None):
+    def check_permission(self, permission, obj=None, message=None):
         """
         Throws a VLEPermissionError when the user does not have the specified permission, as defined by
         has_permission.
         """
         if not self.has_permission(permission, obj):
-            raise VLEPermissionError(permission)
+            raise VLEPermissionError(permission, message)
```
Rickyboy320: Added this for better description in certain cases (where the permission wasn't exactly descriptive of what was actually happening). Should be used sparingly, as consistency throughout messages is preferred.

src/django/VLE/views/entry.py
```diff
@@ -147,20 +142,19 @@ def partial_update(self, request, *args, **kwargs):
...
+            request.user.check_permission('can_grade', assignment)
+
+            if isinstance(grade, (int, float)) or grade < 0:
+                return response.bad_request('Grade must be greater than or equal to zero.')
+
             entry.grade = grade
 
-        if published is not None and not request.user.has_permission('can_publish_grades', assignment):
-            return response.forbidden('You cannot publish entries.')
+        if published:
```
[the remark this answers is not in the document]
Rickyboy320: Good catch, guess I was a little tired :)

src/django/VLE/views/journal.py
Larspolo: Fair enough, herever there is still an issue atm. If a user is not the ownser, but does have the `can_have_journal` it can view this journal. I think it this should be changed to just a `request.user.check_permission('can_view_assignment_journals', journal.assignment)`

Larspolo: image
Logged in as Student (id=1) and got journal info when going to http://localhost:8080/Home/Course/1/Assignment/1/Journal/2

src/django/VLE/views/comment.py
```diff
@@ -89,10 +89,7 @@ def create(self, request):
         assignment = journal.assignment
 
         request.user.check_permission('can_comment', assignment)
-        if request.user != journal.user:
-            request.user.check_permission('can_view_assignment_journals', assignment)
-
-        published = published or not request.user.has_permission('can_grade', assignment)
```
Rickyboy320: Where did this line go?

src/django/VLE/utils/generic_utils.py
```diff
@@ -223,21 +223,23 @@ def update_presets(assignment, presets, template_map):
...
         exists = 'id' in preset
+        id, type, description, deadline, target, template = \
+            optional_params(preset, 'id', 'type', 'description', 'deadline', 'target', 'template')
 
         if exists:
             preset_node = PresetNode.objects.get(pk=preset['id'])
         else:
             preset_node = PresetNode(format=format)
 
         type_changed = preset_node.type != preset['type']
```
Rickyboy320: You forgot to change this one :)

src/django/VLE/views/email.py
```diff
@@ -61,8 +61,9 @@ def recover_password(request):
 
     user = User.objects.get(username=request.data['username'])
 
+    token, = utils.required_params(request.data, 'token')
```
Rickyboy320: Unless you changed this in the frontend as well, this should be recovery_token

src/django/VLE/views/comment.py
engelhamer: @larspolo don't forget this one.

Larspolo: Its fixed
image
You are watching old code

engelhamer: Reckon I did, got lost in Githubs diff 🤯

PR #450, opened 2018-10-15
Changed api request to filter on the right journals

### LINKED ISSUE #447: Hard reloading NonStudentJournal page loads teachers' journals
**Describe the bug**
When coming from the Journal overview page, the next/previous journal controls work as expected. However, when reloading the page, the teachers' journals are loaded alongside the expected journals, which allows you to get into the teachers' journal through the UI (while not dangerous, it is not preferable).

**To Reproduce**
Easiest to work with a course where there is only one user and the teacher.
1. Open the users' journal as the teacher.
2. Note that there are no controls (expected).
3. Hard refresh the page.
4. Note that the controls appear.
5. Click on the controls, and end up at the teacher's journal page.

JowehL:
## Description
Get api for retrieving journals from an assignment is now also filtering on users from a course with the permission to have a journal.

## Summary of changes
- Fixes #447 and removed students from the course


src/django/VLE/views/journal.py
```diff
@@ -44,11 +45,14 @@ def list(self, request):
             return response.unauthorized()
 
         assignment = Assignment.objects.get(pk=request.query_params['assignment_id'])
+        course = Course.objects.get(pk=request.query_params['course_id'])
 
         if not permissions.has_assignment_permission(request.user, assignment, 'can_view_assignment_journals'):
             return response.forbidden('You are not allowed to view assignment participants.')
 
-        queryset = assignment.journal_set.all()
+        ids_in_course = course.participation_set.all().values('user__id').filter(role__can_have_journal=True)
```
Rickyboy320: all() can probably be omitted.

src/django/test/test_get_apis.py
```diff
@@ -190,15 +190,28 @@ def test_assignment_journals(self):
...
+                          status=403,
+                          params={'course_id': course.pk, 'assignment_id': assignment.pk})
+        test.api_get_call(self,
+                          '/journals/',
+                          login,
+                          status=404,
+                          params={'course_id': course.pk, 'assignment_id': self.not_found_pk})
+        test.test_unauthorized_api_get_call(self,
+                                            '/journals/',
+                                            params={'course_id': course.pk, 'assignment_id': self.not_found_pk})
```
Rickyboy320: Could you please add a test where either course_id or assignment_id are not given (and assert that it doesn't return an internal server error)

src/django/VLE/views/journal.py
```diff
@@ -44,11 +45,14 @@ def list(self, request):
             return response.unauthorized()
 
         assignment = Assignment.objects.get(pk=request.query_params['assignment_id'])
+        course = Course.objects.get(pk=request.query_params['course_id'])
```
Rickyboy320: This way of indexing could possibly lead to an internal server error.

src/django/VLE/views/journal.py
```diff
@@ -43,12 +44,16 @@ def list(self, request):
         if not request.user.is_authenticated:
             return response.unauthorized()
 
-        assignment = Assignment.objects.get(pk=request.query_params['assignment_id'])
+        assignment_id, course_id = utils.required_params(request.query_params, "assignment_id", "course_id")
```
Larspolo: For id's the `required_typed_params` is preferred
`utils.required_params(request.query_params, (int, 'assignment_id'), (int, 'course_id'))`
Also single quotes iso double quotes are preferred

PR #458, opened 2018-10-19
Bugfix/security stuff

Larspolo:
## Description
Some more security issue fixes

## Summary of changes
Fixes:
participation.py:155:0: Shouldn't you check for the
`can_edit_course_user_group` permission?
email.py:116:0: Why do you verify emails if this never checked before
sending emails to a user?
email.py:52:0: Why is old style (python2) formatting used here?
user.py:195:0: This seems for something that lti.py should do. Maybe
have function that checks this and generates the correct error responses
(or handle the exception by registering an exception handler)?
user.py:324:0: You are missing comments, those are user data too.
user.py:434:0: You should check this before all other checks
assignment.py:214:0: Permissions should be checked in the route as this
common for the entire code base.
assignment.py:329:0: This function duplicates the function `publish`.


src/django/VLE/utils/error_handling.py
```diff
@@ -17,6 +18,11 @@ class VLEProgrammingError(Exception):
     pass
 
 
+class VLEVerifiedEmail(Exception):
```
Rickyboy320: For naming consistency (and clarity), please suffix with Error.

src/django/VLE/views/assignment.py
```diff
@@ -316,15 +315,7 @@ def published_state(self, request, *args, **kwargs):
...
-                if journal.sourcedid is not None and journal.grade_url is not None:
-                    payload = lti_grade.replace_result(journal)
-                else:
-                    payload = dict()
-
-            return payload
-        else:
-            return False
+        utils.publish_all_assignment_grades(assignment, published)
+        for journal in Journal.objects.filter(assignment=assignment):
```
Rickyboy320: assignment.journal_set would be nicer.

src/django/VLE/views/email.py
```diff
@@ -57,21 +57,22 @@ def recover_password(request):
...
 
-    validators.validate_password(request.data['new_password'])
+    validators.validate_password(new_password)
 
-    user.set_password(request.data['new_password'])
+    user.set_password(new_password)
     user.save()
 
-    return response.success(description='Succesfully changed the password, please login.')
+    return response.success(description='Succesfully changed the password, you can login now.')
```
Rickyboy320: Successfully*

src/django/VLE/views/user.py
```diff
@@ -385,16 +373,14 @@ def upload(self, request):
...
             return response.unauthorized()
 
-        if not (request.FILES and 'file' in request.FILES):
-            return response.bad_request('No accompanying file found in the request.')
-
         assignment_id, content_id = utils.required_params(request.POST, 'assignment_id', 'content_id')
-
         assignment = Assignment.objects.get(pk=assignment_id)
 
         if not Assignment.objects.filter(courses__users=request.user, pk=assignment.pk).exists():
```
Rickyboy320: Can now become user.check_participation (missed in a different PR)

src/django/VLE/views/email.py
Larspolo: image
Fixed everywhere

src/django/VLE/utils/error_handling.py
```diff
@@ -17,6 +18,11 @@ class VLEProgrammingError(Exception):
     pass
 
 
+class VLEVerifiedEmailError(Exception):
+    def __init__(self, message='You need to verify your email before an email can be send to this account.'):
```
engelhamer: `sent` instead of `send`.

src/django/VLE/utils/error_handling.py
```diff
@@ -38,20 +44,37 @@ def __call__(self, request):
...
+        elif isinstance(exception, VLEProgrammingError):
+            return response.internal_server_error(str(exception))
+        elif isinstance(exception, SMTPAuthenticationError):
+            return response.internal_server_error(
+                'Mailserver is not configured correctly, please contact a server admin.')
+
+        # LTI exceptions
+        elif isinstance(exception, jwt.exceptions.ExpiredSignatureError):
+            return response.forbidden(
+                'The canvas link has expired, 15 minutes have passed. Please retry from your LTI instance.')
```
engelhamer: Replace canvas with generic LTI.

src/django/VLE/views/email.py
```diff
@@ -60,21 +60,22 @@ def recover_password(request):
...
 
-    validators.validate_password(request.data['new_password'])
+    validators.validate_password(new_password)
 
-    user.set_password(request.data['new_password'])
+    user.set_password(new_password)
     user.save()
 
-    return response.success(description='Succesfully changed the password, please login.')
+    return response.success(description='Successfully changed the password, you can login now.')
```
engelhamer: `you can now log in`

src/vue/src/components/profile/NotificationCard.vue
```diff
@@ -36,15 +36,15 @@ export default {
             userAPI.update(0, {grade_notifications: isActive})
                 .then(user => {
                     this.$store.commit('user/SET_GRADE_NOTIFICATION', user.grade_notifications)
-                    this.$toasted.success('Grade notification setting updated succesfully.')
+                    this.$toasted.success('Grade notification setting updated Successfully.')
```
engelhamer: `Successfully` should not be capitalised here. You cannot always rely on a find-all-then-replace query 😉 .

src/vue/src/components/profile/NotificationCard.vue
```diff
@@ -36,15 +36,15 @@ export default {
...
                 })
                 .catch(error => { this.$toasted.error(error.response.data.description) })
         },
         getCommentNotification (isActive) {
             userAPI.update(0, {comment_notifications: isActive})
                 .then(user => {
                     this.$store.commit('user/SET_COMMENT_NOTIFICATION', user.comment_notifications)
-                    this.$toasted.success('Comment notification setting updated succesfully.')
+                    this.$toasted.success('Comment notification setting updated Successfully.')
```
engelhamer: Same goes for this one.

src/vue/src/store/modules/user.js
```diff
@@ -94,7 +94,7 @@ const actions = {
                 commit(types.SET_JWT, response.data)
 
                 dispatch('populateStore').then(response => {
-                    resolve('JWT and store are set succesfully.')
+                    resolve('JWT and store are set Successfully.')
```
engelhamer: And again.

src/vue/src/store/modules/user.js
```diff
@@ -123,7 +123,7 @@ const actions = {
                             .then(_ => { resolve() })
                             .catch(error => { reject(error) })
                     } else {
-                        resolve('JWT refreshed succesfully, store was already populated.')
+                        resolve('JWT refreshed Successfully, store was already populated.')
```
engelhamer: You Will Not Believe How This Website Handles Capitalisation! This Seems Like Clickbait.

src/vue/src/store/modules/user.js
```diff
@@ -137,7 +137,7 @@ const actions = {
         return new Promise((resolve, reject) => {
             connection.conn.get('/users/0/').then(response => {
                 commit(types.HYDRATE_USER, response.data)
-                resolve('Store is populated succesfully')
+                resolve('Store is populated Successfully')
```
engelhamer: Alright I'll stop.

src/vue/src/views/UserRoleConfiguration.vue
```diff
@@ -173,7 +173,7 @@ export default {
                 .then(_ => {
                     this.originalRoleConfig = this.deepCopyRoles(this.roleConfig)
                     this.defaultRoles = Array.from(this.roles)
-                    this.$toasted.success('Course roles succesfully updated.')
+                    this.$toasted.success('Course roles Successfully updated.')
```
engelhamer: Oh no.

src/vue/src/views/UserRoleConfiguration.vue
```diff
@@ -194,7 +194,7 @@ export default {
                         .then(_ => {
                             this.deleteRoleLocalConfig(role)
                             this.deleteRoleServerLoadedConfig(role)
-                            this.$toasted.success('Role deleted succesfully!')
+                            this.$toasted.success('Role deleted Successfully!')
```
engelhamer: Last one.

src/django/VLE/views/entry.py
```diff
@@ -127,17 +127,18 @@ def partial_update(self, request, *args, **kwargs):
...
+        published, content_list = utils.optional_params(request.data, 'published', 'content')
 
         entry_id, = utils.required_typed_params(kwargs, (int, 'pk'))
         entry = Entry.objects.get(pk=entry_id)
         journal = entry.node.journal
         assignment = journal.assignment
 
-        if grade is not None:
+        if 'grade' in request.data:
+            grade, = utils.required_typed_params(request.data, (float, 'grade'))
```
Rickyboy320: Might it be worth to create a typed_optional_params instead?

src/django/VLE/views/user.py
```diff
@@ -72,7 +72,6 @@ def retrieve(self, request, pk):
 
         return response.success({'user': serializer.data})
 
-    @permission_classes((AllowAny, ))
```
Rickyboy320: Are you sure this is correct? I'm not sure...

src/django/VLE/views/user.py
```diff
@@ -414,4 +398,10 @@ def set_profile_picture(self, request):
...
         request.user.save()
 
-        return response.success(description='Succesfully updated profile picture')
+        return response.success(description='Successfully updated profile picture')
+
+    def get_permissions(self):
+        if self.request.path == '/users/' and self.request.method == 'POST':
+            return [AllowAny()]
+        else:
+            return [permission() for permission in self.permission_classes]
```
Rickyboy320: Why handle this special case like this? Seems to me a simple annotation is more straightforward (and more localized and therefore clearer)

src/django/VLE/views/entry.py
Larspolo: Then this check still needs to be done, maybe later when its really needed

src/django/VLE/views/user.py
Larspolo: Yes, registration broke, its now fixed again

src/django/VLE/views/user.py
Larspolo: Viewsets need to be done this way

src/django/VLE/views/user.py
Larspolo: Added a registration simple check so that it wont happen again

PR #460, opened 2018-10-19
Feature/is authorized

Larspolo:
## Description
Every request is now checked on is_authenticated by defualt. Some can still


src/django/VLE/views/user.py
```diff
@@ -60,8 +58,7 @@ def retrieve(self, request, pk):
         On success:
             success -- with the user data
         """
-        if not request.user.is_authenticated:
-            return response.unauthorized()
+        print(request.user)
```
Rickyboy320: Lingering print

Larspolo:
When going through all the queries to remove `is_authorized`, these were the only one without that check

PR #466, opened 2018-10-21
Enhancement/async pictures

### LINKED ISSUE #375: asynchronous profile pictures

Airgetfrog:
## Description
Added filesystem to profile pictures and async requests to views with those pictures.

## Summary of changes
- Fixes #375 

## Images
Test it yourself, it mostly concerns the loading.


src/django/VLE/views/user.py
```diff
@@ -377,29 +377,54 @@ def upload(self, request):
...
+        pk -- user ID
+
+        Returns
+        On failure:
+            unauthorized -- when the user is not logged in
+            bad_request -- when the file was not found
+        On success:
+            success -- bytebuffer with image
+        """
+        if not request.user.is_authenticated:
```
mjvkeulen: No longer needed

src/django/VLE/views/user.py
```diff
@@ -377,29 +377,54 @@ def upload(self, request):
...
+            return response.unauthorized()
+        if int(pk) == 0:
+            pk = request.user.id
+
+        user = User.objects.get(pk=pk)
```
mjvkeulen: Would a check be desired to see if the user relate to one another through some course?

src/vue/src/components/assets/ProfilePicture.vue
```diff
@@ -0,0 +1,35 @@
...
+        }
+    },
+    mounted () {
+        if (!this.user.lti_image) {
+            userAPI.downloadProfilePicture(this.user.id)
+                .then(response => {
+                    var reader = new FileReader()
+                    var vm = this
+                    reader.onload = () => {
+                        vm.profilePicture = reader.result
```
libre-man: You can simply use `this` here as a lambda doesn't rebind `this`.

src/django/VLE/views/user.py
Airgetfrog: How would one go about checking this? Checking the user's courses and their participations?

src/vue/src/components/assets/ProfilePicture.vue
```diff
@@ -0,0 +1,35 @@
...
+                    }
+                    reader.readAsDataURL(new Blob([response.data], { type: response.headers['content-type'] }))
+                })
+                .catch(error => {
+                    this.$toasted.error(error.response.data.description)
```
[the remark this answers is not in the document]
Airgetfrog: Is there an intermediate placeholder that could be copied/implemented now or should we do it in one go later? Separation is a good call, will implement.

src/django/VLE/migrations/0007_auto_20181018_1150.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 2.1.2 on 2018-10-18 11:50
+
+from django.db import migrations, models
+
+import VLE.utils.file_handling
+import VLE.utils.storage
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
Larspolo: This can all be done in one merge file, if you want any help, ask me or Rick

src/django/VLE/settings/base.py
```diff
@@ -16,8 +16,8 @@
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-MEDIA_URL = 'media/'
-MEDIA_ROOT = os.path.join(BASE_DIR, MEDIA_URL)
+MEDIA_URL = '/media/'
+MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
```
Larspolo: why this change?

src/django/VLE/utils/file_handling.py
```diff
@@ -12,10 +12,15 @@
...
     """Upload user files into their respective directories. Following MEDIA_ROOT/uID/aID/<file>
 
-    Uploaded files not part of an entry yet, and are treated as temporary untill linked to an entry."""
+    Uploaded files not part of an entry yet, and are treated as temporary until linked to an entry."""
     return str(instance.author.id) + '/' + str(instance.assignment.id) + '/' + filename
 
 
+def get_profile_picture_path(instance, filename):
+    """Upload user profile pictures into their respective directories. Following MEDIA_ROOT/uID/profile_picture"""
+    return str(instance.id) + '/' + 'profile_picture'
```
Larspolo: can be shortened, I would suggest using a `.format` as well. Also change this for the function above

src/django/VLE/utils/storage.py
```diff
@@ -0,0 +1,31 @@
...
+        proposed solution was to override the save method on the model
+        like so (from https://code.djangoproject.com/ticket/11663):
+
+        def save(self, *args, **kwargs):
+            try:
+                this = MyModelName.objects.get(id=self.id)
+                if this.MyImageFieldName != self.MyImageFieldName:
+                    this.MyImageFieldName.delete()
+            except: pass
+            super(MyModelName, self).save(*args, **kwargs)
```
Larspolo: why this save function?

src/django/VLE/validators.py
```diff
@@ -13,12 +13,15 @@
...
-def validate_profile_picture_base64(urlData):
-    """Checks if the original size does not exceed 2MB AFTER encoding."""
-    if len(urlData) > settings.USER_MAX_FILE_SIZE_BYTES * 1.37:
+def validate_profile_picture(urlData):
+    """Checks if size does not exceed 2MB. Checks if picture is a supported image type."""
+    if urlData.size > settings.USER_MAX_FILE_SIZE_BYTES:
         raise ValidationError("Max size of file is {} Bytes".format(settings.USER_MAX_FILE_SIZE_BYTES))
 
+    main, sub = urlData.content_type.split('/')
+    if not (main == 'image' and sub in ['jpeg', 'pjpeg', 'gif', 'png']):
```
Larspolo: maybe add `jpg` as well

src/vue/src/api/auth.js
```diff
@@ -144,6 +144,9 @@ export default {
     downloadFile (url, data, noRedirect = false) {
         return validatedSend(connection.connFile.get, improveUrl(url, data), null, noRedirect)
     },
+    uploadProfilePicture (url, data, noRedirect = false) {
+        return validatedSend(connection.connFile.post, improveUrl(url), data, noRedirect)
+    },
```
Larspolo: Url can be hardcoded here

Larspolo: Or just remove it, as this is the same functions as `uploadFile`

src/vue/src/api/user.js
```diff
@@ -40,14 +40,14 @@ export default {
         return auth.uploadFile('/users/upload/', data)
     },
 
-    /* Get user file. */
-    getUserFile (fileName, authorUID) {
-        return auth.downloadFile('/get_user_file/' + fileName + '/' + authorUID + '/')
+    /* Upload an image. */
+    uploadProfilePicture (data) {
+        return auth.uploadProfilePicture('/users/upload_profile_picture/', data)
```
Larspolo: if `uploadProfilePicture` gets removed, you can change this to `uploadFile` and it would still work

src/django/VLE/settings/base.py
Airgetfrog: Vestigial change, oops!

src/django/VLE/utils/storage.py
Airgetfrog: This is part of the docstring in the original SO answer, not actual code. As for why it's in there, beats me.

src/django/VLE/validators.py
Airgetfrog: I believe `image/jpg` is not an actual mime/content-type, only `image.jpeg` [as given by the standard](https://www.w3.org/Graphics/JPEG/).

mjvkeulen: What validation is set on the default image field? Is this validation not extra?

mjvkeulen: Hm I couldnt figure out the exact error raised. Pillow is definately used to acquire the width/height, and I think the file is simply returned as None if this fails. Maybe you could double check this?

src/vue/src/components/assets/ProfilePicture.vue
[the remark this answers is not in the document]
mjvkeulen: A simple parse is in #465

src/django/VLE/views/user.py
mjvkeulen: Ye afraid so, dont see a fast way to do so?

src/django/VLE/migrations/0008_merge_20181021_1332.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0006_can_view_assignment_journals_rename'),
+        ('VLE', '0007_profile_picture_separation'),
+    ]
+
+    operations = [
```
mjvkeulen: cant this be deleted? Or am I missing something

src/django/VLE/validators.py
Airgetfrog: From what I remember it was either based on trying to open it with a certain library or based on file extension. Wasn't sure about it hence the explicit check. I suppose we can test it by removing this sometime.

src/django/VLE/utils/storage.py
mjvkeulen: I feel overwriting the old profile picture on save would be desired behaviour. Will result in us having no floating profile pictures and only storing active ones.

src/vue/src/components/assets/ProfilePicture.vue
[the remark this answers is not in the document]
mjvkeulen: I personally feel adding the reject hook to the then would be cleaner. This separates an error locally from an API call entirely.

src/django/VLE/migrations/0008_merge_20181021_1332.py
Airgetfrog: Can it be? I thought it was for django to know that the migration is safe, since it's generated after merging. I'll delete it if that isn't the case.

src/vue/src/components/course/CourseParticipantCard.vue
```diff
@@ -134,7 +135,8 @@ export default {
         this.selectedGroup = this.group
     },
     components: {
-        icon
+        icon,
+        profilePicture
```
mjvkeulen: I didnt even know this seperated camelcase to snake, could use this everywhere. (One day)

src/django/VLE/utils/storage.py
Airgetfrog: This is what happens currently (the old one is deleted in this call). I think Lars was referring to the commented function in the docstring.

src/django/VLE/migrations/0008_merge_20181021_1332.py
Larspolo: It is, you could also rename the files to a synchronous order iso both 0007, and remove the merge file

src/vue/src/store/modules/user.js
```diff
@@ -137,10 +137,25 @@ const actions = {
...
+                } else {
+                    connection.connFile.get('/users/0/download_profile_picture/').then(response => {
+                        var reader = new FileReader()
+                        reader.onload = () => {
+                            commit(types.SET_PROFILE_PICTURE, reader.result)
+                            resolve('Store is populated successfully')
+                        }
+                        reader.readAsDataURL(new Blob([response.data], { type: response.headers['content-type'] }))
+                    }, error => {
+                        Vue.toasted.error(error.response.data.description)
```
mjvkeulen: Same issue here, the error cannot be parsed.

src/django/VLE/migrations/0008_merge_20181021_1332.py
Rickyboy320: That's probably the most straightforward method. Keeps the history somewhat clean as well, making future disputes easier to solve.

Remove the merge file, make your 0007 depend on 0006 (since 0006 has been established, and 0007 is 'new')

src/django/VLE/utils/file_handling.py
```diff
@@ -12,8 +12,13 @@
...
 
-    Uploaded files not part of an entry yet, and are treated as temporary untill linked to an entry."""
-    return str(instance.author.id) + '/' + str(instance.assignment.id) + '/' + filename
+    Uploaded files not part of an entry yet, and are treated as temporary until linked to an entry."""
+    return '{}/{}/{}'.format(str(instance.author.id), str(instance.assignment.id), filename)
+
+
+def get_profile_picture_path(instance, filename):
+    """Upload user profile pictures into their respective directories. Following MEDIA_ROOT/uID/profile_picture"""
+    return '{}/profile_picture'.format(str(instance.id))
```
Rickyboy320: ~Assuming instance is the same object as above, shouldn't this be instance.author.id, since the documentation explicitly states it should be an user id.~
Maybe 'instance' should be more descriptive what type of instance it is. I see now that it is an instance of User, but in the method above it is UserFile.

src/django/VLE/views/user.py
Airgetfrog: Now that I think about it, isn't this an expensive operation to do for every profile picture load?

src/django/VLE/validators.py
Airgetfrog: I actually can't seem to find any validation at all when testing. [According to an old SO answer](https://stackoverflow.com/a/7901240) the builtin validation is only used when set via form and not when set manually. Given that it seems best to keep this validation in. We might also want to consider adding some extra checks for malicious files (virus checking, magic numbers, path checking, nosniff headers and more are all possible based on some quick googling) on all file handling in a separate PR in the future.

mjvkeulen: File checking is a slippy slope, this is resolved for me!

src/django/VLE/views/user.py
mjvkeulen: If we don't do this profile pictures are public, I think we should discuss this at our next meeting.

src/vue/src/components/profile/ProfileData.vue
```diff
@@ -89,12 +92,19 @@ export default {
...
-                    this.$store.commit('user/SET_PROFILE_PICTURE', dataURL)
-                    this.profileImageDataURL = dataURL
-                    this.$refs['cropperModal'].hide()
+                    var reader = new FileReader()
+                    reader.onload = () => {
+                        this.$store.commit('user/SET_PROFILE_PICTURE', reader.result)
+                        this.profileImageDataURL = reader.result
+                        this.$refs['cropperModal'].hide()
+                    }
+                    reader.readAsDataURL(blob)
```
mjvkeulen: Would we want a try catch for local errors?

src/vue/src/store/modules/user.js
```diff
@@ -148,10 +148,26 @@ const actions = {
...
+                    connection.connFile.get('/users/0/download_profile_picture/').then(response => {
+                        var reader = new FileReader()
+                        reader.onload = () => {
+                            commit(types.SET_PROFILE_PICTURE, reader.result)
+                            resolve('Store is populated successfully')
+                        }
+                        reader.readAsDataURL(new Blob([response.data], { type: response.headers['content-type'] }))
+                    }, error => {
+                        let enc = new TextDecoder('utf-8')
+                        Vue.toasted.error(JSON.parse(enc.decode(error.response.data)).description)
```
mjvkeulen: This should function should now be a defaul in utils.

src/vue/src/components/course/CourseParticipantCard.vue
engelhamer: Nice discovery :+1:

engelhamer: Nope, there's manual conversions to dashed-names all over.

mjvkeulen:
This has been postponed for now due to different priorities.

PR #473, opened 2018-10-25
Escape toast

C-Bouthoorn:
## Description
Using a malicious URL, everyone is able to inject `<script>` tags in toast messages, which might leak personal information. By using a function to escape all user-input before passing it to toasted, this can be prevented.

## Summary of changes
- Added said function.

### DIFF
--- a/src/vue/src/utils/escape.js
+++ b/src/vue/src/utils/escape.js
@@ -13,4 +13,3 @@ export function escapeHtml (rawhtml) {
 
     return rawhtml.replace(/[&<>"']/g, (match) => specialCharacterMap[match])
 }
-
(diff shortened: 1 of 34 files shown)


src/vue/src/utils/escape.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Replace all characters which might be used to inject HTML with their corresponding special character.
+ * If the rawhtml is not a string, it won't get changed.
+ */
+export function escapeHtml (rawhtml) {
```
engelhamer: I suggest you make this the default export, as is the main purpose of this file. That would also allow for the curly braces to be omitted when importing this function.

C-Bouthoorn: Should be fixed now

engelhamer:
We have implemented a more generic fix for this problem in #480. Closing this now.

PR #477, opened 2018-10-27
Bugfix/entry create date

Larspolo:
## Description
Entries got the creation date of when the server was booted, now its the real creation date. Also small change in comments, last_edited is now set by default, done by django


src/vue/src/components/journal/CommentCard.vue
```diff
@@ -20,10 +20,10 @@
                         <div v-html="comment.text"/>
                         <hr class="full-width"/>
                         <b>{{ comment.author.first_name + ' ' + comment.author.last_name }}</b>
-                        <span v-if="comment.published && !comment.last_edited" class="timestamp">
+                        <span v-if="comment.published && $root.beautifyDate(comment.last_edited) == $root.beautifyDate(comment.timestamp)" class="timestamp">
```
engelhamer: I think you should use `===`.

src/django/VLE/migrations/0010_auto_20181027_1403.py
```diff
@@ -0,0 +1,51 @@
...
+        migrations.AlterField(
+            model_name='comment',
+            name='last_edited',
+            field=models.DateTimeField(default=django.utils.timezone.now),
+            preserve_default=False,
+        ),
+        migrations.AlterField(
+            model_name='entry',
+            name='last_edited',
+            field=models.DateTimeField(default=django.utils.timezone.now),
```
Larspolo: Defaults had to be set because older comments may have this field set to None. If this is the case, it now sets it to the time that this migration is applied (or for future nodes, it gets set to the time of uploading)

src/django/VLE/models.py
```diff
@@ -633,24 +640,26 @@ class Entry(models.Model):
...
+    creation_date = models.DateTimeField(editable=False)
+    last_edited = models.DateTimeField()
 
     def is_due(self):
         return (self.node.preset and self.node.preset.is_due()) or self.node.journal.assignment.is_due()
 
+    def save(self, *args, **kwargs):
+        if not self.pk:
+            self.creation_date = timezone.now()
+        self.last_edited = timezone.now()
```
Larspolo: I would like to apply move the factory to these save functions as well. But if we do that, we may want to split all the models in their own file. What are your thoughts on this?

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -20,10 +20,10 @@
                         <div v-html="comment.text"/>
                         <hr class="full-width"/>
                         <b>{{ comment.author.first_name + ' ' + comment.author.last_name }}</b>
-                        <span v-if="comment.published && !comment.last_edited" class="timestamp">
-                            {{ $root.beautifyDate(comment.timestamp) }}<br/>
+                        <span v-if="comment.published && $root.beautifyDate(comment.last_edited) === $root.beautifyDate(comment.creation_date)" class="timestamp">
```
Larspolo: I use beautify date because the `last_edited` and `creation_date` can be off by some milliseconds

src/django/VLE/models.py
engelhamer: I have no objection to either method.

src/vue/src/components/journal/CommentCard.vue
engelhamer: Funny solution.

PR #480, opened 2018-10-28
Bugfix/xss

mjvkeulen:
## Description
Reworked front-end API calls to include a connection arguments object. Allowing for the removal of a lot of duplicate rejection and success handling message display. Allowing for the sanitization of success and or error messages in a concentrated location.

## Summary of changes
Each API call accepts a connection argument object, which expects the following keys and values as default:
```
const DEFAULT_CONN_ARGS = {
    redirect: true,
    customSuccessToast: false,
    responseSuccessToast: false,
    customErrorToast: false
}
```
This object defines how success and error responses are handled and toasted by default.

Changes can be made by overwriting the DEFAULT_CONN_ARGS keys in an API call.
Handled errors are redirected by default when present in ERRORS_TO_REDIRECT unless redirect set to false.
Handled errors messages default to: response.data.description, unless customErrorToast set.
Handled successes do not redirect or display a message unless:

- responseSuccessToast is set, toasting the response description
- customSuccessToast is set, toasting the given message.


src/vue/src/api/user.js
```diff
@@ -1,62 +1,46 @@
...
             entry_id: entryID,
             node_id: nodeID,
             content_id: contentID
-        })
+        }, connArgs)
     },
 
-    GDPR (id = 0) {
-        return auth.downloadFile('users/' + id + '/GDPR/')
+    GDPR (id, connArgs = auth.DEFAULT_CONN_ARGS) {
```
Larspolo: image

image

throttlers work with detail iso description, this should now also be caught

Larspolo: Also it promps with the error creation file locally, but this time it is a server error, as this was due to requesting it too many times

src/vue/src/components/assets/PdfDisplay.vue
```diff
@@ -115,7 +115,7 @@ export default {
             this.$toasted.error('Password handling is not implemented.')
         },
         error (err) {
-            this.$toasted.error(err)
+            this.$toasted.error(sanitization.escapeHtml(err))
```
Larspolo: why isnt this handled in the auth.js?

src/vue/src/components/course/CourseGroupModal.vue
```diff
@@ -52,17 +52,15 @@ export default {
     },
     methods: {
         createUserGroup () {
-            groupApi.create({
+            groupAPI.create({
                 name: this.form.groupName,
                 course_id: this.cID,
                 lti_id: this.lti_id
-            })
+            }, {customSuccessToast: 'Successfully created group.'})
```
Larspolo: This seems a bit weird. This is on 1 row and the previous object is on multiple rows, maybe this can be extended a few rows as well

src/vue/src/components/group/GroupCard.vue
```diff
@@ -41,22 +41,17 @@ export default {
             groupAPI.update(this.cID, {
                 old_group_name: this.group,
                 new_group_name: this.form.newGroupName
-            })
+            }, {customSuccessToast: 'Successfully updated the group.'})
```
Larspolo: same here

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -106,9 +106,8 @@ export default {
         }
     },
     created () {
-        assignmentAPI.get(this.aID)
+        assignmentAPI.get(this.aID, null)
```
Larspolo: Why not null by default?

src/vue/src/utils/sanitization.js
```diff
@@ -0,0 +1,19 @@
+var entityMap = {
```
Larspolo: var -> cost?

src/vue/src/views/Home.vue
```diff
@@ -77,15 +77,13 @@ export default {
     created () {
         this.loadCourses()
 
-        assignmentAPI.getUpcoming()
+        assignmentAPI.getUpcoming(null)
```
Larspolo: Why not null by default?

src/vue/src/components/journal/JournalNonStudent.vue
mjvkeulen: As in javascript one can't access function parameters by keyword, setting multiple default parameters makes it so that accessing any but the first default requires arguments for all up to that point. As such, I felt it would be most consistent to ensure the connection argument was the final argument to an API call, making any other arguments explicit. 
I did consider including data as an optional connection argument, defaulting to null. However data is inconsistently used. For example, sometimes part of the data argument is used to construct the url, or sometimes the data argument is filtered and only partly used. To take this approach an even more substantial rework is required, possibly even allowing to get rid of the dedicated api call files all together as each call is then made generic.

src/vue/src/utils/sanitization.js
mjvkeulen: Sorry, could you elaborate a little?

src/vue/src/api/auth.js
```diff
@@ -109,7 +157,7 @@ export default {
 
     /* Change password. */
     changePassword (newPassword, oldPassword) {
-        return this.update('users/password', { new_password: newPassword, old_password: oldPassword })
+        return this.update('users/password', {new_password: newPassword, old_password: oldPassword})
```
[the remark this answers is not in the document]
mjvkeulen: I think we should take this to our next meeting, adding this to the style discussion on function calls and make a final decision then.

src/vue/src/api/auth.js
```diff
@@ -2,13 +2,65 @@ import connection from '@/api/connection'
...
+        router.app.$toasted.error(connArgs.customErrorToast)
+    }
+}
+
+/* Lowers the connection count and toast a success message if a custom one is provided or responseSuccessToast is set. */
+function handleSuccess (resp, connArgs) {
+    setTimeout(function () { store.commit('connection/CLOSE_API_CALL') }, 300)
+
+    if (connArgs.responseSuccessToast) {
+        router.app.$toasted.success(sanitization.escapeHtml(resp.data.description))
```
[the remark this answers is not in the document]
mjvkeulen: Not currently used, but making this generic won't hurt!

src/vue/src/components/journal/JournalNonStudent.vue
mjvkeulen: I will fix this by adding additional validation to the packConnArgs function, allowing null where previously applicable.

src/vue/src/api/auth.js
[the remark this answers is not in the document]
mjvkeulen: Also, the regex does not compare object syntax only, as scope brackets are included.

src/vue/src/utils/sanitization.js
Larspolo: Why is `var` used here? shouldnt it be `const`? Previous time I used `var` I had to chance it to `let`

src/vue/src/utils/sanitization.js
```diff
@@ -0,0 +1,19 @@
...
+    '"': '&quot;',
+    "'": '&#39;',
+    '/': '&#x2F;',
+    '`': '&#x60;',
+    '=': '&#x3D;'
+}
+
+/* Taken from https://github.com/janl/mustache.js/blob/master/mustache.js#L60 */
+export default {
+    escapeHtml (string) {
```
engelhamer: Why not export this particular function as default? I don't think `sanitization.js` will serve any other purpose.

src/vue/src/api/auth.js
```diff
@@ -2,13 +2,81 @@ import connection from '@/api/connection'
...
+            data = genericUtils.parseArrayBuffer(error.response.data)
+        } else {
+            data = error.response.data
+        }
+
+        /* The Django throttle module uses detail as description. */
+        var message = data.description ? data.description : data.detail
+        if (message) { router.app.$toasted.error(sanitization.escapeHtml(message)) }
+    } else {
+        router.app.$toasted.error(connArgs.customErrorToast)
```
engelhamer: Shouldn't this be escaped as well?

src/vue/src/utils/sanitization.js
engelhamer: That would also allow for directly using `escapeHtml(..)` instead of `sanitization.escapeHtml(..)`.

src/vue/src/api/auth.js
mjvkeulen: customToasst are static strings

engelhamer: Ah I see.

src/vue/src/api/auth.js
```diff
@@ -2,13 +2,81 @@ import connection from '@/api/connection'
...
+        var message
+        if (resp.data instanceof ArrayBuffer) {
+            message = genericUtils.parseArrayBuffer(resp.data).description
+        } else {
+            message = resp.data.description
+        }
+
+        if (message) { router.app.$toasted.success(sanitization.escapeHtml(message)) }
+    } else if (connArgs.customSuccessToast) {
+        router.app.$toasted.success(sanitization.escapeHtml(connArgs.customSuccessToast))
```
engelhamer: Custom toasts are static strings so this needs not be escaped 😉.

src/vue/src/utils/sanitization.js
mjvkeulen: I feel it's not unlikely we may want to sanitize other input. Therefore I think this is not required.

PR #481, opened 2018-10-28
sctripped all content where we use v-html in the frontend

Larspolo:
## Description
sctripped all content where we use v-html in the frontend


libre-man:
Why are you trying to strip all tags instead of simply replacing `<` and alike?

Larspolo:
We want to allow the tags, this is editable by a rich text editor that allows for header, paragraphs, images(?), links and what not. We only want to disable the script tags, as that would allow for XSS through the rich text editor when hacked around it

libre-man:
It is really way better to keep a whitelist for this kind of stuff: this might introduce a formatting bug but NEVER a security bug (when the whitelist is kept correctly of course). As you have complete control over the editor this is quite easy to do.

PR #484, opened 2018-10-30
Security review lti

BigDGW:
## Description
Handels the LTI issue specified in the security review

## Summary of changes
- LTI logic removed from user.py and placed in lti.py
- Email is not given when forgotten password using only username
- lti.py constant are now enums
- LTI authentication is now handeld with a early exit
- Users are protected from stolen jwt_params attacks
- Opening of the role_config file is only done once in settings
- A initial submission is done when a assignment with entries in linked to canvas and the needed sourced id and grade_url has been supplied by the owner of those entries
- Some minor style changes


src/vue/src/components/account/RegisterUser.vue
```diff
@@ -54,13 +54,14 @@ export default {
...
             if (this.lti) {
                 this.form.username = this.lti.username
-                this.form.firstname = this.lti.firstname
-                this.form.lastname = this.lti.lastname
-                this.form.email = this.lti.email
+                // this.form.firstname = this.lti.firstname
+                // this.form.lastname = this.lti.lastname
+                // this.form.email = this.lti.email
                 this.form.ltiJWT = this.lti.ltiJWT
+                console.log('tester')
```
Rickyboy320: Debug code's still present

libre-man:
I noticed that Canvas added a new option in an LTI passback, which is `submittedAt`. We started using it at CodeGrade and it makes setting submission dates way easier. See [here](https://github.com/instructure/canvas-lms/commit/4b3fb3aff920184f6db243a0a7b79400a27ac433) for more details.

Larspolo:
"All lti security issues" doesnt really show what the problem are, maybe you should get a bit more descriptive so I can actually check if they are gone now?

src/django/VLE/views/email.py
```diff
@@ -39,12 +39,13 @@ def forgot_password(request):
     # We are retrieving the username based on either the username or email
     try:
         user = User.objects.get(username=username)
+        email = 'your recovery address'
```
Larspolo: ?

src/django/VLE/views/email.py
```diff
@@ -39,12 +39,13 @@ def forgot_password(request):
...
     except User.DoesNotExist:
         user = User.objects.get(email=email)
 
     email_handling.send_password_recovery_link(user)
     return response.success(
-        description='An email was sent to {}, please follow the email for instructions.'.format(user.email))
+        description='An email was sent to {}, please follow the email for instructions.'.format(email))
```
Larspolo: what is that change?

src/django/VLE/views/email.py
BigDGW: This is on request of Maarten as in the current build you get the see the email of the username you gave. What is a leak of someones personal data

src/django/VLE/views/email.py
BigDGW: email is a parameter that was given by the person that requested that he forgot his password and if it was not used then it set to a place holder. This way the email corrensponding to the account is never leaked.

src/django/VLE/views/assignment.py
```diff
@@ -263,9 +265,11 @@ def upcoming(self, request):
         deadline_list = []
 
         # TODO: change query to a query that selects all upcoming assignments connected to the user.
```
Larspolo: This can be removed

src/django/VLE/views/assignment.py
```diff
@@ -3,6 +3,8 @@
 
 In this file are all the assignment api requests.
 """
+import datetime
```
Larspolo: A django timezone would be preferred over an datetime `import django.utils.timezone`

src/django/VLE/lti_grade_passback.py
```diff
@@ -149,22 +158,31 @@ def parse_return_xml(self, xml):
                 'description': description}
 
 
-def needs_grading(journal, nID):
+def needs_grading(journal, node):
     """Give the teacher a needs grading notification in lti instancie."""
-    secret = settings.LTI_SECRET
-    key = settings.LTI_KEY
+    if journal.sourcedid is not None and journal.grade_url is not None:
```
Larspolo: Its preferred to use:
```
if journal.sourcedid is None or journal.grade_url is None:
    return
```
This is to reduce indentation on the rest of the code

src/django/VLE/lti_grade_passback.py
```diff
@@ -149,22 +158,31 @@ def parse_return_xml(self, xml):
...
+        secret = settings.LTI_SECRET
+        key = settings.LTI_KEY
 
-    jID = journal.pk
-    aID = journal.assignment.pk
-    cID = journal.assignment.courses.order_by('-startdate').first().pk
+        nID = node.pk
+        jID = journal.pk
+        aID = journal.assignment.pk
+        cID = journal.assignment.courses.order_by('-startdate').first().pk
```
Larspolo: Maybe this should also be given in the function, for now its only possible to grade from inside a course, that course can be used to grade. And if no courses are supplied, I would prefer updating the grades on all the courses iso just the latest (or maybe a setting in the course edit page?)

src/django/VLE/lti_grade_passback.py
```diff
@@ -23,7 +23,10 @@ def __init__(self, key, secret, journal, send_score=False, result_data=None):
         self.secret = secret
         self.url = None if journal is None else journal.grade_url
         self.sourcedid = None if journal is None else journal.sourcedid
-        if send_score and journal.assignment is not None and journal.assignment.points_possible is not None:
+        self.timestamp = submitted_at
+
+        if (send_score and journal is not None and journal.assignment is not None and
```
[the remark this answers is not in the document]
BigDGW: if i remove the () then python doesnt recognise it continues on the next line so i need to add \

what is better?

src/django/VLE/lti_grade_passback.py
```diff
@@ -149,22 +158,31 @@ def parse_return_xml(self, xml):
                 'description': description}
 
 
-def needs_grading(journal, nID):
+def needs_grading(journal, node):
```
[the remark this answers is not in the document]
BigDGW: what do you mean it is not changed? 
i made the change because i now needed the node it self while previously i only needed the id to create a url

src/django/VLE/lti_grade_passback.py
BigDGW: Even if we later want to send it to multiple courses that wont just be fixed with a function. As we then need to support multiple journal.grade_url and journal.sourcedid. What currently isn't supported as if a assignment is linked to a te same assignment but then year 2 in the VLE (think AVI collogium) then you dont want to send it to previous years edition

If we want to do it,  it is most certainly out of the scope of this branch and also of our current sprint.

src/django/VLE/lti_grade_passback.py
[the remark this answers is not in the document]
Larspolo: \ is better, but maybe it fits on one line if you also use my second suggestion?

src/django/VLE/lti_grade_passback.py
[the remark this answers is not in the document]
Larspolo: I ment the other keywords like `nID` and`jID` to `node_id` and `journal_id` respectively

src/django/VLE/views/lti.py
```diff
@@ -124,45 +144,34 @@ def lti_launch(request):
...
+    params['exp'] = datetime.datetime.utcnow() + datetime.timedelta(minutes=15)
+    lti_params = encode_lti_params(params)
 
+    try:
+        if user is None:
+            query = QueryDict(mutable=True)
+            query['state'] = LTI_STATES.NO_USER.value
+            query['lti_params'] = lti_params
+            query['username'] = params['custom_username']
+            return redirect(lti.create_lti_query_link(query))
```
Larspolo: In this case I would prefer using an `if else` structure because then you dont have to return the query on two seperate lines

src/django/VLE/views/user.py
```diff
@@ -95,13 +106,11 @@ def create(self, request):
...
-        if jwt_params:
-            lti_params = lti.decode_lti_params(jwt_params)
-            lti_id, user_image = utils.optional_params(lti_params, 'user_id', 'custom_user_image')
-            is_teacher = settings.ROLES['Teacher'] in lti_launch.roles_to_list(lti_params)
-        else:
-            lti_id, user_image, is_teacher = None, None, False
+
+        lti_id, user_image, full_name, email, is_teacher = get_lti_params(
+            request, 'user_id', 'custom_user_image', 'custom_user_full_name', 'custom_user_email')
+        if full_name:
```
Larspolo: Maybe this can be inside the `get_lti_params` as well?

src/django/VLE/settings/development.py
```diff
@@ -28,7 +28,7 @@
 
 with open(LTI_ROLE_CONFIG_PATH) as role_config:
     ROLES = json.load(role_config)
-    LTI_ROLES = dict((ROLES[k], k) for k in ROLES)
+    LTI_ROLES = {value: key for (key, value) in ROLES.items()}
```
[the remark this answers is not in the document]
BigDGW: In code we need both from lti role to our default roles and the other way around. So we have ROLES (ours to lti) and LTI_ROLES (lti to ours)

src/django/VLE/views/lti.py
BigDGW: this wil be removed in an other pull request as then full name will always be used and not first and last name

src/django/VLE/views/user.py
BigDGW: this wil be removed in an other pull request as then full name will always be used and not first and last name

src/django/VLE/lti_grade_passback.py
```diff
@@ -149,22 +157,33 @@ def parse_return_xml(self, xml):
...
+
     secret = settings.LTI_SECRET
     key = settings.LTI_KEY
 
-    jID = journal.pk
-    aID = journal.assignment.pk
-    cID = journal.assignment.courses.order_by('-startdate').first().pk
+    node_id = node.pk
+    journal_id = journal.pk
+    aissignmnet_id = journal.assignment.pk
```
mjvkeulen: Hm, this got me good for a sec.

src/django/VLE/views/lti.py
```diff
@@ -35,6 +41,11 @@ def encode_lti_params(jwt_params):
     return jwt.encode(jwt_params, settings.SECRET_KEY, algorithm='HS256').decode('utf-8')
 
 
+def split_fullname(fullname):
+    splitname = fullname.split(' ')
+    return splitname[0], fullname[len(splitname[0])+1:]
```
mjvkeulen: Why are we splitting and doing a lengthwise access?

Larspolo: will get fixed in Joey's branch coming later this week

PR #503, opened 2018-10-31
Fix expired deadlines in upcoming

### LINKED ISSUE #501: Upcoming deadlines show assignments that are past the deadline
**Describe the bug**
Upcoming deadlines includes assignments that are past the deadline 

**To Reproduce**
1. Make an assignment with a date before today
2. Go to the home page
3. See the assignment in the upcoming deadline

**Expected behavior**
Upcoming deadlines and no assignments that are already past the deadline

**Screenshots**
image

JowehL:
## Description
Expired deadlines no long show up in the deadline deck

## Summary of changes
- Fixes #501

### DIFF
--- a/src/django/VLE/views/assignment.py
+++ b/src/django/VLE/views/assignment.py
@@ -13,6 +13,8 @@ import VLE.utils.responses as response
 from VLE.models import Assignment, Course, Entry, Lti_ids
 from VLE.serializers import AssignmentSerializer
 from VLE.utils.error_handling import VLEMissingRequiredKey, VLEParamWrongType
+from django.db.models import Q
+from datetime import datetime
 
 
 class AssignmentView(viewsets.ViewSet):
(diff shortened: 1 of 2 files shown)


JowehL:
@Larspolo At the moment we can filter the expired deadlines in the backend in the serializer(which I don't recommend since we maybe want to use expired deadlines in other areas) or the assignment.py. And I don't think its clean to loop through all assignments again to check for expired deadlines.

Filtering the deadlines on the front-end would also be better for the server, maybe we could rename it to get_deadlines or create a separate serializer for it?

Larspolo:
I would suggest using the following instead, this works a bit cleaner imo
image

Larspolo:
http://localhost:8080/AssignmentsOverview should give all the assignments back, currently it only returns the upcoming.

Larspolo:
Closing this PR because I made a lot more changes to this branch to fix deadline/assignment lists over all pages

PR #509, opened 2018-11-12
Bugfix/deadline

### LINKED ISSUE #501: Upcoming deadlines show assignments that are past the deadline

Larspolo:
## Description
Overall rework of the assignment list queries.

## Summary of changes
- Fixes #501 
- Deadlines now show "unpublished" badge when unpublished
- Assignment routing is the same everywhere
- TodoCards now always show stats
- some more smaller changes


src/django/VLE/views/assignment.py
```diff
@@ -43,18 +46,18 @@ def list(self, request):
...
+            course = Course.objects.get(pk=course_id)
+            request.user.check_participation(course)
+            courses = [course]
+        except VLEParamWrongType:
+            course = None
+            courses = request.user.participations.all()
 
-        serializer = AssignmentSerializer(assignments, many=True, context={'user': request.user, 'course': course})
+        query = Assignment.objects.filter(courses__in=courses).distinct()
+        viewable = [assignment for assignment in query if request.user.can_view(assignment)]
```
JowehL: Can’t you combine the can_view assignment in the line above with a filter can_view_assignment

Larspolo: That would be nice to have, but as that requires looking at the participation of the role in all the courses of the assignment and then checking if that user has the correct rights (which are also multiple) I choose to do it like this. As I dont know how I could implement such a thing, and now when we change when a student can view an assignment, its automatically fixed right here as well.

src/django/VLE/serializers.py
```diff
@@ -150,28 +151,27 @@ class Meta:
...
-            return nodes[0].deadline
-
         # If the user has a journal, take the first upcoming not submitted deadline
-        else:
+        if 'user' in self.context and self.context['user'].has_permission('can_have_journal', assignment):
             try:
                 journal = Journal.objects.get(assignment=assignment, user=self.context['user'])
             except Journal.DoesNotExist:
-                return None
+                return assignment.due_date
```
Larspolo: @mjvkeulen @engelhamer @BigDGW  Do we want to return the due date or the lock date here? I still dont exactly know when to use which

BigDGW: The due date is the deadline for a student and lockdate is the latest possible time a student can submit something. As this is just for students i dont understand how it could be possible that he is part of the assignment but doesnt have a journal. As that should always be the case. if not then a error has occurred. And the journal should just be created.

src/django/VLE/serializers.py
```diff
@@ -150,28 +151,27 @@ class Meta:
...
 
             return deadlines[0]['preset__deadline']
 
+        # If the user doesnt have a journal, take the deadline that is the first upcoming deadline
+        else:
+            nodes = assignment.format.presetnode_set.all().order_by('deadline')
+            if not nodes:
+                return None
+            return nodes[0].deadline
```
BigDGW: Is returning the first deadline not a bit generic shouldnt you take into account deadlines where the TA or Teacher (those without a journal) have to do something like grading or give feedback?

src/django/VLE/views/assignment.py
BigDGW: From a performance stand point it would be better to work it into the query as this way you do way more calls to the database.

src/django/VLE/views/assignment.py
```diff
@@ -256,20 +259,19 @@ def upcoming(self, request):
...
-        # TODO: change query to a query that selects all upcoming assignments connected to the user.
-        for course in courses:
-            if request.user.is_participant(course):
-                for assignment in Assignment.objects.filter(courses=course.id, is_published=True).all():
-                    deadline_list.append(
-                        AssignmentSerializer(assignment, context={'user': request.user, 'course': course}).data)
+        query = Assignment.objects.filter(
+            Q(lock_date__gt=datetime.now()) | Q(lock_date=None), courses__in=courses
+        ).distinct()
+        viewable = [assignment for assignment in query if request.user.can_view(assignment)]
```
BigDGW: Same point as above this is putting unnecessary strain on the database

src/django/VLE/views/assignment.py
JowehL: What you can do is filtering the courses on the can_view_assignment first on line 56(51?), it seems more django like than using a list comprehension

src/django/VLE/serializers.py
Larspolo: That would be nice, maybe I will fix that if its easy to do, but its not really for this PR.

src/django/VLE/serializers.py
mjvkeulen: Dennis explained the difference accurately, @Larspolo does this answer your question?

src/django/VLE/serializers.py
```diff
@@ -150,28 +151,27 @@ class Meta:
...
 
             # TODO Incorporate assigment end date
             deadlines = journal.node_set.exclude(preset=None).values('preset__deadline').order_by('preset__deadline')
             if not deadlines:
-                return None
+                return assignment.due_date
 
             return deadlines[0]['preset__deadline']
 
+        # If the user doesnt have a journal, take the deadline that is the first upcoming deadline
```
mjvkeulen: In what scenario does a user not have a journal?

src/django/VLE/serializers.py
Larspolo: True, then due_date is indeed the preferred option to return by default. But in this case I think indeed that the try except should be removed, as everyone should have a journal be default. I will update it accordingly

engelhamer: Agree with all the above.

src/django/VLE/serializers.py
engelhamer: Would definitely be a useful addition!

src/django/VLE/views/assignment.py
Larspolo: > What you can do is filtering the courses on the can_view_assignment first on line 56(51?), it seems more django like than using a list comprehension

This does not work, as `can view assignment` is per assignment within the course different. 

> From a performance stand point it would be better to work it into the query as this way you do way more calls to the database.

True, but it takes a lot of time, and there are other things to fix as well. Currently I only loop through the assignments that are connected to a user, and that is a maximum of 3 right now. Every server should be able to handle this. I will add a todo so it will be checked in some later stage.

src/django/VLE/serializers.py
Larspolo: Old comment, will update it

src/django/VLE/serializers.py
Larspolo: "maybe I will fix that if its easy to do" , it was not easy, but fixed it anyway. Please check this PR again because I improved the deadlines, improved stats, and separated teacher and student deadlines

src/django/VLE/views/assignment.py
Larspolo: Its not that much compared to the work that needs to be put into creating that query

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -1,63 +1,87 @@
...
 </script>
 
 <style lang="sass" scoped>
 @import "~sass/modules/colors.sass"
 
-h6
-    display: inline
+.right
+    float: right
+.red
```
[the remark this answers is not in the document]
engelhamer: You could add another sass file for that. Also it can be more verbose (e.g. `text-red`).

src/django/VLE/utils/generic_utils.py
```diff
@@ -55,6 +52,23 @@ def required_typed_params(post, *keys):
...
+
+def optional_typed_params(post, *keys):
+    if keys and not post:
+        return [None] * len(keys)
+
+    result = []
+    for func, key in keys:
+        if key in post and post[key] != '':
+            try:
+                result.append(func(post[key]))
```
libre-man: Why not simply do an `isinstance` check?

Larspolo: How would that work? Its still a string, so that would always return false right?

libre-man: Why are you sending strings to the server if they need to be a different type? But yeah, in that case it wouldn't work.

src/vue/src/api/assignment.js
```diff
@@ -1,6 +1,7 @@
 import auth from '@/api/auth'
 
 export default {
+
```
mjvkeulen: Extra newline?

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -1,67 +1,79 @@
...
+        <h5 class="deadline-name">{{ deadline.name }}</h5>
+        <b-badge
+            v-if="!deadline.is_published"
+            class="ml-2 mt-2">
+            Unpublished
+        </b-badge>
+        <br />
+        {{ course.abbreviation }}
+
+        <h6 v-if="this.deadline.deadline && (this.deadline.stats.needs_marking + deadline.stats.unpublished) ">
```
mjvkeulen: Extra space before closing  `"`

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -1,67 +1,79 @@
...
+            <template>{{ timeLeft[1] }} ago</template>
```
mjvkeulen: Is there an advantage of using template tags here?

src/vue/src/sass/partials/colors.sass
```diff
@@ -0,0 +1,56 @@
+@import '~sass/modules/colors.sass'
+
+.text-dark-blue
```
mjvkeulen: Hm, these should be reduced to a mixin accepting a colour type. Using a variable loader this would then be usable in our templates. Remind me!

src/django/VLE/serializers.py
```diff
@@ -257,13 +307,13 @@ def get_student(self, journal):
...
     def get_stats(self, journal):
-        entries = utils.get_journal_entries(journal)
         return {
-            'acquired_points': utils.get_acquired_points(entries),
-            'graded': utils.get_graded_count(entries),
-            'published': utils.get_published_count(entries),
-            'submitted': utils.get_submitted_count(entries),
-            'total_points': utils.get_points_possible(journal),
+            'acquired_points': journal.node_set.filter(entry__published=True)
+            .values('entry__grade').aggregate(Sum('entry__grade'))['entry__grade__sum'],
```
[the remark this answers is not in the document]
Larspolo: This is how flake likes it, as its in a dict

PR #510, opened 2018-11-13
Feature/groups lti

### LINKED ISSUE #485: Groups from VLE
The groupname that is gotten from datanose should be used to put the user in the same group

**User story**
I want to be in the same group in ejournal as in the VLE

BigDGW:
## Description
The custom_group_name from datanose is now used to create and place users in the group with that name.

## Summary of changes
- Fixes #485


src/django/VLE/lti_launch.py
```diff
@@ -106,10 +107,16 @@ def check_course_lti(request, user, role):
 
     if lti_couples.count() > 0:
         course = lti_couples[0].course
-        if user not in course.users.all():
+        group = group_utils.get_and_init_group(request.get('custom_group_name', ''), course)
+        # always update the group through lti params
+        if user in course.users.all():
```
mjvkeulen: This should be done via query using exists()

BigDGW: Why? I think the best would be to preload course.users, because that way it will only be 1 call to the database and for performance it is best to reduce the number of database calls (queries) as much as possible.

BigDGW: But all querys need to be run sequentieel because of race conditions. While python code is run multithreaded

libre-man: Your database is properly optimised for stuff like this, it uses threads... For an `exist` a read lock is probably acquired, which makes sure the table isn't mutated while checking the condition, although it wouldn't surprise me that databases like postgres employ even finer grained locking. Furthermore, getting all users from your database will cost you in serialization and deserialization, your ORM is also written in python and isn't very quick.

src/django/VLE/lti_launch.py
```diff
@@ -106,10 +107,16 @@ def check_course_lti(request, user, role):
...
+            participation = Participation.objects.get(user=user, course=course)
+            participation.group = group
```
engelhamer: I would suggest only updating the group if is specified (i.e. not None). Otherwise adding a group in the application _alone_ would not be possible. (Correct me if I'm wrong, since I cannot test with this parameter right now)

mjvkeulen: This is currently the case as group will be None if the lti_param is not set (and thus has an empty val)

PR #513, opened 2018-11-25
Feature/sanitize html front aswell

mjvkeulen:
## Description
Sandboxed all v-html to iframes to block all script execution.


src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -0,0 +1,55 @@
...
+        marginwidth="0"
+        marginheight="0"
+        scrolling="no"
+        class="w-100 theme-iframe"
+    />
+</template>
+
+<script>
+export default {
+    name: 'Test',
```
engelhamer: This should be changed.

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -145,7 +146,8 @@ export default {
         'url-input': urlInput,
         'pdf-display': pdfDisplay,
         'file-download-button': fileDownloadButton,
-        'image-file-display': imageFileDisplay
+        'image-file-display': imageFileDisplay,
+        sandboxedIframe
```
engelhamer: Still surprises me that this works straight away.

src/vue/src/main.js
```diff
@@ -8,58 +8,7 @@ import '../node_modules/bootstrap/dist/css/bootstrap.css'
...
-import 'vue-awesome/icons/long-arrow-down'
-import 'vue-awesome/icons/comments'
-import 'vue-awesome/icons/cog'
-import 'vue-awesome/icons/clock-o'
-import 'vue-awesome/icons/print'
-import 'vue-awesome/icons/github'
-import 'vue-awesome/icons/linkedin'
-import 'vue-awesome/icons/file'
-import 'vue-awesome/icons/search'
+import '@/helpers/vue-awesome-icons.js'
```
engelhamer: Smart.

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -17,7 +17,7 @@
                             <icon name="edit"/>
                             Edit
                         </b-button>
-                        <div v-html="comment.text"/>
+                        <sandboxed-iframe :content="comment.text"/>
```
engelhamer: This introduces some peculiar padding: The iframce clears for the buttons floating on the right.

screenshot from 2018-11-26 16-43-28

engelhamer: Nice!

engelhamer:
Changed the colors to match the hourglass and timestamp.

screenshot from 2018-11-28 17-36-16

src/vue/src/components/journal/CommentCard.vue
engelhamer: Changed the colors to match the hourglass and timestamp.

screenshot from 2018-11-28 17-36-16

src/vue/src/Header.vue
```diff
@@ -19,8 +19,14 @@
 
         <b-collapse is-nav id="nav-collapse">
             <b-navbar-nav class="mr-auto">
-                <b-nav-item :to="{ name : 'Home' }">Courses</b-nav-item>
-                <b-nav-item :to="{ name : 'AssignmentsOverview' }">Assignments</b-nav-item>
+                <b-nav-item :to="{ name : 'Home' }">
+                    <icon name="home"/>
+                    Courses
```
Larspolo: Courses stay lit up when going inside a course, and it also switched back top courses being lit when going from assignments overview to a different page. Can this be fixed in this PR or maybe just create another issue?

src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -0,0 +1,55 @@
...
+        },
+        setCustomStyle (obj) {
+            var doc = obj.contentWindow.document
+
+            var css = `
+body {
+    font-family: 'Roboto', sans-serif;
+    text-decoration: none;
+    color: #252C39;
+}`
```
Larspolo: Maybe add link style as well?
Now:
image

Before:
image

src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -0,0 +1,55 @@
...
+
+            var style = document.createElement('style')
+            style.type = 'text/css'
+            style.appendChild(document.createTextNode(css))
```
Larspolo: Cant this also be done by just using an url to a stylesheet? I would prefer that over typing all the css in here

src/vue/src/Header.vue
engelhamer: That is desired behaviour, and has always been the case. Maybe we could highlight 'Assignments' once the user is in an assignment, that would be nice!

engelhamer: Doesn't have anything tot do with sanitization (actually my previous addition did nog either, sorry for that!), so another issue would be appropriate.

src/vue/src/components/assets/SandboxedIframe.vue
engelhamer: That would be nice, tackles a lot of issues at once (the current implementation obviously doesn't scale).

mjvkeulen: Could use a stylesheet, if all we want to do is change link styling I would rather just append.

mjvkeulen: Using theme blue

image

PR #515, opened 2018-11-27
Fixed LTI header login

### LINKED ISSUE #339: Internal server error when entering through LTI but logging in normally after
# Bug report
Internal server error occurs after entering through LTI as an unlinked teacher and using the header login immediately after, without creating/linking an account.

## Expected Behavior
LTI session to be ended and user to be logged in normally.

## Actual Behavior
Internal server error is returned with LTI session related message.

1. Open AVI/Portfolio through LTI as a teacher.
2. Use the header bar to go to login view.
3. Attempt to log in normally.

engelhamer:
## Description
Fixed an issue that caused a 500 error when logging in via the header whilst arriving via LTI.

## Summary of changes
- Fixes #339


src/vue/src/views/Login.vue
```diff
@@ -19,11 +19,20 @@ export default {
...
         handleLoginSucces () {
-            if (this.$root.previousPage === null || this.$root.previousPage.name === null ||
-                this.$root.previousPage.name === 'PasswordRecovery' || this.$root.previousPage.name === 'ErrorPage' ||
-                this.$root.previousPage.name === 'Login') {
+            if (this.$root.previousPage === null) {
                 this.$router.push({name: 'Home'})
-            } else {
+            }
+
+            switch (this.$root.previousPage.name) {
```
mjvkeulen: Good solution, however elsewhere we use set membership for this allowing for a more verbose name. What is your preference? 

From router:
`unavailableWhenLoggedIn.has(to.name)`

engelhamer: Where would you suggest we define that set? I do not think that defining it inside this function will be very helpful.

engelhamer: It does appear as if we could use the `permissionlessContent` set from the router though.

mjvkeulen: That set could be moved to utils/constants then?

engelhamer: Alright, I'll do that.

src/vue/src/utils/constants/router_constraints.js
```diff
@@ -0,0 +1,21 @@
...
+])
+
+const UNAVAILABLE_WHEN_LOGGED_IN = new Set([
+    'Login',
+    'Guest',
+    'Register'
+])
+
+export default {
+    PERMISSIONLESS_CONTENT: PERMISSIONLESS_CONTENT,
```
mjvkeulen: I feel we should use the shorthand were possible, e.g:

```
export default {
    PERMISSIONLESS_CONTENT,
    UNAVAILABLE_WHEN_LOGGED_IN
}
```

mjvkeulen: We should probably enforce this style rule, but there are many more pressing items.

src/vue/src/views/Login.vue
```diff
@@ -20,12 +21,11 @@ export default {
...
             if (this.$root.previousPage === null || this.$root.previousPage.name === null ||
-                this.$root.previousPage.name === 'PasswordRecovery' || this.$root.previousPage.name === 'ErrorPage' ||
-                this.$root.previousPage.name === 'Login') {
+                routerConstraints.PERMISSIONLESS_CONTENT.has(this.$root.previousPage.name)) {
                 this.$router.push({name: 'Home'})
-            } else {
-                this.$router.push({name: this.$root.previousPage.name, params: this.$root.previousPage.params})
             }
+
+            this.$router.push({name: this.$root.previousPage.name, params: this.$root.previousPage.params})
```
Larspolo: Why isnt this in an else statement? seems a bit odd (without a return)

Larspolo: Also, I cant get this system to work, when I
3. logout
4. go to http://localhost:8080/Home/Course/1
5. (when redirected to login page) login
I dont go to  http://localhost:8080/Home/Course/1, I start on the homepage again.

Can this b4e fixed in this PR as well?

src/vue/src/utils/constants/router_constraints.js
engelhamer: Followed the same format as in the other constant files here.

src/vue/src/views/Login.vue
engelhamer: Are steps 1 and 2 missing or is the numbering just odd? The behaviour you describe does not seem desirable, but I don't see how this could originate from the changes I made.

I omitted the else statement since it's simply not required here (and you know that I enjoy removing lines of course 😉).

Larspolo: Numbering is off, and I will create an issue with the correct numbering so that it can be fixed in the future

PR #523, opened 2018-12-06
Postgres for development

mjvkeulen:
## Description

- Added Postgres to the development environment, ensuring the same conditions are met during development as production. Opening the actual Postgres benefits, inter alia, array fields and JSON fields. 
- Cleaned the makefile somewhat, Travisci should be faster.
- `make shell`, quickly get into the Django shell, just kill the shell to go back to the root dir.
- `make run-test arg=<test>` e.g. `make run-test arg=jwt` to run all jwt tests.


src/django/test/test_delete_apis.py
```diff
@@ -76,12 +76,13 @@ def test_delete_assignment(self):
         assign1.courses.add(course2)
         assign2.courses.add(course1)
 
-        test.api_del_call(self, '/assignments/' + str(assign1.pk) + '/?course_id=1', login)
-        assignment = Assignment.objects.get(pk=1)
+        # Stupid test to begin with, glad this is replaced soon
+        test.api_del_call(self, '/assignments/' + str(assign1.pk) + f'/?course_id={self.course.pk}', login)
```
Larspolo: I prefer to use python 3.5 string formatting (`'{}'.format(5)`), because then we support python3.5 as well, and our uva server currently has 3.5

src/django/test/test_delete_apis.py
```diff
@@ -76,12 +76,13 @@ def test_delete_assignment(self):
...
+        assignment = Assignment.objects.get(pk=assign1.pk)
         self.assertEquals(assignment.courses.count(), 2)
 
-        test.api_del_call(self, '/assignments/1/?course_id=1', login)
-        self.assertEquals(Assignment.objects.filter(pk=1).count(), 1)
+        test.api_del_call(self, f'/assignments/{assign1.pk}/?course_id={self.course.pk}', login)
```
Larspolo: same here. Maybe sometime we need to change it to 3.6, as its cleaner imo, but then it needs to be changed everywhere, and also the server needs to be updated

src/django/test/test_update_apis.py
```diff
@@ -45,15 +45,15 @@ def test_update_course_roles(self):
...
+        result = test.api_get_call(self, '/roles/', login, params={'course_id': self.course.pk})
 
         roles = result.json()['roles']
         for role in roles:
             if role['name'] == 'TA2':
                 role['can_grade'] = 1
 
         roles.append(serialize.RoleSerializer(factory.make_role_default_no_perms('test_role', self.course)).data)
-        test.api_patch_call(self, '/roles/1/', {'roles': roles}, login)
+        test.api_patch_call(self, f'/roles/{self.course.pk}/', {'roles': roles}, login)
```
Larspolo: .

src/django/test/test_delete_apis.py
mjvkeulen: Sharp f-strings are 3.6

PR #531, opened 2018-12-09
Fixed an issue that blocked all hyperlinks in sandboxed iframes

engelhamer:
## Description
Fixed an issue that blocked all hyperlinks in sandboxed iframes.

### DIFF
--- a/src/vue/src/components/assets/SandboxedIframe.vue
+++ b/src/vue/src/components/assets/SandboxedIframe.vue
@@ -1,7 +1,7 @@
 <template>
     <iframe
         @load="init($event)"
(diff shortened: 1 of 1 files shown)


src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <iframe
         @load="init($event)"
-        sandbox="allow-same-origin"
+        sandbox="allow-same-origin allow-popups"
```
engelhamer: Allows for launching a new page or window from within a sandboxed iframe.

PR #535, opened 2018-12-11
Added group migration file

Larspolo:
## Description
Forgot the migration file


mjvkeulen:
Did we decide on using custom names for the migrations files?

PR #536, opened 2018-12-11
Updated LTI publishment state handling

engelhamer:
## Description
The publishment state of an assignment on eJournal is now automatically updated based on the LTI parameter.


src/django/VLE/lti_launch.py
```diff
@@ -139,7 +139,7 @@ def check_assignment_lti(request):
     if lti_couples.count() > 0:
         assignment = lti_couples[0].assignment
         if 'custom_assignment_publish' in request:
-            assignment.is_published = request['custom_assignment_publish'] == 'true'
```
engelhamer: This conversion of the LTI parameter has been done before calling this function. Therefore the comparison used to always fail.

PR #542, opened 2019-01-18
Persistant filters

### LINKED ISSUE #521: Improve user-friendliness of journal filtering
Currently, there are a number of quirks in the journal filtering functionality that slow down the grading process:

- The sorting order (_ascending_ or _descending_) is reset every time a user visits the assignment page. For example when navigating from a journal via the back button in the browser. This order should persist like the sorting criterion, which can be accomplished by adding it to the query.
- The group filter is reset to one's group on every visit of the assignment page. It should maintain state (lke the above).
- The default sorting order (_ascending_) is not intuitive when a user specifies the sorting criterion '_marking needed_'. All journals that have no entries left to mark (i.e. require _no_ action) are presented at the top. It would be better to set the order to ascending by default for this option.
- All sort and search data is discarded when a user navigates back to the assignment page via the breadcrumb. This is very annoying because one has to manually configure these every time.
- Journals on the assignment page are automatically filtered by the user's own group. However, when there are no journals for that group filtering makes no sense. Instead it would be more helpful to show _all_ journals (i.e. not apply a filter at all) in this case.
- Next and previous buttons ignore group filters.

We should improve the persistence of the filtering in order to speed up the grading process.

**User story**
As a teacher I want to set a filter once and not have to update it until I leave the assignment or desire to change the filter myself.

mjvkeulen:
## Description
Moved all filtering preferences (journal, course edit, assignment overview, deadline/TODO decks) to the store and should persist in local storage. Journal filtering options remain when grading and navigating by breadcrumb and or router history.

## Summary of changes
- Fixes #521


src/vue/src/store/constants/mutation-types.js
```diff
@@ -9,5 +9,20 @@ export const SET_PROFILE_PICTURE = 'SET_PROFILE_PICTURE'
...
+export const SET_JOURNAL_SEARCH_VALUE = 'SET_JOURNAL_SEARCH_VALUE'
+export const SET_JOURNAL_SORT_BY = 'SET_JOURNAL_SORT_BY'
+export const SET_COURSE_MEMBERS_SORT_ASCENDING = 'SET_COURSE_MEMBERS_SORT_ASCENDING'
+export const SET_COURSE_MEMBERS_VIEW_ENROLLED = 'SET_COURSE_MEMBERS_VIEW_ENROLLED'
+export const SET_COURSE_MEMBERS_GROUP_FILTER = 'SET_COURSE_MEMBERS_GROUP_FILTER'
+export const SET_COURSE_MEMBERS_SEARCH_VALUE = 'SET_COURSE_MEMBERS_SEARCH_VALUE'
+export const SET_COURSE_MEMBERS_SORT_BY = 'SET_COURSE_MEMBERS_SORT_BY'
+export const SET_ASSIGNMENT_OVERVIEW_SORT_ASCENDING = 'SET_ASSIGNMENT_OVERVIEW_SORT_ASCENDING'
+export const SET_ASSIGNMENT_OVERVIEW_SEARCH_VALUE = 'SET_ASSIGNMENT_OVERVIEW_SEARCH_VALUE'
+export const SET_ASSIGNMENT_OVERVIEW_SORT_BY = 'SET_ASSIGNMENT_OVERVIEW_SORT_BY'
```
Larspolo: Why not set them to integers (as that would save some memory)?

src/vue/src/views/Assignment.vue
```diff
@@ -128,28 +129,34 @@ export default {
...
-                    this.selectedFilterGroupOption = participant.group.name
-                }
-            })
-
-        if (this.$route.query.sort === 'sortFullName' ||
-            this.$route.query.sort === 'sortUsername' ||
-            this.$route.query.sort === 'sortMarking') {
-            this.selectedSortOption = this.$route.query.sort
+                .then(groups => {
+                    if (!groups.some(group => { return (group.name === this.getJournalGroupFilter) })) {
```
Larspolo: `group => group.name === this.getJournalGroupFilter`

src/vue/src/views/Assignment.vue
```diff
@@ -128,28 +129,34 @@ export default {
...
 
-        if (this.$route.query.search) {
-            this.searchVariable = this.$route.query.search
+        if (!this.getJournalGroupFilter) {
+            participationAPI.get(this.cID)
+                .then(participant => {
+                    /* Group can be null */
+                    if (participant.group && participant.group.name) {
+                        this.setJournalGroupFilter(participant.group.name)
+                        if (!this.filteredJournals.length) {
```
Larspolo: Maybe this?
```
                    if (participant.group && participant.group.name && this.filteredJournals.length > 0) {
                        this.setJournalGroupFilter(participant.group.name)
                    }
```

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -309,6 +286,12 @@ export default {
         'journal-end-card': journalEndCard
     },
     computed: {
+        ...mapGetters({
+            getJournalSortBy: 'preferences/journalSortBy',
```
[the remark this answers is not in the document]
mjvkeulen: I disagree, as these are truly getters and are not variables that can be assigned to

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -327,26 +310,32 @@ export default {
             function checkFilter (user) {
                 var username = user.student.username.toLowerCase()
                 var fullName = user.student.name
-                var searchVariable = self.searchVariable.toLowerCase()
+                var searchVariable = self.getJournalSearchValue.toLowerCase()
```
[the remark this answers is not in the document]
mjvkeulen: Whereas we now compare lowercase only this entirely implementation dependent. I would find it strange to abstract to the getter of a value?

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -327,26 +310,32 @@ export default {
...
 
                 return username.includes(searchVariable) ||
                        fullName.includes(searchVariable)
             }
 
+            function groupFilter (assignment) {
+                if (self.getJournalGroupFilter) {
+                    return assignment.student.group === self.getJournalGroupFilter
```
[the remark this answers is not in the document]
mjvkeulen: Groups can't be filtered by manual input, only by selection options. As these variables are loaded in with capitals and all this would be redundant but I understand it could sate one's OCD, I think I nearly put it on there too haha.

src/vue/src/components/journal/JournalNonStudent.vue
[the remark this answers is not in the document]
Larspolo: For the other map getter we didnt write the get as well, shouldnt it at least be consistent?

mjvkeulen:
> All these computed getters and setters for each variable, are you sure that is necessary?

Well, necessary to work with v-model. This just reveals the mechanic under the hood; as these getters and setters are present anyway.

src/vue/src/store/constants/mutation-types.js
Larspolo: haha fair, too bad there is not good support for real enums in JS

src/vue/src/components/journal/JournalNonStudent.vue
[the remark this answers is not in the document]
mjvkeulen: It's true that I sometimes just named the getter similarly to what the previous variable was called. Wasn't farming lines I guess, and am not sure what we should do with this tbh.

engelhamer:
When applying the filter for the user's own group it is also important to note that when there are no journals for that group filtering makes no sense. Instead it would be more helpful to show all journals (i.e. not apply a filter at all) in this case. (Not sure if this is done correctly at the moment since filtering on the own group does not take place).

mjvkeulen:
> Good changes, however things have now gotten a little bit _too_ persistent: If I filter on a given string and group in one course, the application will try to do so for every other assignment I visit as well. This messes with group filters: a group filter in one course does not exist in another and therefore the filter is reset to _no filter at all_ rather than _my own group_.

Would defaulting to ones own group be a workable solution here? (Provided that group again has any journals).

mjvkeulen:
> Also the filtering and searching in the course editor does not seem very logical to me: why would you need persistence if there is absolutely no need to change pages when working on something there? Seeing back a search query I used a few days ago would only confuse me there.

I see where you're coming from, my thought process was that this page will most likely be used for user-specific pages (be it journal overview or other admin functionality), making it only natural to use persistent filters here as well. I could remove the functionality for now, but I am not sure if that's desired as I think there's value in being able to navigate through the site without losing filters.

mjvkeulen:
> > Good changes, however things have now gotten a little bit _too_ persistent: If I filter on a given string and group in one course, the application will try to do so for every other assignment I visit as well. This messes with group filters: a group filter in one course does not exist in another and therefore the filter is reset to _no filter at all_ rather than _my own group_.
> 
> Would defaulting to ones own group be a workable solution here? (Provided that group again has any journals). Hm I guess your solution of keeping one set of filters per assignment is cleaner, which is then reset when visiting a different assignment. I'll make it so.

This is now implemented

src/django/VLE/management/commands/preset_db.py
```diff
@@ -22,112 +22,154 @@ class Command(BaseCommand):
...
+        for c in courses_examples.values():
             course = Course(pk=c["pk"], name=c["name"], abbreviation=c["abbr"], startdate=c["start_date"],
-                            enddate=c["end_date"], author=author)
+                            enddate=c["end_date"], author=c["author"])
             course.save()
-            factory.make_role_student('Student', course)
-            factory.make_role_ta('TA', course)
-            factory.make_role_teacher('Teacher', course)
 
-            role_teacher = Role.objects.get(name='Teacher', course=course)
```
engelhamer: Since you removed these lines the imported `Role` class is unused.

src/vue/src/views/Assignment.vue
```diff
@@ -120,36 +121,51 @@ export default {
...
+
+            if (groups) { this.groups = groups }
+
+            /* If there are no groups or the current group filter yields no journals, remove the filter. */
+            if (!groups || !groups.some(group => group.name === this.getJournalGroupFilter)) {
+                this.setJournalGroupFilter(null)
+            }
+
+            /* If the group filter has not been set, set it to the group of the user provided that yields journals. */
+            if (this.getJournalGroupFilter === null && participant && participant.group && participant.group.name) {
```
engelhamer: This unfortunately still appears to fail: even when all users in a course are in the same group no filter is applied.

src/django/VLE/management/commands/preset_db.py
engelhamer: This yields a linter error and breaks the CI build.

engelhamer:
My bad, I guess something went wrong while testing.

PR #545, opened 2019-01-19
Fix for RTE inline mode combo with hidden footer

### LINKED ISSUE #474: Richtext editor crashes when using the inlinemode on true and footer on false
**Describe the bug**
When using the richtext editor, where the values of :displayinline is set to true and the :footer is set to false, the texteditor will crash and stay in the inlinemode even when you select the richtext editor.
It will throw the following error: TypeError: editor.theme.panel.find(...)[0] is undefined

**To Reproduce**
1. Make a richtext-editor with the settings :displayInlinemode="true" and :footer="false" 
2. Try to display the richtexteditor on a page
3. it will cause an error

**Expected behavior**
When the settings of  :displayInlinemode="true" and :footer="false"  are set the richtext-editor should display the richtext-editor in inlinemode when not selected and not in displayinlinemode when selected (the menubar with the tools should appear) and when selected the footer should not be shown, since this is set to false.

**Screenshots**
Now it just crashed:
image

It should display in in displayinlinemode when not selected:
image

And it should show the menu etc when selected, but without the footer (this one has the footer in it):
image

mjvkeulen:
## Description
Check added that prevents hiding the footer if its never initialized in the first place.

## Summary of changes
- Fixes #474

### DIFF
--- a/src/vue/src/components/assets/TextEditor.vue
+++ b/src/vue/src/components/assets/TextEditor.vue
@@ -174,18 +174,18 @@ export default {
 
             editor.theme.panel.find('toolbar')[0].$el.hide()
             if (!this.basic) { editor.theme.panel.find('menubar')[0].$el.hide() }
-            editor.theme.panel.find('#statusbar')[0].$el.hide()
+            if (this.footer) { editor.theme.panel.find('#statusbar')[0].$el.hide() }
 
             editor.on('focus', function () {
                 if (!vm.basic) { editor.theme.panel.find('menubar')[0].$el.show() }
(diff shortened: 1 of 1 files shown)


engelhamer:
So are we going to update the instances of the editor right away? Do you remember where we wanted to use this?

mjvkeulen:
> Why aren't we going to update the instances of the editor right away? Do you remember where we wanted to use this?

We liked the minimalistic display of the RTE for comments and certain fields. But with the toggle we are currently using on the RTE for certain fields I don't know if the footer should also be hidden?

engelhamer:
I guess you could hide the footer for comments then?

PR #546, opened 2019-01-19
Usermodel creation field limit checks.

### LINKED ISSUE #472: Usernames over 30 characters cause the database to crash
**Describe the bug**
If a username with more than 30 characters is entered, the database crashes.

**To Reproduce**

0. `sudo make install && sudo make deploy && sudo make serve`
1. Go to `localhost/Register`
2. Fill in for username `a reaaaaaaaaaaaaaaaaaaally long uuuuuuuuuusernaaaaaaaame`

**Expected behaviour**
The database not to crash

**Desktop:**
 - OS: Ubuntu 18.04
 - Browser: Firefox
 - Version: `develop` branch

**Stacktrace:**

```
ERROR:django.request:Internal Server Error: /users/, referer: http://localhost/Register
 Traceback (most recent call last):, referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute, referer: http://localhost/Register
     return self.cursor.execute(sql, params), referer: http://localhost/Register
 psycopg2.DataError: value too long for type character varying(30), referer: http://localhost/Register
 , referer: http://localhost/Register
 , referer: http://localhost/Register
 The above exception was the direct cause of the following exception:, referer: http://localhost/Register
 , referer: http://localhost/Register
 Traceback (most recent call last):, referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner, referer: http://localhost/Register
     response = get_response(request), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response, referer: http://localhost/Register
     response = self.process_exception_by_middleware(e, request), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 124, in _get_response, referer: http://localhost/Register
     response = wrapped_callback(request, *callback_args, **callback_kwargs), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view, referer: http://localhost/Register
     return view_func(*args, **kwargs), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/rest_framework/viewsets.py", line 103, in view, referer: http://localhost/Register
     return self.dispatch(request, *args, **kwargs), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/rest_framework/views.py", line 483, in dispatch, referer: http://localhost/Register
     response = self.handle_exception(exc), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/rest_framework/views.py", line 443, in handle_exception, referer: http://localhost/Register
     self.raise_uncaught_exception(exc), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/rest_framework/views.py", line 480, in dispatch, referer: http://localhost/Register
     response = handler(request, *args, **kwargs), referer: http://localhost/Register
   File "/var/www/ejournal//django/VLE/views/user.py", line 134, in create, referer: http://localhost/Register
     verified_email=True if lti_id else False), referer: http://localhost/Register
   File "/var/www/ejournal//django/VLE/factory.py", line 44, in make_user, referer: http://localhost/Register
     user.save(), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/contrib/auth/base_user.py", line 73, in save, referer: http://localhost/Register
     super().save(*args, **kwargs), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/base.py", line 718, in save, referer: http://localhost/Register
     force_update=force_update, update_fields=update_fields), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/base.py", line 748, in save_base, referer: http://localhost/Register
     updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/base.py", line 831, in _save_table, referer: http://localhost/Register
     result = self._do_insert(cls._base_manager, using, fields, update_pk, raw), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/base.py", line 869, in _do_insert, referer: http://localhost/Register
     using=using, raw=raw), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method, referer: http://localhost/Register
     return getattr(self.get_queryset(), name)(*args, **kwargs), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/query.py", line 1136, in _insert, referer: http://localhost/Register
     return query.get_compiler(using=using).execute_sql(return_id), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/models/sql/compiler.py", line 1289, in execute_sql, referer: http://localhost/Register
     cursor.execute(sql, params), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 68, in execute, referer: http://localhost/Register
     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 77, in _execute_with_wrappers, referer: http://localhost/Register
     return executor(sql, params, many, context), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute, referer: http://localhost/Register
     return self.cursor.execute(sql, params), referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/utils.py", line 89, in __exit__, referer: http://localhost/Register
     raise dj_exc_value.with_traceback(traceback) from exc_value, referer: http://localhost/Register
   File "/var/www/ejournal/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute, referer: http://localhost/Register
     return self.cursor.execute(sql, params), referer: http://localhost/Register
 django.db.utils.DataError: value too long for type character varying(30), referer: http://localhost/Register
 , referer: http://localhost/Register
```

mjvkeulen:
## Description
Addition of a response that parses a validation error (which can contain multiple fields and multiple errors per field), to a bad request which should be readable except it can get lengthy when there are multiple fields and errors. 
Front end checks added to the register component complacent with Django's AbstractUser class. Backend now first cleans the model to catch any validation errors.

## Summary of changes
- Fixes #472


src/django/VLE/views/user.py
```diff
@@ -139,17 +138,20 @@ def create(self, request):
...
-                                 verified_email=True if lti_id else False)
+        user = User(username=username, email=email, first_name=first_name, last_name=last_name, lti_id=lti_id,
+                    is_teacher=is_teacher, verified_email=True if lti_id else False,
+                    profile_picture=user_image if user_image else '/static/unknown-profile.png')
+        user.set_password(password)
+
+        try:
+            user.full_clean()
+        except ValidationError as e:
+            return response.validation_error(e)
```
Larspolo: Why not put all this in the `make_user` function?

mjvkeulen: The factory one? Afaik this is the only hook a user is actually created, factory.make_user is basically testing support.

Larspolo: I would prefer the creation of objects being extracted to either the models (which I want to fully switch to one day) or in the factory. As that would reduce the complexity of this function, and make it reusable

mjvkeulen: You don't want a conditional response abstracted to the factory.make_user right?

Larspolo: But maybe it can throw an error, and then only catch the error in this function?

Larspolo: Also maybe the validation error can be put into the error_handle?

src/django/VLE/utils/responses.py
```diff
@@ -146,31 +142,22 @@ def value_error(message=None):
...
-    else:
-        return not_found(description='File not found.')
+def validation_error(err):
+    """Formats a validation error into a readable bad_request response."""
+    resp = ""
+    field_count = 1
 
+    for field, validation_errors in err.__dict__["error_dict"].items():
+        # Wrong with field:
+        resp += "{}) {}: ".format(field_count, field)
```
engelhamer: Why only this closing parenthesis? Also string building like this isn't very pretty.

Larspolo: This creates `1) first_name: first_name is too long. 2) ...`

engelhamer: Ah I see, simply list formatting..

src/django/VLE/views/user.py
mjvkeulen: I do prefer the solution of top-level catching the validation error (even if the formatting is different from what we currently have). So this would require the response to be able to handle the formatting of both single and multiple field length validation errors. I will look into this!

src/django/VLE/utils/responses.py
```diff
@@ -146,31 +142,22 @@ def value_error(message=None):
...
+        field_count += 1
+        for validation_error in validation_errors:
+            for msg in validation_error:
+                resp += msg + ' '
```
Larspolo: inconsistent use of quotes. I mostly try to use "" for strings that will be seen in the frontend, and '' for backend only. Although this is not forced, 2 different uses in 1 function is a bit extreme

Larspolo: In other languages, true, but there it would also be used as a char value, not a string concatenation. Also for python there is no difference in use of quotes. For this reason I still find this inconsistent.

PR #547, opened 2019-01-19
Tutorial tours

### LINKED ISSUE #278: Tutorial for first-time users
# Feature request

## Description
Some advanced functionality (e.g. the format editor) of the application is quite complicated, and therefore rather confusing for first-time users. Adding a step-by-step tutorial for these pages when a user first visits them would be helpful.

## User story
As a user I want a step-by-step tutorial on how to use more advanced, complicated features.

## User interface
[Vue-tour](https://github.com/pulsardev/vue-tour) should provide for some good functionality for this.

### LINKED ISSUE #345: Redesign of the "stackable templates" feature

engelhamer:
## Description
Added tour functionality which can be used to explain the user how things work. Also wrote the first tour for the assignment (format) editor, which automatically starts whenever a new assignment is created.

peek 2019-01-20 00-18

## Summary of changes
- Fixes #278.
- Fixes #345 (Templates are now all in one card, and the 'stackable' part is defined as 'unlimited templates' both in the tour as well as in the template editor itself, including an explanation there.

peek 2019-01-20 00-19

- Renamed the 'format editor' to 'assignment editor' (telling users about a 'format' only causes confusion and is not really relevant).


mjvkeulen:
> We could consider only showing the tutorial once per user, which would require a database call. What are your thoughts on that?

This was my first thought when checking the PR. My suggestion is to add a don't show this again checkbox, which is stored in a preference table in our backend or sth similar?

src/vue/src/views/FormatEdit.vue
```diff
@@ -9,12 +9,41 @@
     <b-row class="outer-container-timeline-page" no-gutters>
         <b-col md="12" lg="8" xl="9" class="inner-container-timeline-page">
             <b-col md="12" lg="auto" xl="4" class="left-content-timeline-page">
-                <bread-crumb v-if="$root.lgMax()" class="main-content">&nbsp;</bread-crumb>
-                <timeline @select-node="selectNode" @add-node="addNode" :selected="currentNode" :nodes="nodes" :edit="true"/>
+                <bread-crumb
+                    v-if="$root.lgMax()"
+                    v-intro="'Welcome to the assignment editor!<br/>This is where you can configure the structure of your assignment. Proceed with this tutorial to learn more.'"
+                    v-intro-step="1">
+                    <icon
```
mjvkeulen: I would prefer a right align of this icon, currently:
image

I would prefer it to be less part of the title.

src/vue/src/views/FormatEdit.vue
```diff
@@ -47,21 +76,28 @@
...
-                <available-template-card
-                    v-for="template in templatePool"
-                    :key="template.t.id"
-                    @click.native="showTemplateModal(template)"
-                    :template="template"
-                    @delete-template="deleteTemplate"/>
-                <b-button class="add-button grey-background full-width multi-form" @click="showTemplateModal(newTemplate())">
+            <div
+            :class="{ 'input-disabled' : saveRequestInFlight }"
+            v-intro="'Every assignment contains customizable <i>templates</i> which specify what the contents of each journal entry should be. There are two different types of templates:<br/><br/><ul><li><b>Unlimited templates</b> can be freely used by students as often as they want</li><li><b>Preset-only templates</b> can be used only for preset entries in the timeline</li></ul>You can preview and edit a template by clicking on it. More templates can be added by clicking the \'Create New Template\' button.'"
```
mjvkeulen: I dont know if the bullet point: 
image
Is sufficiently clear, opinions?

src/vue/package.json
```diff
@@ -15,14 +15,17 @@
   },
   "dependencies": {
     "bootstrap": "^4.1.3",
+    "bootstrap-tour": "^0.12.0",
```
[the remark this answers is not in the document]
engelhamer: Yup, sorry. This was one of the tour packages that I considered but forgot to remove 🙃.

src/vue/src/views/FormatEdit.vue
engelhamer: I decided to put it here because the 'edit' button is already floating right on other pages. This way we can keep things consistent and  they will not clash with one another if we decide to make other tutorials on such pages.

engelhamer: Maybe making it a little bit lighter will do?

src/vue/src/views/FormatEdit.vue
```diff
@@ -9,12 +9,41 @@
...
+                    <icon
+                        v-intro="'That\'s it! If you have any more questions, do not hesitate to contact us via the feedback button below. This tutorial can be consulted again by clicking the question mark.'"
+                        v-intro-step="4"
+                        name="question"
+                        scale="1.75"
+                        class="edit-icon"
+                        @click.native="startTour"/>
+                </bread-crumb>
+                <timeline
+                    v-intro="'The timeline forms the basis for a journal. Assignment details can also be changed here, by clicking the first node.<br/><br/>The timeline contains a node for every entry. You can add two different types of nodes to it:<br/><br/><ul><li><b>Preset entries</b> are entries with a specific template which have to be completed before a set deadline</li><li><b>Progress deadlines</b> are point targets that have to be met before a set deadline</li></ul>New presets can be added via the \'+\' node. Click any node to view its contents.'"
```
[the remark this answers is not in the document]
engelhamer: I agree, but splitting it up into smaller strings and concatenating those looked like a mess as well..

src/vue/src/views/FormatEdit.vue
engelhamer: In contrast to the point directly above it, and with the timeline itself as the follow-up step, I think it should be clear. Maybe we should have some real users test it?

src/vue/src/components/format/FormatEditAvailableTemplate.vue
```diff
@@ -0,0 +1,58 @@
...
+export default {
+    props: ['template'],
+    data () {
+        return {
+            isActive: this.template.available
+        }
+    },
+    methods: {
+        emitDeleteTemplate () {
+            if (confirm('Are you sure you want to delete template "' + this.template.t.name + '" from this format?')) {
```
Larspolo: Maybe change this to assignment as well? And just get rid of the word "format" for the user?

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -18,7 +18,25 @@
...
-            <b-input class="mb-2 mr-sm-2 mb-sm-0 multi-form theme-input" id="template-name" v-model="template.name" placeholder="Template name" required/>
+            <b-card class="no-hover">
+                <b-input class="mb-2 mr-sm-2 mb-sm-0 multi-form theme-input" id="template-name" v-model="template.name" placeholder="Template name" required/>
+                <div v-if="!formatSettings.available" class="template-availability">
+                    <b-button v-on:click.stop @click="toggleActive" class="delete-button">
+                        <icon name="times"/>
+                        Preset-only
+                    </b-button>
+                    <icon name="info-circle"/>
+                    This template can be used only for preset entries in the timeline
```
Larspolo: *can only be used?

src/vue/src/views/FormatEdit.vue
```diff
@@ -9,12 +9,41 @@
...
         <b-col md="12" lg="8" xl="9" class="inner-container-timeline-page">
             <b-col md="12" lg="auto" xl="4" class="left-content-timeline-page">
-                <bread-crumb v-if="$root.lgMax()" class="main-content">&nbsp;</bread-crumb>
-                <timeline @select-node="selectNode" @add-node="addNode" :selected="currentNode" :nodes="nodes" :edit="true"/>
+                <bread-crumb
+                    v-if="$root.lgMax()"
+                    v-intro="'Welcome to the assignment editor!<br/>This is where you can configure the structure of your assignment. Proceed with this tutorial to learn more.'"
+                    v-intro-step="1">
+                    <icon
+                        v-intro="'That\'s it! If you have any more questions, do not hesitate to contact us via the feedback button below. This tutorial can be consulted again by clicking the question mark.'"
```
Larspolo: This suggests that you needs to click the create new template button
image
however when you do it, it creates this awkward moment where you have to finish the tutorial first to start editing the template

src/vue/src/views/FormatEdit.vue
Larspolo: Why use journal here, and not assignment?
"Assignment details can also be changed here, by clicking the first node." -> "To change the name, start date and more details, you need to click on the first node."

"New presets can be added" -> "New nodes can de added"

src/vue/src/views/FormatEdit.vue
mjvkeulen: Possibly, I feel it might disturb me as I interpret it as part of the header itself?

src/vue/src/views/FormatEdit.vue
mjvkeulen: Best way to find out!

src/vue/src/views/FormatEdit.vue
engelhamer: image

Hm, this doesn't seem a very good solution either. I think the darker version is still the most clear, any other thoughts?

engelhamer: image

Having it float on the right causes quite some trouble and would result in inconsistencies with other (future) tutorials

src/vue/src/components/format/FormatEditAvailableTemplate.vue
engelhamer: Definitely!

src/vue/src/views/FormatEdit.vue
engelhamer: Removed.

src/vue/src/views/FormatEdit.vue
engelhamer: Good suggestions, replaced.

src/vue/src/components/template/TemplateEdit.vue
engelhamer: The order is arbitrary here, but changed it anyways. Also expanded the senctence to `This template can only be used for preset entries *you add* to the timeline`

src/vue/src/views/FormatEdit.vue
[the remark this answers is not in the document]
engelhamer: Is that alright with you?

engelhamer:
All preferences (including those already present in the database for grade and comment notifications) have been moved to a separate database table. User preferences are loaded into the local storage upon login, and will be checked from there once the user visits the editor page. If it happens to be the first visit of a user the database preference setting will be updated.

Also the text has been updated to not have any shadows on hover.

src/django/VLE/views/preferences.py
```diff
@@ -0,0 +1,68 @@
...
+        pk -- user ID
+
+        Returns:
+        On failure:
+            unauthorized -- when the user is not logged in
+            not found -- when the user doesn't exist
+        On success:
+            success -- with the preferences data
+        """
+        if int(pk) == 0:
```
mjvkeulen: A user should always be logged in when retrieving preferences, as such I think its fair to enforce the use of a `pk`? Working with a default of 0 would not be needed as a user always knows his `id`.

src/django/VLE/views/preferences.py
```diff
@@ -0,0 +1,68 @@
...
+        On failure:
+            unauthorized -- when the user is not logged in
+            forbidden -- when the user is not superuser or pk is not the same as the logged in user
+            not found -- when the user doesnt exist
+            bad request -- when the data is invalid
+        On success:
+            success -- with the updated preferences
+        """
+        pk, = utils.required_typed_params(kwargs, (int, 'pk'))
+        if pk == 0:
```
mjvkeulen: Same thing would apply here, the id should always be known?

src/vue/src/api/preferences.js
```diff
@@ -0,0 +1,8 @@
+import auth from '@/api/auth'
+
+export default {
+    update (id, data, connArgs = auth.DEFAULT_CONN_ARGS) {
+        return auth.update('preferences/' + id, data, connArgs)
```
mjvkeulen: I feel we should start working with `preferences/${id}` like syntax

src/django/VLE/models.py
```diff
@@ -207,6 +201,37 @@ def to_string(self, user=None):
...
+
+class Preferences(models.Model):
+    """Preferences.
+
+    Describes the preferences of a user:
+    - show_format_tutorial: whether or not to show the assignment format tutorial.
+    - grade_notifications: whether or not to receive grade notifications via email.
+    - comment_notifications: whether or not to receive comment notifications via email.
+    """
+    user = models.OneToOneField(
```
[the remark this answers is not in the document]
engelhamer: If I understand correctly that would allow for behaviour similar to the default `XXX_set` syntax. Wouldn't you want something like that everywhere then?

src/django/VLE/serializers.py
```diff
@@ -103,6 +103,13 @@ def get_permissions(self, user):
         return perms
 
 
+class PreferencesSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Preferences
+        fields = '__all__'
+        read_only_fields = ('id', )
```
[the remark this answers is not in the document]
engelhamer: Hm, maybe making the `user` explicit would be better here.

src/vue/src/api/preferences.js
engelhamer: Personally I dislike the required backticks (which are required there) instead of quotes in that syntax.. If we want to do this I think it should be done for every api call, not just the single one in this pull request in my opinion.

src/vue/src/store/modules/user.js
```diff
@@ -148,7 +142,13 @@ const actions = {
         return new Promise((resolve, reject) => {
             connection.conn.get('/users/0/').then(response => {
                 commit(types.HYDRATE_USER, response.data)
-                resolve('Store is populated successfully')
+                connection.conn.get('/preferences/' + response.data.user.id).then(response => {
+                    commit(types.HYDRATE_PREFERENCES, response.data)
```
mjvkeulen: This hydrates the preferences in the user module, not in the preference module. Solution would be:
commit(`preferences/${types.HYDRATE_PREFERENCES}`, response.data, { root: true })

Note that the current get request is nested conditionally to the user hydrate, which is functionally not required. This obviously slower, but I dont think the user will ever notice.

mjvkeulen: Prob also means the hydrate preferences in the user module can be removed.

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: Hm `XXX_set`? It does open the backwards relation by means of `user.preference` (when dealing with a user instance).

src/vue/src/store/modules/user.js
engelhamer: I didn't notice the distinction, thanks for that!

src/vue/src/components/format/FormatEditAvailableTemplate.vue
```diff
@@ -0,0 +1,58 @@
...
+
+export default {
+    props: ['template'],
+    data () {
+        return {
+            isActive: this.template.available
+        }
+    },
+    methods: {
+        emitDeleteTemplate () {
```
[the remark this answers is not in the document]
engelhamer: How strange, I will look into that.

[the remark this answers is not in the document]
engelhamer: Turns out this is also the case on the development branch. I did not change anything related to this problem in this pull request.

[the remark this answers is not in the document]
engelhamer: Fixed anyways.

PR #548, opened 2019-01-21
Removed unnecessary data from participation serializer

Larspolo:
## Description
Removed unnecessary data from participation serializer and also added a 0 when points are null


src/vue/src/components/assets/ProgressBar.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <div>
         <h4 class="progress-percentage float-right">{{ progressPercentage }}%</h4>
-        <h5 class="progress-percentage"><b>{{ currentPoints }}</b> Points</h5><br/>
+        <h5 class="progress-percentage"><b>{{ currentPoints ? currentPoints : 0 }}</b> Points</h5><br/>
```
mjvkeulen: As the format of multiple queries is changed by removing the depth = 1 serialization, are we sure this is the only required change?

Larspolo: This was just an extra thing that I found

Larspolo: Everything else should be good

PR #562, opened 2019-02-03
Fullname support only, no first and last name anymore

### LINKED ISSUE #402: Stop using first and last name for users

Larspolo:
## Description
Migrated from first and last name to full name

## Summary of changes
- Fixes #402


src/django/VLE/models.py
```diff
@@ -115,13 +115,18 @@ class User(AbstractUser):
...
     - email: email of the user.
     - verified_email: Boolean to indicate if the user has validated their email adress.
     - USERNAME_FIELD: username of the username.
     - password: the hash of the password of the user.
     - lti_id: the DLO id of the user.
     """
 
+    full_name = models.CharField(
+        null=True,
+        max_length=50
```
mjvkeulen: Currently, first name and last name can be of a total length of 180 chars, add the space and we need a higher max length.

src/django/VLE/migrations/0012_user_full_name.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.2 on 2018-11-06 15:07
```
mjvkeulen: 2x 0012? Can we not just merge these manually into one migration file?

src/django/VLE/models.py
```diff
@@ -115,13 +115,18 @@ class User(AbstractUser):
...
+    - full_name: full name of the user
     - email: email of the user.
     - verified_email: Boolean to indicate if the user has validated their email adress.
     - USERNAME_FIELD: username of the username.
     - password: the hash of the password of the user.
     - lti_id: the DLO id of the user.
     """
 
+    full_name = models.CharField(
+        null=True,
```
mjvkeulen: This is probably not desired right? I understand this creates complications for the migration files, having to first add the field with `null=True`, run some python, change the field. But afaik this is the correct way to handle this.

src/django/VLE/utils/email_handling.py
```diff
@@ -68,7 +68,7 @@ def send_email_feedback(user, files, topic, ftype, feedback, user_agent, url):
     f_body += "THE FEEDBACK:\n{}".format(feedback[0])
 
     r_subject = "[eJournal] Submitted feedback"
-    r_body = "Hi {},\n\n".format(user.first_name)
+    r_body = "Hi {},\n\n".format(user.full_name)
```
engelhamer: This sentence is now a bit weird: A casual (rather informal) `Hi!`, but followed by the (quite formal sounding) full name of the user. Unfortunately I do not see a clear solution for this so I guess we will just keep it this way.

PR #577, opened 2019-02-06
Improved entry handling between due and lock date

### LINKED ISSUE #530: Entry handling in between due and lockdate
**Describe the bug**
If a student creates an entry past the due date but before the lock date all is well. This entry however can no longer be edited. The editing succeeds front end, but the server sends an error.

**Expected behavior**
- Editing should still be possible in between due and lock date
- Front end should reset content to last saved if not allowed to edit.

engelhamer:
## Description
Entries can now be added until the lock date, and preset entries whose deadline has already passed won't show an edit button anymore.

## Summary of changes
- Fixes #530


src/django/VLE/models.py
```diff
@@ -709,7 +709,7 @@ class Entry(models.Model):
     )
 
     def is_due(self):
-        return (self.node.preset and self.node.preset.is_due()) or self.node.journal.assignment.is_due()
+        return (self.node.preset and self.node.preset.is_due()) or self.node.journal.assignment.is_locked()
```
Larspolo: Why isnt it due when assignment is due?

src/django/VLE/models.py
```diff
@@ -660,7 +660,7 @@ class PresetNode(models.Model):
     )
 
     def is_due(self):
-        return self.deadline < now() or self.format.assignment.is_due()
+        return self.deadline < now()
```
Larspolo: Why isnt it due when assignment is due?

engelhamer: The deadline of a preset node can only ever be before the due date of an assignment. Additionally, a preset node must always have a deadline. Therefore this extra check was useless and only confusing.

src/django/VLE/models.py
engelhamer: The old situation resulted in a situation where entries could no longer be modified once the assignment is due. However, students should still be able to modify them until the assignment is locked. Unless the entry corresponds was preset by a teacher with a given due date, that is.

Issue #294, opened 2018-09-02
Additional assignment grading, separate from entry grades.


mjvkeulen:
# Feature request

## Description
It is possible student using our platform earn credits for their assignments separate from the actual entries. For example taking points with them from the previous year. An intuitive solution that allows teachers to allocate these points is required. As well as a solution to display these additional points in a sensible manner.

## Suggestion
We can display these points at the start of the assignment node. This will allow for simple adjustments and will fit well thematically.
An interface for the allocation is also required, preferably one that allows for bulk allocation with the same description. 

## Requirement
Bulk input should support CSV files consisting of usernames and points. Points should then be added to their journals automatically.

PR #579, opened 2019-02-09
Bonus points

Larspolo:
## Description
Added the possibility for graders to add a bonus grade to a journal

# Fixes
- #294 

#### Teacher
image

#### Student
image


engelhamer:
I would suggest adding the bonus part tot the student card (so above the point count and under the name) rather than its current location.

engelhamer:
Also a question mark with tooltip would be helpful. It might not immediately become clear what the function of these points is.

mjvkeulen:
Currently, these points magically appear to a student. I would like to be able to add a bonus node to the timeline as an educator, consisting of a rich-text description and bonus points.

We also require a mass import function, consisting of:
- username/email
- bonus points
- date
- description

engelhamer:
Personally I am not a fan of a very extensive bonus system: a simple field out here is fine, in addition to a bulk import feature. I feel like more complicated stuff is not worthwile, since most 'bonus' points will simply be added to a particular entry anyways.

We can still add a node to a student's journal if they were granted bonus points (so a single one, all the way at the start) or include that in the start node.

src/vue/src/components/assets/Tooltip.vue
```diff
@@ -0,0 +1,21 @@
+<template>
```
engelhamer: I should've done this right away :upside_down_face:

src/django/VLE/views/assignment.py
```diff
@@ -273,6 +274,44 @@ def upcoming(self, request):
...
+                bonus = float(bonus)
+                user = User.objects.get(username=str(username))
+                journal = Journal.objects.get(assignment=assignment, user=user)
+                journal.bonus_points = bonus
+                journal.save()
+            except ValueError:
+                return response.bad_request(
+                    'The bonus file is not correctly formatted. Please use "[username],[bonuspoints]"')
+            except User.DoesNotExist:
+                return response.bad_request(
```
engelhamer: I think we should go on with parsing the file on a not found user, and simply print a list of wrong lines afterwards. Currently all points up until a wrong name are saved, but the rest is not (which seems unlogical).

mjvkeulen: I believe nothing should be saved unless the format is correct for all entries, how the error handling should take place is debatable. I'll push a suggestion, see what you guys think.

mjvkeulen: The toast will ignore string formatting, so an error output file is required. Demanding more custom upload field logic.

src/vue/src/views/Assignment.vue
```diff
@@ -32,11 +32,20 @@
             </div>
             <b-button
                 v-if="$hasPermission('can_publish_grades') && assignmentJournals.length > 0"
-                class="add-button full-width"
+                class="add-button full-width multi-form"
                 @click="publishGradesAssignment">
                 <icon name="upload"/>
                 Publish all Grades for this Assignment
             </b-button>
+            <file-upload-input
```
mjvkeulen: Currently a page refresh is required after both an incorrect upload (in order to clear the file upload button) or after a successful upload in order to see the bonus points applied.

mjvkeulen:
So the toast ignores newlines and is shown too briefly. Ideally we want a file with error output. This requires some tweaks to the file upload component. We are really increasing the scope haha

engelhamer:
A modal that shows a list (returned from the API call) would do right. I think a file would be too much of a hassle.

engelhamer:
image

Added a modal with some more explanation. Especially the fact that points are overwritten is important to note.

engelhamer:
Currently the server crashes if a user uploads something that is not a csv (e.g. an image). This should be fixed.

engelhamer:
Errors are now displayed as follows.

out-11

src/vue/src/components/assets/file_handling/FileUploadInput.vue
```diff
@@ -25,19 +25,22 @@ export default {
...
+            default: 'No file chosen'
         },
         contentID: {
             default: null
         }
     },
     data () {
         return {
-            placeholderText: 'Select a file.',
+            placeholderText: 'No file chosen',
```
Larspolo: I preferred "select a file." Why this change?

engelhamer: Because it normally displays the file name in this location once chosen. This seemed more intuitive to me.

PR #583, opened 2019-02-14
Rich text field descriptions

### LINKED ISSUE #470: Rich text field description
Rich text functionality for template field descriptions.

**User story**
As a user I want to be able to use rich text in field descriptions, e.g. for bullet lists or markup.

**User interface**
Preferably the inline version of the editor, for the sake of space inside a card.

### LINKED ISSUE #478: changing id's of richtexteditors and then updating setcontent makes te textarea unclickable sometimes
**Describe the bug**
When using the same rich-texteditor component multiple times, by changing the id's in the meantime (when switching between modals for example) the textarea will become unusable after sometime. 

More technical details:
This happens after using the tinymce setContent function in the id watcher, but sometimes it will work correctly and sometimes not.

**To Reproduce**
0. Go to the branch enhancement/rich-descriptions (This contains the changed template modals where the bug is visible)
1. Go to the format-editor
2. Keep switching between two template modals (for example the default text and the colliqium templates)
3. After a while in one or both of the modals, the first rich-texteditors textarea will become unclickable in the modal, also no content will be rendered.
4. If displayInline is set to false, you can click on the menu bar and it will switch back to the right format, with a clickable textarea and the loaded content. If displayInline is set to true, there is nothing, except reloading the page that will fix the texteditor.

**Expected behavior**
It should display the richtext-editor in a correct state, where the textarea is clickable and the content is loaded in properly.

**Screenshots**
After switchin one time it works correctly:
image

Try to switch back to default text with inlinemode on true, the textarea isn't loading the content and not clickable:
image

After switching two times with displayinline on false (working correctly):
image

After switching a few more times, the text area isn't loading the content and not clickable:
image

engelhamer:
## Description
Field descriptions now contain rich text content.

## Summary of changes
- Fixes #470 
- Fixes #478 
- Changed the look of todo cards

image


src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -128,15 +139,22 @@ export default {
...
     },
     components: {
         'content-single-column': ContentSingleColumn,
         icon,
         'draggable': draggable,
+        'text-editor': textEditor,
         'template-preview': TemplatePreview
     },
+    computed: {
+        tt () {
```
mjvkeulen: debug remnant?

engelhamer: Good spot, should be removed indeed.

src/vue/src/components/journal/JournalStartCard.vue
```diff
@@ -1,19 +1,33 @@
...
+                Due date: {{ $root.beautifyDate(assignment.due_date) }}<br/>
+            </b>
+            <b v-if="assignment.lock_date">
+                <span v-if="new Date(assignment.lock_date) < new Date()">
+                    This assignment has been locked<br/>
+                </span>
+                Lock date: {{ $root.beautifyDate(assignment.lock_date) }}<br/>
+            </b>
+            <b v-if="assignment.points_possible">
+                Points possible: {{ assignment.points_possible }}<br/>
```
Larspolo: Maybe a tooltip or rewrite "points possible" it may not be descriptive enough.

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -5,36 +5,38 @@
...
-            <span class="text-red" v-else>{{ timeLeft[1] }} late</span>
-            <span class="float-right">{{ $root.beautifyDate(deadline.deadline) }}</span>
-        </h6>
+        <span v-if="this.deadline.deadline">
+            <!-- Teacher deadline shows last submitted entry date  -->
+            <span v-if="this.deadline.stats.needs_marking + deadline.stats.unpublished">
+                <icon name="eye" class="fill-grey shift-up-3"/> {{ timeLeft[1] }} ago<br/>
+            </span>
+            <!-- Student deadline shows last not submitted deadline -->
+            <span v-else>
```
Larspolo: Missing eye icon

PR #592, opened 2019-02-27
Added sort by points

### LINKED ISSUE #568: Sort by amount points
In the assignment overview a possibility to sort the journals on points gathered would be nice to have.

**User story**
As a teacher go to the assignment overview
Sort by points gathered
See the most amount of points on top

Larspolo:
## Description
Added sort by points in Assignment overview

image

## Summary of changes
- Fixes #568


src/vue/src/store/constants/preference-types.js
```diff
@@ -6,7 +6,8 @@ export const TODO_SORT_OPTIONS = new Set([
 export const JOURNAL_SORT_OPTIONS = new Set([
     'name',
     'username',
-    'markingNeeded'
+    'markingNeeded',
+    'points'
```
mjvkeulen: Doesn't the store need to be updated as well to keep working with persistent filters?

engelhamer: Completely forgot about that! Guess it should.

Larspolo: It already gets updated

PR #594, opened 2019-02-28
Styled emails

### LINKED ISSUE #270: Styled emails
# Feature request

## Status quo
Emails are currently plain text and purely functional, this can be much improved. 

## Description
Emails send by eJourn.al should visually be representative of the applications theme.

## User story
A user excepts an email verification mail, or a password reset link to look professional and inline with the theme of the corresponding website.

## User interface
Something like this.
image

engelhamer:
## Description
Added HTML email templates in eJournal style. Will also proceed with some general UI tweaks to the application itself in the near future that resemble the style of this more (i.e. rounded corners).

### Call to action template
Can be used for things like comment notifications (click to view comment), email verification (click to verify) etc.

image

### Generic template
Can be used for several paragraphs in a single email. Currently in use for the feedback feature.

image

### Compatibility
Since lots of email clients do not support external web fonts, we kinda have to be in luck for _Roboto_ to be displayed. This is the case on Android devices of course, so there's that, but most of the times the user will see a (similar) font. Currently _Helvetica_ or _Arial_.

Mobile scaling works fine as well. _Note_: the example below is on an Android Gmail client and thus should be able to show the _Roboto_ font unlike the above examples.

image

## Summary of changes
- Fixes #270


src/django/VLE/utils/email_handling.py
```diff
@@ -1,101 +1,122 @@
...
+    r_email_data['heading'] = 'Development feedback'
+    r_email_data['main_content'] = []
+    r_email_data['main_content'].append('Thank you for your feedback! Below you will find a copy:')
+    r_email_data['main_content'].append(feedback[0])
+    if len(files) > 0:
+        r_email_data['main_content'].append('Uploaded attachments have been added to this e-mail as well.')
+    r_email_data['main_content'].append('''
+    We might reply to your feedback to ask some questions or just to say thanks!''')
+    r_email_data['main_content'].append('Kind regards,')
+    r_email_data['main_content'].append('The eJournal Team')
```
Larspolo: This e in eJournal is not blue

src/django/VLE/utils/email_handling.py
```diff
@@ -1,101 +1,122 @@
...
 
     attachments = []
     for file in files:
         attachments.append((file.name, file.read(), file.content_type))
 
-    reply = EmailMessage(
-        subject=r_subject,
-        body=r_body,
+    reply = EmailMultiAlternatives(
+        subject='Thanks for your feedback!',
```
Larspolo: Thank you for the (or your) feedback*

src/django/email_templates/call_to_action.html
```diff
@@ -0,0 +1,77 @@
...
+                                        </tr>
+                                    </table>
+                                </td>
+                            </tr>
+                        </table>
+                    </tr>
+                </table>
+                <p style="color: grey; text-align: center">
+                    Sent by  <a href="https://www.eJournal.app" style="color: grey; text-decoration: none"><span style="color: #22648A">e</span>Journal</a>
+                    - Don't like these emails? <a href="{{ email_data.profile_url }}" style="color: grey; text-decoration: underline">Update your preferences</a>.
```
Larspolo: Currently these cannot be edited, maybe we can leave this out for now?

src/django/VLE/utils/email_handling.py
```diff
@@ -1,101 +1,122 @@
...
-{recovery_link}\
-'''.format(recovery_link=recovery_link)
-
-    email = EmailMessage(
-        subject='eJourn.al password recovery',
-        body=email_body,
+    email_data['main_content'] = '''\
+    We have received a request for password recovery. If it was you who made this request, \
+    please click the button below to set a new password. If you have not made this \
+    request please ignore this email.'''
```
Larspolo: I would prefer if this was extracted to a file as well.

engelhamer: So add a specific HTML template for the feedback then? I chose to reuse as much as possible but agree that this looks a bit cluggish (and can therefore deserve its own template). Happy to add it!

src/django/VLE/utils/email_handling.py
engelhamer: Since we don't parse HTML from these variables, but if we make a separate feedback template it can be blue ;)

src/django/VLE/utils/email_handling.py
engelhamer: Fine by me.

src/django/email_templates/call_to_action.html
engelhamer: Right.

src/django/VLE/settings/base.py
```diff
@@ -82,7 +82,7 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': (os.path.join(BASE_DIR, 'email_templates'),),
```
mjvkeulen: Should we nest email_templates in a template directory to be future proof?

src/django/VLE/utils/email_handling.py
mjvkeulen: What is the benefit? It might look a bit cleaner, but its clear and leaves one dedicated space to work in when changing a password recovery link for example?

src/django/email_templates/generic.html
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html>
+
+<head>
```
mjvkeulen: Django's templates are quite powerful, I would suggest moving the head to a dedicated template for example and include it where needed. E.g.  `{% include "templates/email_templates/head.html" %}`

Similar deconstruction can be applied, making Lars his request for a blue eJournal possible without too much repetition.

src/django/VLE/settings/base.py
engelhamer: Possible, but what other kind of templates are we going to use?

src/django/email_templates/generic.html
engelhamer: Good idea!

src/django/VLE/settings/base.py
mjvkeulen: I think we will stick to our full split and keep our vuejs frontend out of all templates. But if we ever want to approach something differently, maybe using jinja2 templates for rendering speed I could see that happening. But no concrete examples for now!

src/django/VLE/utils/email_handling.py
engelhamer: I'm also in favor of keeping some things that only use a single main content in this file. Extracting the feedback part seems more logical (and I will do that).

engelhamer:
Implemented requested changes. Feedback mails now look like this:

image

src/django/templates/email_templates/feedback.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html>
+
+{% include "head.html" %}
+
+<body>
+    <table style="width: 100%; font-family: 'Roboto', 'Helvetica', 'Arial', sans-serif; padding: 20px">
+        <tr>
+            <td style="padding: 10px" align="center">
+                <table style="width: 100%; max-width: 600px">
```
mjvkeulen: Maybe also add a logo include?

engelhamer: Good idea, also more compatible with future changes (if we get a real logo with more than just text).

PR #601, opened 2019-03-04
Upcoming deadline notifications

### LINKED ISSUE #600: Upcoming deadline notification
A notification when a student has an upcoming deadline in their journal that is not yet finished.

This should be checked
- One week in advance
- One day in advance

**User interface**
A call to action email with a link to the corresponding entry.

**Related issue**
This is part of #213.

engelhamer:
## Description
Email notifications for upcoming deadlines.

image

## Summary of changes
- Fixes #600


src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _send_deadline_mails(deadline_query):
+    """_send_deadline_mails
+
+    This sends mails to the users who are connected to the PresetNodes query that is send with.
+
+    Arguments:
+    deadline_query -- query of PresetNodes
+    """
+    deadlines = deadline_query.filter(node__journal__user__preferences__upcoming_deadline_notifications=True,
```
mjvkeulen: We also need to filter whether the deadline has already been met. Because no one wants to receive notifications about upcoming work that's already been finished.

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,80 @@
...
+                                      node__journal__user__verified_email=True)\
+                              .values('node__journal', 'deadline')
+
+    for deadline in deadlines:
+        journal = Journal.objects.get(pk=deadline['node__journal'])
+        # TODO: Check if the preset node does not yet have an entry (in which case no mail should be sent).
```
mjvkeulen: I don't quite follow this TODO

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,80 @@
...
+        # TODO: Replace student with can have journal permission?
+        if Participation.objects.filter(user=journal.user, course__in=journal.assignment.courses.all(),
+                                        role__name='Student').exists():
+            print(journal.user.email, deadline['deadline'])
+
+            assignment = journal.assignment
+            course = assignment.courses.all().first()
```
mjvkeulen: This could yield a course the student is no participant in. Filter this based on participation, then select the course with the most recent creation data or something else time-based?

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,80 @@
...
+            email_data = {}
+            email_data['heading'] = 'Upcoming deadline'
+            email_data['main_content'] = '''\
+            You have an unfinished deadline coming up for {} in {}.'''.format(course.name, assignment.name)
+            email_data['extra_content'] = 'Date: {}'.format(deadline['deadline'])
+            # TODO: Replace node ID
```
mjvkeulen: This makes sense, but we currently don't process `nID` query string parameters right? So we would need to add this, including the select the nID from query param functionality frontend.

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,80 @@
...
+def _send_deadline_mails(deadline_query):
+    """_send_deadline_mails
+
+    This sends mails to the users who are connected to the PresetNodes query that is send with.
+
+    Arguments:
+    deadline_query -- query of PresetNodes
+    """
+    deadlines = deadline_query.filter(node__journal__user__preferences__upcoming_deadline_notifications=True,
+                                      node__journal__user__verified_email=True)\
```
mjvkeulen: I believe chaining filters like this is incorrect, this query will select rows with users who have either notifications=True *or* verified_email=True instead of both.

src/django/VLE/tasks/beats/notifications.py
engelhamer: That's what that other TODO (on the entry in a node) is about

src/django/VLE/tasks/beats/notifications.py
engelhamer: Check if a student has already filled in the entry.

src/django/VLE/tasks/beats/notifications.py
engelhamer: We do process it. This is also used for directly pointing the user to a node via an LTI submission link!

src/django/VLE/tasks/beats/notifications.py
engelhamer: I think that your previous comment on the fact that this could yield a course the user does not participate in is correct. That is more important to me than which course in particular is used.

src/django/VLE/tasks/beats/notifications.py
Larspolo: `node__entry__vle_coupling=Entry.NEED_SUBMISSION` sometimes (like more than we would like) fails to get updated on submission. Therefore I changed it to a more foolproof check

src/django/VLE/tasks/beats/notifications.py
Larspolo: `Participation.objects.filter(user=journal.user, course__in=journal.assignment.courses.all(), role__can_have_journal=True).order_by('-startdate').first().course` would work.

src/django/VLE/tasks/beats/notifications.py
mjvkeulen: Remind me to ask for an explanation of why it fails during standup

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,89 @@
...
+def _send_deadline_mails(deadline_query):
+    """_send_deadline_mails
+
+    This sends mails to the users who are connected to the PresetNodes query that is send with.
+
+    Arguments:
+    deadline_query -- query of PresetNodes
+    """
+    # Remove all filled entryedeadline, and remove where the user does not want to recieve an email.
+    no_submissions = Q(type=Node.ENTRYDEADLINE, node__entry__isnull=True) | Q(type=Node.PROGRESS)
```
mjvkeulen: How do the `Node.types` affect the no_submissions category here?

src/django/VLE/tasks/beats/notifications.py
```diff
@@ -0,0 +1,89 @@
...
+    notifications_enabled = Q(node__journal__user__preferences__upcoming_deadline_notifications=True)
+    verified_email = Q(node__journal__user__verified_email=True)
+    deadlines = deadline_query.filter(notifications_enabled & verified_email & no_submissions)\
+                              .values('node', 'node__journal', 'deadline', 'type', 'target')
+    for deadline in deadlines:
+        journal = Journal.objects.get(pk=deadline['node__journal'])
+        # Only send to users who have a journal
+        if not Participation.objects.filter(user=journal.user, course__in=journal.assignment.courses.all(),
```
mjvkeulen: This filter in the nested loop is pretty heavy, could we work this into the query itself as we already have access to the assignment and user?

Larspolo: Progress node is only to indicate what amount of points have to be reached. This does not include an entry itself, its merely the sum of previous entries

src/django/VLE/tasks/beats/notifications.py
Larspolo: Progress node is only to indicate what amount of points have to be reached. This does not include an entry itself, its merely the sum of previous entries

src/django/VLE/tasks/beats/notifications.py
Larspolo: Will have to improve my query skills for that. This requires some effort which we can postpone to after the deploy.

PR #605, opened 2019-03-08
Enhancement/nosniff

### LINKED ISSUE #344: Description of assignment not loading

mjvkeulen:
## Description
Hardened the Nginx settings by the addition of the `X-Content-Type-Options "nosniff"`
Worked through all the TextEditor todos in the process.

## Summary of changes
- TextEditor now supports `v-model`, seems like not loading descriptions are fixed in the process of cleanup...
- TextEditor `minifiedTextArea` property:
image
- As you can see, the text editor now also supports `placeholder` text.
- Fixed compatibility issues with modals
- No more ``tinymce.XX Resource interpreted as Stylesheet but transferred with MIME type text/html: XXX`` warnings
- Skins are actually applied, time for a custom skin at some point as we adopt the code with the PR anyway: http://skin.tinymce.com/
- fixes #344, but this will still require staging testing. No issues detected whilst using low traffic debug functions.


engelhamer:
Does the `v-model` directive solve #344?

engelhamer:
Also: what's with the whitespace above the placeholder?

mjvkeulen:
> Also: what's with the whitespace above the placeholder?

Thats the default `.mce-edit-area` padding, something which we should make our own at some point. I could not get the selector right, but such changes could be passed to the `content_style` config data of the TextEditor component.

engelhamer:
Nice to see #344 has actually been fixed!

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -47,9 +47,10 @@
                                 v-if="showEditors"
                                 :id="`rich-text-editor-field-${template.id}-${field.location}`"
                                 :basic="true"
-                                :footer="false"
-                                :givenContent="field.description ? field.description : ''"
-                                @content-update="field.description = $event"
+                                :displayInline="true"
+                                :minifiedTextArea="true"
+                                placeholder="Optional description"
```
engelhamer: The inline display looks a bit funny: there is way more padding at the top than necessary, but less at the sides. Compare:

image

with:

image

The location of the placeholder should be the same as regular text.

Also, when you click the texteditor above the placeholder text, a cursor appears while the placeholder is not removed. The formatting tools also do not become available then.

engelhamer: Additionally, the placeholder is in regular Roboto whereas the actual description is in Roboto Condensed. I think the texteditor as a whole should use only Roboto, since that will be used when the contents are displayed as well.

src/vue/src/components/assets/TextEditor.vue
```diff
@@ -112,7 +112,8 @@ export default {
                 content_style: `
                     @import url('https://fonts.googleapis.com/css?family=Roboto+Condensed|Roboto:400,700');
                     body {
-                        font-family: "Roboto Condensed";
+                        font-family: 'Roboto Condensed';
```
engelhamer: Should add `sans-serif` as a backup.

src/vue/src/components/assets/TextEditor.vue
```diff
@@ -299,6 +301,12 @@ export default {
                 {start: '* ', cmd: 'InsertUnorderedList'},
                 {start: '- ', cmd: 'InsertUnorderedList'}
             ]
+        },
+        /* NOTE: Called from parent. */
+        clearContent () {
+            this.editor.setContent('')
+            this.editor.execCommand('fontName', false, 'roboto condensed', {skip_focus: true})
```
engelhamer: Should the font name not be capitalised?

PR #617, opened 2019-03-14
Resize sandboxed iframes

### LINKED ISSUE #616: Sandboxed iframe content sometimes disappears on resize
**Describe the bug**
The content of sandboxed iframes (used for displaying rich text) can disappear from view upon resizing the application browser window.

**To Reproduce**
1. Go to a journal entry with a (long) rich text content
2. Resize the window
3. Observe that some text disappears

**Expected behavior**
Everything to remain visible: the frame should maintain appropriate size.

**Screenshots**
Before:
image

After:
image

**Desktop (please complete the following information):**
 - OS: Windows 110
 - Browser: Google Chrome
 - Version 73.0.3683.75

**Additional context**
Probably has to do with the js resize of the window upon load. Something like that should also happen on resize.

mjvkeulen:
## Description
Moved window width flags to computed properties. Rescale sandboxed iframes on window resize.

- fixes #616


src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -19,21 +19,36 @@ export default {
             required: true
         }
     },
+    data () {
+        return {
+            iframe: null
+        }
+    },
+    computed: {
+        // QUESTION: How does one watch this.$root.windowWidth directly?
```
engelhamer: Wouldn't know from the top of my head.

mjvkeulen:
> image
> 
> When I visit this page for the first time (reload also), it now shows up like this. This is not the case on develop. After a window resize, it goes back to normal, but still

I can't reproduce your issue on chrome or firefox, could you look into what's causing it for your machine?

PR #621, opened 2019-03-16
Post 1.1 deploy tweaks

### LINKED ISSUE #615: Richt text editor placeholder does not appear on subsequent use
**Describe the bug**
If a placeholder has been set for a rich text editor (e.g. for the comment card) it disappears once you type in that boxe. However, if the box is cleared (e.g. when sending the comment) the placeholder does not re-appear again.

**To Reproduce**
1. Look at placeholder in comment card
2. Write a comment
3. Click on 'send'
3. Observe that here is no placeholder, just an empty editor

**Expected behavior**
The placeholder to re-appear.

mjvkeulen:
# Summary of changes
- Fixes #615
- Additional tasks for logwatch
- Ansible security role touch up
- Ansible compatibility with debian stretch
- Sort by name fix
- Sync with required tweaks for deploy


src/django/VLE/settings/production.py
```diff
@@ -39,7 +39,8 @@
 SECURE_BROWSER_XSS_FILTER = True
 
 DEBUG = False
-SECURE_SSL_REDIRECT = True
+# TODO Reenable
```
engelhamer: Should this not be investigated first?

system_configuration_tools/roles/web/templates/maintenance_index.html.j2
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<title>Site Maintenance</title>
+<style>
+  body { text-align: center; padding: 150px; }
+  h1 { font-size: 50px; color #22648A; }
+  body { font: 20px Helvetica, sans-serif; color: #252C39; }
```
engelhamer: Would be nice to use the Roboto font here as well (just add the import). Also maybe add a logo and use our own background.

mjvkeulen: Tweak it to your liking

mjvkeulen: Don't forget that for a normal deploy this will be up for upto 10 minutes

src/django/VLE/settings/production.py
mjvkeulen: Default is False, which is what was active on the previous deploy (along with many other softer settings to be fair). Nginx forces any incoming to SSL, but ideally this should fixed as well.

src/django/VLE/lti_launch.py
```diff
@@ -37,6 +37,13 @@ def is_valid(self, request):
         consumers secret en key
         """
         method, url, head, param = self.parse_request(request)
+
+        # Disable when SSL_REDIRECT = True
+        url_split = url.split(':')
+        if url_split[0] == 'http':
+            url_split[0] = 'https'
+            url = ':'.join(url_split)
```
Larspolo: Have you checked whether or not the data it is still sent with https from Canvas to the server?

mjvkeulen: Nginx forwards https requests with http internally. When Django redirected these requests to https with `SECURE_SSL_REDIRECT` we arrived in a loop. By making Django recognize the redirect header (`SECURE_PROXY_SSL_HEADER`) with `https` set as secure, we now avoid the loop and can effectively tell if the request is secure (`.is_secure()`).

PR #622, opened 2019-03-18
Permission can edit staff comment

### LINKED ISSUE #613: Allow teacher to manage assistant's comments
The ability for a teacher to edit or delete comments made by their teaching assistants.

mjvkeulen:
- Fixes #613


src/django/VLE/factory.py
```diff
@@ -320,7 +320,8 @@ def make_role_default_no_perms(name, course, can_edit_course_details=False, can_
                                can_delete_course_user_group=False, can_edit_course_user_group=False,
                                can_add_assignment=False, can_delete_assignment=False, can_edit_assignment=False,
                                can_view_all_journals=False, can_grade=False, can_publish_grades=False,
-                               can_have_journal=False, can_comment=False, can_view_unpublished_assignment=False):
+                               can_have_journal=False, can_comment=False, can_edit_staff_comment=False,
+                               can_view_unpublished_assignment=False):
```
Larspolo: This really needs a rework one day

src/django/VLE/models.py
```diff
@@ -932,6 +933,29 @@ class Comment(models.Model):
...
+        Raises a VLEProgramming error when misused.
+        """
+        if not isinstance(user, User):
+            raise VLEProgrammingError("Expected instance of type User.")
+
+        if user == self.author or user.is_superuser:
+            return True
+
+        return user.has_permission('can_edit_staff_comment', self.entry.node.journal.assignment) and \
+            not self.author.has_permission('can_have_journal', self.entry.node.journal.assignment)
```
Larspolo: I think its better to check on `not self.author == self.entry.node.journal.user`

src/django/VLE/views/comment.py
```diff
@@ -139,14 +139,16 @@ def partial_update(self, request, *args, **kwargs):
...
-        if not (comment.author.id == request.user.id or request.user.is_superuser):
+        if not comment.can_edit(request.user):
             return response.forbidden('You are not allowed to edit this comment.')
 
+        comment.last_edited_by = request.user
+        comment.save()
         text, = utils.required_params(request.data, 'text')
         serializer = CommentSerializer(
-            comment, data={'text': text}, partial=True)
+            comment, data={'text': text, 'last_edited_by': request.user}, partial=True)
```
Larspolo: This can be removed right? Its already stored above

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -125,6 +129,9 @@ export default {
         this.setComments()
     },
     methods: {
+        canEditComment (commentAuthorID) {
+            return this.$hasPermission('can_edit_staff_comment') || this.$store.getters['user/uID'] === commentAuthorID
```
Larspolo: As a teacher I now get an edit button for the students comment, but I cannot edit them

engelhamer: Would be the easiest to show it for everybody but the author of the journal in that case (since no other people with the permission to have a journal themselves can view other journals, let alone comment on them).

src/django/VLE/views/comment.py
mjvkeulen: Hah the serializer field is read only as well, left over from my struggles.

src/vue/src/components/journal/CommentCard.vue
mjvkeulen: Is it ok if I serialize the can_have_journal permission along with a comment? Or how would you guys like to see this fixed?

engelhamer: That would be unnecessary if you add a check for the journal author front-end.

src/django/VLE/models.py
mjvkeulen: As the permission is on a per assignment level, if we only check if the author of the comment is not equal to the one editing staff from unrelated courses with the permission to edit comments could do so?

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -130,7 +130,7 @@ export default {
     },
     methods: {
         canEditComment (comment) {
-            return (this.$hasPermission('can_edit_staff_comment') && !comment.can_have_journal) || this.$store.getters['user/uID'] === comment.author.id
+            return this.$hasPermission('can_edit_staff_comment') && (comment.author.id !== comment.journal_author)
```
Larspolo: You should also be able to edit your own comments right? This is not allowed rn

src/django/VLE/serializers.py
```diff
@@ -316,8 +316,8 @@ def get_author(self, comment):
     def get_last_edited_by(self, comment):
         return None if not comment.last_edited_by else comment.last_edited_by.full_name
 
-    def get_can_have_journal(self, comment):
-        return comment.author.has_permission('can_have_journal', comment.entry.node.journal.assignment)
+    def get_journal_author(self, comment):
+        return comment.entry.node.journal.user.pk
```
Larspolo: The author should already be known when gathering comments. Do we want to still have this then?

src/django/VLE/models.py
```diff
@@ -945,7 +945,9 @@ def can_edit(self, user):
         """
         Returns whether the given user is allowed to edit the comment:
             - Has to be the author or super_user
-            - Otheriwse has to have the permission 'can_edit_staff_comment' and be a comment of a staff member.
+            - Otheriwse has to have the permission 'can_edit_staff_comment' and edit a non staff comment.
```
Larspolo: and edit a non journal-author* comment.

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -125,6 +132,9 @@ export default {
         this.setComments()
     },
     methods: {
+        canEditComment (comment) {
+            return this.$hasPermission('can_edit_staff_comment') && (comment.author.id !== this.journal.student.id)
```
engelhamer: Editing own comments should also always be allowed, regardless of permission.

src/django/VLE/migrations/0017_can_edit_staff_comment.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='comment',
+            name='last_edited_by',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='last_edited_by', to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AddField(
+            model_name='role',
+            name='can_edit_staff_comment',
+            field=models.BooleanField(default=False),
```
Larspolo: A method needs to be added that sets it on `True` for teachers

src/vue/src/components/journal/CommentCard.vue
```diff
@@ -125,6 +132,10 @@ export default {
         this.setComments()
     },
     methods: {
+        canEditComment (comment) {
+            return (this.$store.getters['user/uID'] === comment.author.id ||
```
Larspolo: unnecessary outer brackets

src/django/VLE/migrations/0017_can_edit_staff_comment.py
mjvkeulen: will do

src/django/test/test_comment.py
```diff
@@ -60,25 +101,132 @@ def test_create(self):
...
+            status=200 if should_succeed else 403
+        )
+
+        comment_after_op = Comment.objects.get(pk=comment.pk)
+        assert comment_before_op is not comment_after_op, "Comments should not be the same object instance"
+
+        if should_succeed:
+            comment_resp = comment_resp['comment']
+            assert comment_resp['text'] == update_msg, 'Text should be updated'
+            assert comment_resp['creation_date'] == old_comment_resp['creation_date']
```
Larspolo: Missing error message

src/django/test/test_comment.py
```diff
@@ -60,25 +101,132 @@ def test_create(self):
...
+
+    def test_delete_as_student(self):
+        # Student should be allowed to delete his own comment
+        self.check_comment_delete(self.comment, self.student, True)
+        # Student should not be allowed to delete a TA's comment
+        self.check_comment_delete(self.TA_comment, self.student, False)
+        # Student should not be allowed to delete a Teachers comment
+        self.check_comment_delete(self.teacher_comment, self.student, False)
+
+    def check_comment_update(self, comment, user, should_succeed):
```
Larspolo: Maybe change should succeed to status (and default it to 200). This way, this function can also check for other types of errors

src/django/test/test_comment.py
```diff
@@ -60,25 +101,132 @@ def test_create(self):
...
+        else:
+            assert_comments_are_equal(comment_before_op, comment_after_op)
+
+    def check_comment_delete(self, comment, user, should_succeed):
+        comment_before_op = Comment.objects.get(pk=comment.pk)
+
+        api.delete(self, 'comments', params={'pk': comment.pk}, user=user, status=200 if should_succeed else 403)
+
+        if should_succeed:
+            assert not Comment.objects.filter(pk=comment.pk).exists()
```
Larspolo: Missing error message

src/django/test/test_comment.py
```diff
@@ -60,25 +101,132 @@ def test_create(self):
...
+            comment_before_op.save()
+            assert Comment.objects.filter(pk=comment.pk).exists()
+        else:
+            comment_after_op = Comment.objects.get(pk=comment.pk)
+            assert comment_before_op is not comment_after_op, "Comments should not be the same object instance"
```
Larspolo: Is this check necessary?

src/django/test/test_comment.py
```diff
@@ -60,25 +101,132 @@ def test_create(self):
...
+        TA_role = self.journal.assignment.courses.first().role_set.get(name='TA')
+        TA_role.can_edit_staff_comment = True
+        TA_role.save()
+        # TA with 'can_edit_staff_comment' should be allowed to edit his own comment
+        self.check_comment_update(self.TA_comment, self.TA, True)
+        # TA with 'can_edit_staff_comment' should not be allowed to edit a students comment
+        self.check_comment_update(self.comment, self.TA, False)
+        # TA with 'can_edit_staff_comment' should be allowed to edit a Teachers comment
+        self.check_comment_update(self.teacher_comment, self.TA, True)
+        self.journal.assignment.courses.first().role_set.filter(name='TA').update(can_edit_staff_comment=False)
```
Larspolo: Change this (and below) to `TA_role.can_edit_staff_comment = False; TA_role.save()`?

src/django/test/test_comment.py
mjvkeulen: behavior changes depending on success or not, which could, of course, be handled but let's not for this

PR #625, opened 2019-03-29
Added privacy notice to login and profile page

engelhamer:
## Description
Previously the privacy notice was only accessible when not logged in, on the guest page. Since most traffic doesn't originate from there it is good to have it visible elsewhere.


src/vue/src/components/assets/Footer.vue
```diff
@@ -0,0 +1,30 @@
...
+    <footer>
+        <hr/>
+        &copy; 2019 - eJournal.app -
+        <a href="/static/eJournal-privacy-notice.pdf" class="text-grey">
+            Privacy notice
+        </a>
+        <tooltip tip="We reserve the right to change this privacy policy as we may deem necessary from time to time or as may be required by law. Any changes will be immediately posted on the website and you are deemed to have accepted the terms of the privacy policy on your first use of the website following the alterations."/>
+    </footer>
+</template>
+<script>
```
mjvkeulen: spacing?

src/vue/src/components/assets/Footer.vue
```diff
@@ -0,0 +1,30 @@
...
+import tooltip from '@/components/assets/Tooltip.vue'
+import icon from 'vue-awesome/components/Icon'
+
+export default {
+    components: {
+        tooltip,
+        icon
+    }
+}
+
```
mjvkeulen: extra space?

PR #628, opened 2019-04-03
Catch non-existing journal errors for bonus points import

mjvkeulen:
## Description
Check if users are actual participants of the course and display an appropriate error.


src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
```diff
@@ -26,6 +26,12 @@
                     <b>&emsp;&emsp;&emsp;{{ lineNumber }})</b> {{ username }}<br/>
                 </span>
             </div>
+            <div v-if="errorLogs.non_participants">
+                <b>The following users are not participants of the assignment:</b><br/>
```
engelhamer: Maybe add a message that these users might have to open the assignment in the LMS, since journals are only created for LTI users once they visit the assignment (and set up the grade url) if the assignment existed before they joined the course.

mjvkeulen: Same thing goes for not existing users tho?

engelhamer: Hm that's true, but this scenario could also occur when a user has opened one assignment but not another (which is a quite likely scenario and therefore might need some further explanation).

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
```diff
@@ -18,8 +18,8 @@
         <div v-if="errorLogs">
             <b class="text-red">Errors in file:</b>
             <div v-if="errorLogs.non_participants || errorLogs.unknown_users" class="text-dark-grey mb-1">
-                <i>Note: it is likely that one or more of the users reported as unkown or non participant, still need
-                   to visit eJournal from Canvas.</i>
+                <i>Note: it is likely that one or more of the users reported as unkown or non participant still need
+                   to visit this assignment on the LMS (Canvas).</i>
```
mjvkeulen: This no longer explicitly states that the users need to visit eJournal from the learning environment, the clarification that it's concerning this specific assignment is an improvement however.

PR #631, opened 2019-04-05
Ansible baselink variable moved to inventory file

mjvkeulen:
## Description
Moved the ansible baselink variable to the inventory file so that it's linked with the specific server.


system_configuration_tools/inventory
```diff
@@ -1,5 +1,5 @@
 [staging]
-test.ejournal.app ansible_user=mel env=stage
+test.ejournal.app ansible_user=mel env=stage baselink=https://test.ejournal.app
 
 [production]
-uva.ejournal.app env=production
+uva.ejournal.app env=production baselink=https://uva.ejournal.app
```
engelhamer: There must be a more elegant solution than hardcoden a specific instance subdomain.

Larspolo: This is pretty elegant imo. Now we can have a different one per domain. Maybe [production] is now not the right word, but inside here we can have just a list of every host we want to deploy to.

mjvkeulen: I think the approach of setting host specific variables in the inventory file is correct. Might not seem obvious with our current setup of one server for front back and db however.

engelhamer: Stuff like that should be configurered on the machine, not in a public repository for the project itself.

PR #633, opened 2019-04-06
Always enable a maintenance page during deployment

mjvkeulen:
## Description
Always enable a maintenance page during deployment


src/vue/package-lock.json
```diff
@@ -1567,7 +1567,7 @@
       "dev": true
     },
     "bootstrap": {
-      "version": "4.3.1",
+      "version": "4.1.3",
```
Larspolo: Why is this changes?

system_configuration_tools/roles/web/tasks/setup_maintenance_page.yml
```diff
@@ -0,0 +1,15 @@
...
+    state: directory
+    owner: www-data
+    group: www-data
+    mode: 0755
+    recurse: yes
+
+- name: Create the maintenance page
+  template: src=maintenance_off.html
+            dest={{ maintenance_dir }}/maintenance_off.html
+            mode=0664
```
Larspolo: missing break at end of file

src/vue/package-lock.json
mjvkeulen: Good catch, it should not something messy with me switching branches I assume

PR #635, opened 2019-04-10
Restart celery on deploy

mjvkeulen:
## Description
Any modifications to celery related code executed by workers requires a reload before it takes effect.


system_configuration_tools/roles/celery/tasks/restart_celery.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+- name: Restart {{ celery_application_name }}_beat
+  supervisorctl:
+    name: "{{ celery_application_name }}_beat"
+    state: restarted
```
Larspolo:
```TASK [celery : Restart celery_beat] *********************************************
fatal: [test.ejournal.app]: FAILED! => changed=false 
  msg: |-
    celery_beat: ERROR (not running)
    celery_beat: ERROR (spawn error)
```
This task fails when its not running

Larspolo: image

mjvkeulen: This is desired behaviour. When **deploying** to a correctly **provisioned** server celery beat will be up and running. This is simply feedback something is wrong, which is correct.

mjvkeulen: (Can't restart something if it just crashes right)

Larspolo: Nothing else seems to fail, so what else could be going wrong? Cuz it doesnt log much else

mjvkeulen: Currently celery beat has timezones issues

PR #636, opened 2019-04-13
Updated favicon

engelhamer:
## Description
Updated the favicon. Comparison between new (left) and old (right):

image


src/vue/static/favicon/site.webmanifest
```diff
@@ -1,19 +1,55 @@
 {
-    "name": "eJournal",
-    "short_name": "eJournal",
+    "name": "",
```
mjvkeulen: What are these names used for that they are now blank?

engelhamer: This is used for e.g. 'installing' the website on a handheld device as a browser shortcut. By default the site name is used, so it's not really needed to specify, but I reckon it won't hurt to do so. Will add this and the taskbar color back in.

Issue #573, opened 2019-02-04
More intuitive adding of nodes in timeline


Larspolo:
**Describe the solution you'd like**
Adding nodes to format editor needs to be in a more intuitive way upon pressing the new node button.

Currently, a node is immediately added upon clicking the add node. We should have this work the same as adding a node to a journal by a student.

PR #645, opened 2019-04-16
Only show available dates in date pickers

### LINKED ISSUE #618: Availability range for datetime picker
Currently, a datetime picker can be used to select any date. Also ones that are not appropriate at all, such as a preset deadline that is after the assignment deadline. The user is confronted with an error message that the setting could not be saved that way, but it would be more elegant (and user-friendly) to disallow selection of invalid dates in the first place.

**User story**
As a user I only want to be presented with valid dates in date time pickers, according to the context.

**User interface**
Some sort of range in the interface (flatpickr) itself like so (with in this case all week-ends blocked):

Screenshot_20190314-233414__01

More info at https://flatpickr.js.org/examples/

engelhamer:
## Description
Dates can now only take on valid values. Rules that have to be followed are as follows. Hold tight:

- Course unlock date must be before course lock date
- Course lock date must be after course unlock date
- Assignment unlock date must be before any entry unlock, due or lock date within that assignment, and before assignment due date and lock date
- Assignment due date must be in between assignment unlock date and lock date, and before any entry due date within that assignment
- Assignment lock date must be after any entry unlock, due or lock date within that assignment, and after assignment due date and unlock date
- Entry unlock date must be before entry due date and entry lock date, before assignment due date and lock date, and after assignment unlock date
- Entry due date must be between entry unlock date and lock date, before assignment due date and lock date, and after assignment unlock date
- Entry lock date must be after entry unlock and due date, after assignment unlock date, and before assignment lock date

image

Following all these rules it is still possible to mess with some dates, but only when changing the preset type from entry to progress and back. This is because progress presets do not have any unlock and lock date to adhere to. This will, however, be no problem anymore after #573 (since the preset type will be immutable from that moment onwards).

## Summary of changes
- Fixes #618


engelhamer:
> I wasn't able to set an entry lock date before the assignment due date?
This has now been fixed.
> Assignments are still decoupled from their respective course start and end date, but with our system of shared assignments this is a feature right?
Indeed, limiting this would render existing assignments useless.

PR #647, opened 2019-04-17
Improved web request handling

### LINKED ISSUE #585: Excessive 'Token expired' toasts
**Describe the bug**
If a user was previously logged in but their token has expired, they will be redirected to a login screen when they try to visit a page. Upon arrival, two (sometimes more) toasts stating that their token has expired (or might be invalid) pop up.

**To Reproduce**
1. Log in
2. Wait for the token to expire
3. Visit a page
4. Observe an abundance of toasts

**Expected behavior**
Only one toast.

mjvkeulen:
## Description
Connection count handling is moved to Vuex by use of Axios interceptors.
Refactored token refresh handling:
Now all requests which fail due to an invalid access token are queued.
Only a single access token refresh request is performed when encountering such a failure.
 - On success, all queued requests are retried.
 - On failure, all queued requests are failed silently, and the user is redirected to login.

Please note that this PR continues from the Airbnb linter branch, which ideally should be approved first at some point.

## Fixes
- fixes #585


mjvkeulen:
> This is branched from the wrong branch

Well this is in the note, and intentional?

src/vue/src/api/auth.js
```diff
@@ -62,8 +62,6 @@ function toastError (error, connArgs) {
 /* Lowers the connection count and toast a success message
  * if a custom one is provided or responseSuccessToast is set. */
 function handleSuccess (resp, connArgs) {
-    setTimeout(() => { store.commit('connection/CLOSE_API_CALL') }, 300)
```
engelhamer: The timeouts were there to avoid a flashing spinner.

mjvkeulen: Fair point. However, timeouts on connection counts are not the right approach. Minimal show time on the spinner would be. The current approach is much more robust, as the connection count is always guaranteed to be correct.

engelhamer: How would you suggest to do this?

mjvkeulen: I've pushed a potential solution

PR #648, opened 2019-04-17
Added new release alert

### LINKED ISSUE #469: Change log tracking and update publishing
Keeping track of added features in a format which is understandable to our user base. Giving insight to what additions are made beyond the Github history. Preferably published to a location that is easy to access on our site, e.g. our landing page.

engelhamer:
## Description
Added an alert beneath the breadcrumb to show that a new version of eJournal has become available. This will be shown until a week after the release. If the user dismisses the alert they will not see it again until a new version is released (storing the version they don't want to see anymore in the database).

image

## Summary of changes
- Fixes #469


src/vue/src/views/Home.vue
```diff
@@ -65,13 +65,13 @@ export default {
         }
     },
     components: {
-        'content-columns': contentColumns,
```
mjvkeulen: :(

src/django/VLE/migrations/0020_preferences_hide_version_alert.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('VLE', '0019_unlock_due_lock'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='preferences',
+            name='hide_version_alert',
+            field=models.TextField(default='v0.0', max_length=10),
```
mjvkeulen: Is this done for comparison reasons? Making it nullable would be might accurate, don't mind either way however.

engelhamer: This does not carry any meaning. Ir can be changed to null indeed.

src/vue/src/components/assets/VersionAlert.vue
```diff
@@ -0,0 +1,41 @@
...
+        return {
+            version: CurrentRelease.version,
+            message: CurrentRelease.message,
+            date: CurrentRelease.date
+        }
+    },
+    computed: {
+        showVersionAlert () {
+            return this.message !== '' &&
+                this.$store.getters['preferences/hideVersionAlert'] !== this.version &&
```
mjvkeulen: I prefer reading the syntax made available by mapGetters, but there still only a couple vuex uses in this component.

src/vue/src/components/assets/BreadCrumb.vue
```diff
@@ -36,17 +36,20 @@
                 </span>
             </h1>
         </div>
+        <version-alert/>
```
Larspolo: I would prefer this being added above the rest as its not part of the actual content

PR #658, opened 2019-04-21
Fixed an issue where optional file upload fields would still be required

### LINKED ISSUE #652: Non-required file upload throws an error when left empty
**Describe the bug**
The back end always denies requests that contain `None` for a file upload field.

**To Reproduce**
1. Make a template with a non-required file upload field
2. Try to post it as a student
3. Observe that the server returns a bad request stating `One of your files was not correctly uploaded, please try gain.`

**Expected behavior**
Non-required fields to be non-required.

mjvkeulen:
Bugfix for non required journal file fields

- Fixes #652


src/django/VLE/views/entry.py
```diff
@@ -80,12 +80,12 @@ def create(self, request):
 
             if field.type in ['i', 'f', 'p']:  # Image, file or PDF
                 user_file = file_handling.get_temp_user_file(request.user, assignment, content['data'])
-                if user_file is None:
+                if user_file:
+                    file_handling.make_permanent_file_content(user_file, created_content, node)
+                elif user_file is None and field.required:
                     node.entry.delete()
                     return response.bad_request('One of your files was not correctly uploaded, please try gain.')
```
engelhamer: Again*

engelhamer:
Wasn't the problem the other way around? This also fixes #652 then.

src/django/VLE/views/entry.py
mjvkeulen: I don't think anyone objects to fixing old typos, the online editor is quite good for fixing such finds I am sure?

mjvkeulen:
> Wasn't the problem the other way around? This also fixes #652 then.

The other way round? This fixes both empty non required file field creation as well as updating them with a value (patch) later on.

src/django/VLE/views/entry.py
```diff
@@ -80,11 +80,11 @@ def create(self, request):
 
             if field.type in ['i', 'f', 'p']:  # Image, file or PDF
                 user_file = file_handling.get_temp_user_file(request.user, assignment, content['data'])
-                if user_file is None:
+                if user_file:
+                    file_handling.make_permanent_file_content(user_file, created_content, node)
+                elif user_file is None and field.required:
```
Larspolo: I would prefer reordering these if stmts, first `if user_file is None and field.required: return ...`. This way you first check if something is wrong, and if there isn't it will go through and do something else (`if user_file: make permanent`)

PR #662, opened 2019-04-27
Split preset creation and modification

### LINKED ISSUE #573: More intuitive adding of nodes in timeline

### LINKED ISSUE #611: Changing preset node type from entry to progress hides existing entry content
**Describe the bug**
... and instead shows the progress check itself. The entry is not lost, and points also still contribute to the total score.

**To Reproduce**
1. Create an assignment with a preset entry
2. Fill it in as a student
3. Grade it as a teacher
4. Change the preset type to progress
5. View the student's journal
6. Observe that the entry contents are not visible anymore

**Expected behavior**
The entry to remain visible, instead of the progress check.

**Additional context**
I think we should not allow changing the type of preset after creation (can be done in #573). That way this type of situation can never occur anymore. Changing the template for a preset entry does not cause any trouble.

engelhamer:
## Description
Adding a new preset node to an assignment can now be done in a dedicated card. This is where the type of node is chosen and details are provided. This allows users to revisit the node configuration before adding it to the format.

image

After having been added to the format the preset type is fixed. The other preset settings can however still be edited.

image

## Summary of changes
- Fixes #573
- Fixes #611


engelhamer:
Template previews are now hidden behind a button for a more organised view:

image

Also, the choice for the type of preset does not have a default value anymore. This makes it more explicit:

image

src/vue/src/components/format/FormatPresetNodeCard.vue
```diff
@@ -169,9 +146,17 @@
                 type="number"
                 class="theme-input"
                 placeholder="Amount of points"
-                min="0"
+                min="1"
```
Larspolo: 0.5 points are not allowed?

engelhamer: The main reason for this is that a minimum of 0 would make no sense: a goal that is instantly met by anyone who has a journal. Having another minimum than 1 would seem arbitrary, and I would not expect legitimate goals of less than one point anyways.

src/vue/src/components/journal/JournalStudent.vue
```diff
@@ -195,7 +195,7 @@ export default {
     watch: {
         currentNode () {
             if (this.currentNode !== -1
-                && this.currentNode !== this.nodes.length
+                && !this.currentNode >= this.nodes.length
```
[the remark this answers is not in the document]
engelhamer: I believe not, but this can probably be replaced by smaller than instead of greater than or equal.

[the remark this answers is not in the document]
engelhamer: Turn's out you were right indeed. Will replace it.

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -369,7 +369,7 @@ export default {
         },
         discardChanges () {
             if (this.currentNode !== -1
-                && this.currentNode !== this.nodes.length
+                && !this.currentNode >= this.nodes.length
```
mjvkeulen: Remaining

src/vue/src/components/journal/JournalStudent.vue
```diff
@@ -245,42 +245,36 @@ export default {
...
              */
-            if (this.currentNode !== -1
-                && this.currentNode !== this.nodes.length
-                && this.nodes[this.currentNode].type === 'd'
-                && this.nodes[this.currentNode].entry === null
-                && !this.isLocked()) {
-                if (this.$refs['entry-prev'].checkChanges()
-                    && !window.confirm('Progress will not be saved if you leave. Do you wish to continue?')) {
-                    return false
+            if (this.currentNode !== -1 && !this.currentNode >= this.nodes.length) {
```
mjvkeulen: Remaining

Issue #603, opened 2019-03-07
Upgrade to vue CLI 3.X / webpack 4.X


mjvkeulen:
**Describe the solution you'd like**
Upgrade to our frontend to use Vue CLI 3.X and/or Webpack 4.X to keep up with new releases and patching any vulnerabilities.

mjvkeulen:
Fixed in #665

PR #665, opened 2019-04-28
Upgrade to the latest version of webpack by making use of Vuecli 3

mjvkeulen:
## Description
Updated the front end to make use of the latest Webpack and Node environment. Allowing all dependencies to be updated, removing any found vulnerabilities.

### Note
One indirect vulnerability remains `tar`, via `node sass` via `node gyp`. However, this should be resolved soonish see:
https://github.com/nodejs/node-gyp/issues/1714


src/django/VLE/migrations/0021_unknown_profile_picture.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.1.7 on 2019-03-06 11:16
+
+from django.db import migrations, models
+
+
+def update_unknown_profile_pic_url(apps, schema_editor):
+    User = apps.get_model('VLE', 'User')
+    for user in User.objects.all():
+        if user.profile_picture == '/static/unknown-profile.png':
+            user.profile_picture = '/unknown-profile.png'
```
Larspolo: Maybe we should set this to `None` and add an alt to the places where we use the pf in the frontend

src/vue/package.json
```diff
@@ -1,33 +1,34 @@
 {
-  "name": "eJournal",
+  "name": "ejournal",
```
Larspolo: I prefer eJournal

src/vue/src/Header.vue
```diff
@@ -202,7 +202,8 @@ export default {
         &:hover
             > svg
                 fill: $theme-medium-grey !important
-        &.active
```
Larspolo: The active isnt orange anymore

src/vue/src/views/UserRoleConfiguration.vue
```diff
@@ -48,7 +48,8 @@
                             <!-- eslint-disable -->
                             <b-form-checkbox
                                 :class="{ 'input-disabled': essentialPermission(role, permission) }"
-                                v-model="roleConfig[getIndex(role)][permission]"/>
+                                v-model="roleConfig[getIndex(role)][permission]"
+                                inline/>
```
Larspolo: shouldnt this `/>` be on a newline?

src/vue/package.json
mjvkeulen: Should be non capitalized

src/vue/src/Header.vue
mjvkeulen: This was specifically addressed in 1850ff18f452457e59ef6d3e729e54ed23b7a3d6, are you sure?

src/vue/src/views/UserRoleConfiguration.vue
mjvkeulen: Yes, hm why is this not caught by the linter tho

src/vue/src/Header.vue
mjvkeulen: Ah the behaviour is different in the way that currently only the exact match is highlighted. Is this prefered or not? So should courses be highlighted when viewing a journal page, or should this only happen when visiting the view `Course`?

This change is as simple as moving from `router-link-exact-active` to `router-link-active` in the selector.

engelhamer: I'd be in favor of `router-link-active`, just like how it used to be.

mjvkeulen: Alright, I will update accordingly

src/vue/src/views/UserRoleConfiguration.vue
mjvkeulen: Ah silly this block is disabled, for syntax which is no longer disallowed even.

mjvkeulen:
fixes #603

PR #676, opened 2019-05-17
Added grade history

### LINKED ISSUE #459: Immutable grade history
A record of all grading steps should be kept.

**User story**
As a user with the privilege 'can_see_grade_history' it should be possible to see an immutable grade record. This can be used to solved grading disputes later down the line.

### LINKED ISSUE #602: Allow unpublishing of grades
Currently, grades cannot be unpublished. A user can untick the publish box and click save, but the following toast appears:

image-1

It would be more useful if it would be possible to unpublish grades. E.g. when you want to make some changes to some grades but not publish them right away.

engelhamer:
## Description
image

image

## Summary of changes
- Fixes #459
- Fixes #602


src/django/VLE/migrations/0022_grade_history.py
```diff
@@ -0,0 +1,75 @@
...
+    dependencies = [
+        ('VLE', '0021_unknown_profile_picture'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Grade',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('entry', models.ForeignKey(editable=False, on_delete=django.db.models.deletion.CASCADE, related_name='+', to='VLE.Entry')),
```
mjvkeulen: Why did you decide not to create a backwards relationship?

src/django/VLE/serializers.py
```diff
@@ -418,14 +424,16 @@ def get_content(self, entry):
         return ContentSerializer(entry.content_set.all(), many=True).data
 
     def get_editable(self, entry):
-        return entry.grade is None and not entry.is_locked()
+        return (entry.grade is None or entry.grade.grade is None) and not entry.is_locked()
```
mjvkeulen: Move this check to a function for the Entry model?

src/django/VLE/serializers.py
```diff
@@ -418,14 +424,16 @@ def get_content(self, entry):
...
 
     def get_grade(self, entry):
         # TODO: Add permission can_view_grade
         if 'user' not in self.context or not self.context['user']:
             return None
-        if entry.published or self.context['user'].has_permission('can_grade', entry.node.journal.assignment):
-            return entry.grade
+        grade = entry.grade
+        if grade is not None and (grade.published or
```
mjvkeulen: Same thing here, if grade would suffice as the options are None or a valid Grade instance?

src/django/VLE/tasks/grading.py
```diff
@@ -3,22 +3,25 @@
...
-    if published:
-        Comment.objects.filter(entry__node__journal__assignment=assignment) \
-                       .exclude(entry__grade=None).update(published=True)
-        for journal in assignment.journal_set.all():
-            if journal.sourcedid is not None and journal.grade_url is not None:
-                lti_grade.replace_result(journal)
+    entries = Entry.objects.filter(node__journal__assignment=assignment).exclude(grade=None)
+
+    for entry in entries:
+        factory.make_grade(entry, user, entry.grade.grade, True)
```
mjvkeulen: User instance is not JSON serializable so this will fail when the message is passed to a worker?

src/django/VLE/views/entry.py
```diff
@@ -117,6 +115,7 @@ def partial_update(self, request, *args, **kwargs):
         content_list, = utils.required_typed_params(request.data, (list, 'content'))
         entry_id, = utils.required_typed_params(kwargs, (int, 'pk'))
         entry = Entry.objects.get(pk=entry_id)
+        graded = entry.grade is not None and entry.grade.grade is not None
```
mjvkeulen: Someone can submit an entry before the deadline has passed, which can then be graded but is not published. The response you are not allowed to edit graded entries is a bit magical for this scenario. Or can this not occur to begin with?

src/django/VLE/views/entry.py
```diff
@@ -177,12 +176,13 @@ def destroy(self, request, *args, **kwargs):
         pk, = utils.required_typed_params(kwargs, (int, 'pk'))
 
         entry = Entry.objects.get(pk=pk)
+        graded = entry.grade is not None and entry.grade.grade is not None
```
mjvkeulen: Yes, I believe moving this to an entry function would be nice.

engelhamer:
> Good stuff man!
> 
> Some issues spotted, and one more question:
> If students have permission to view the grade history, where should they access this?

They can't, see: https://github.com/eJourn-al/eJournal/pull/676/files#diff-62f9524da5687c49df68ec84eea44938R414

This feature is intended for graders only.

src/django/VLE/views/entry.py
engelhamer: Though it is not expected, teachers can indeed decide to grade entries before they are due. We could prevent that by not allowing teachers to grade entries until their lock date has passed, but then what if no such date constraints are present? I think that we should leave this up to the teachers themselves (this is also how it used to be).

src/django/VLE/migrations/0022_grade_history.py
```diff
@@ -0,0 +1,75 @@
...
+    for entry in Entry.objects.all():
+        if entry.grade is not None:
+            Grade.objects.create(entry=entry, grade=entry.grade, published=entry.published, author=None)
+
+
+def link_entry_grades(apps, schema_editor):
+    Grade = apps.get_model('VLE', 'Grade')
+    Entry = apps.get_model('VLE', 'Entry')
+    for entry in Entry.objects.all():
+        entry.grade = Grade.objects.get(entry=entry).first()
```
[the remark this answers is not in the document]
engelhamer: Ah that's right, `DoesNotExist` should be caught and grade should subsequently be set to `None`.

src/django/VLE/migrations/0022_grade_history.py
engelhamer: There is a foreign key for the latest grade embedded in the entry itself.

src/django/VLE/tasks/grading.py
mjvkeulen: I think it's that or provide the pk to the function and get the corresponding instance.

engelhamer: That would work 👍

mjvkeulen:
> This feature is intended for graders only.

I can get behind this approach, but at the same time, I feel that as a student being able to view the grade history of ones own entries has value as well. E.g. seeing whether a disputed mark has been updated. 

What do you guys think?

src/django/VLE/migrations/0022_grade_history.py
```diff
@@ -0,0 +1,75 @@
...
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+def create_entry_grades(apps, schema_editor):
+    Grade = apps.get_model('VLE', 'Grade')
+    Entry = apps.get_model('VLE', 'Entry')
+    for entry in Entry.objects.all():
+        if entry.grade is not None:
```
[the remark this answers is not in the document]
Larspolo: A grade of `0` is different from `None`

src/django/VLE/migrations/0022_grade_history.py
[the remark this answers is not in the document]
Larspolo: also `get` should be `filter` I think

src/django/VLE/migrations/0022_grade_history.py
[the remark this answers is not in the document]
engelhamer: That is true, Lars, but the grade variable in an entry will now always point to a grade object, which in turn contains the actual grade as well as details on authorship and publishment.

src/django/test/test_gradepassback.py
```diff
@@ -163,21 +164,23 @@ def test_check_if_need_VLE_publish_journals_nothing_needed(self):
...
+                   user=self.teacher)
         lti_grade.change_entry_vle_coupling(self.journal, Entry.GRADING)
 
         entry = Entry.objects.get(pk=entry.pk)
         assert entry.vle_coupling == Entry.GRADING, 'Check if change_entry_vle_coupling works'
 
     def test_replace_result_no_url(self):
         """Hopefully doesnt crash."""
-        factory.Entry(node__journal=self.journal, published=True)
+        factory.Entry(node__journal=self.journal)
```
Larspolo: This test is now a different test. This now tests if the grade is not updated when its not published iso when there is no grade_url. Great test, but I think this is already tested somewhere else

src/django/test/test_statistics.py
```diff
@@ -21,11 +22,9 @@ def setUp(self):
...
         """Test the journal stats functions in the serializer."""
         entries = utils.get_journal_entries(self.journal)
-        for i in range(len(entries)):
-            if i > 0:
-                entries[i].grade = 1
-                entries[i].published = True
-                entries[i].save()
+        for entry in entries[1:]:
+            api.create(self, 'grades', params={'entry_id': entry.id, 'grade': 1, 'published': True},
+                       user=self.journal.assignment.courses.first().author)
```
Larspolo: With this call, this will also crash when the grades api is broken. This shouldnt be the case, this should only test whether the journal stats are correct or not. I would suggest changing it back to manually adding a grade from where its tested.

src/vue/src/components/entry/ProgressNode.vue
```diff
@@ -38,8 +38,9 @@ export default {
                 if (node.nID === this.currentNode.nID) { return true }
 
                 if (node.type === 'e' || node.type === 'd') {
-                    if (node.entry && node.entry.grade && node.entry.published && node.entry.grade !== '0') {
-                        tempProgress += parseFloat(node.entry.grade)
+                    if (node.entry && node.entry.grade && node.entry.grade.published
+                        && node.entry.grade.grade !== '0') {
```
Larspolo: why is 0 excluded? not that it matters, but still

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -293,7 +293,7 @@ export default {
             for (let i = 0; i < this.nodes.length; i++) {
                 if ('entry' in this.nodes[i] && this.nodes[i].entry) {
                     const entry = this.nodes[i].entry
-                    if (('grade' in entry && entry.grade === null) || ('published' in entry && !entry.published)) {
+                    if ('grade' in entry && (entry.grade.grade === null || !entry.grade.published)) {
                         if (i < min) { min = i }
```
Larspolo: Why is this if inside another if?

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -293,7 +293,7 @@ export default {
             for (let i = 0; i < this.nodes.length; i++) {
                 if ('entry' in this.nodes[i] && this.nodes[i].entry) {
                     const entry = this.nodes[i].entry
-                    if (('grade' in entry && entry.grade === null) || ('published' in entry && !entry.published)) {
+                    if ('grade' in entry && (entry.grade.grade === null || !entry.grade.published)) {
```
Larspolo: below (TimelineNodeCircle.vue:158) we use `if (entry && entry.grade)`, and here its `'grade' if entry`. This is a bit inconsistent. I would suggest using the first method as that checks every case where something can be null, and the latter only checks if entry is null

Larspolo:
image

When loading the entry that has no grades, it gives an error

Larspolo:
If grade and published state do not change, should it really insert another grade entry? Imo it should only update when there is an actual update. (Accidental) double clicks now insert multiple entries in the grade history.

image

Larspolo:
image

Table overflow on mobile devices like the iPhone X, 6/7/8 and Galaxy S9

src/django/VLE/migrations/0022_grade_history.py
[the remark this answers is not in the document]
engelhamer: Yes, then `first` will yield None if the queryset is empty.

src/django/VLE/migrations/0022_grade_history.py
```diff
@@ -0,0 +1,75 @@
...
+        ),
+        migrations.AddField(
+            model_name='role',
+            name='can_view_grade_history',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AlterField(
+            model_name='entry',
+            name='grade',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to='VLE.Grade'),
```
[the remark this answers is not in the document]
engelhamer: Since we do not want Django to automatically set up backwards relationships. Instead, we maintain only a single *grade* for each entry: the most recent one.

src/django/VLE/views/grade.py
```diff
@@ -0,0 +1,91 @@
...
+            forbidden -- when its not their own journal, or the user is not allowed to grade that journal
+        On success:
+            success -- with all historic grades corresponding to the entry
+        """
+        entry_id, = utils.required_typed_params(request.query_params, (int, "entry_id"))
+
+        entry = Entry.objects.get(pk=entry_id)
+        assignment = entry.node.journal.assignment
+
+        request.user.check_permission('can_view_grade_history', assignment)
```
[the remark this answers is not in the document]
engelhamer: Why would that be? This permission should only be available if users can view the journal itself and using only this check is valid for this API call in my opinion: this does not relate to the entry content itself.

src/django/VLE/views/grade.py
```diff
@@ -0,0 +1,91 @@
...
+            success -- with the assignment data
+        """
+        entry_id, grade, published = utils.required_typed_params(request.data, (int, 'entry_id'), (float, 'grade'),
+                                                                 (bool, 'published'))
+
+        entry = Entry.objects.get(pk=entry_id)
+        journal = entry.node.journal
+        assignment = journal.assignment
+
+        request.user.check_permission('can_grade', assignment)
```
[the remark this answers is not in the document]
engelhamer: I don't understand this scenario.

src/django/test/factory/grade.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class GradeFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = 'VLE.Grade'
+
+    entry = factory.SubFactory('test.factory.entry.EntryFactory')
+    grade = 1
+    published = True
+    date = faker.date()
```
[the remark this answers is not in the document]
engelhamer: We randomise things like course start/end dates as well. Why would this be any different?

src/django/test/test_entry.py
```diff
@@ -128,19 +128,20 @@ def test_update(self):
         self.journal.assignment.save()
 
         # Grade and publish an entry
-        api.update(self, 'entries/grade', params={'pk': entry['id'], 'grade': 5}, user=self.student, status=403)
-        api.update(self, 'entries/grade', params={'pk': entry['id'], 'grade': 5}, user=self.teacher)
+        api.create(self, 'grades', params={'entry_id': entry['id'], 'grade': 5, 'published': True}, user=self.student,
+                   status=403)
+        api.create(self, 'grades', params={'entry_id': entry['id'], 'grade': 5, 'published': True}, user=self.teacher)
```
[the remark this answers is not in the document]
engelhamer: I'd prefer not to, since grades are so strongly connected to entries. For example the 'grade' field in an entry should be updated for every new grade.

src/django/test/test_statistics.py
engelhamer: Due to the way in which the `grade` field in the Entry model now works it is necessary to perform exactly the steps that the API takes. It would be a hassle to do this here.

src/vue/src/components/entry/ProgressNode.vue
engelhamer: I guess just to skip the float parse and addition. Don't see much other reason.

src/django/test/factory/grade.py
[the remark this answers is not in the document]
Larspolo: Not in the test factory, that is the preset db (and should alson not be done imo)

src/django/VLE/views/grade.py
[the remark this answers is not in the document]
Larspolo: for a proper error response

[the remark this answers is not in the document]
Larspolo: also, when an assignment is unpublished, but a user is allowed to see the grade history, this returns true. However, I think that when something is unpublished, you should not be able to see the grade history

src/django/VLE/views/grade.py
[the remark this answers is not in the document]
Larspolo: see prev comment

src/django/test/test_statistics.py
Larspolo: that create fcuntion doesnt do a whole lot more than creating a new grade (besides from publishing comments, but that is not used here). I think using factory boy here would be a better solution

src/django/VLE/views/grade.py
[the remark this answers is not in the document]
engelhamer: Yes, but that cannot happen. If the assignment is unpublished there will not be any entries that can be graded or viewed the grade history of.

src/django/test/test_statistics.py
engelhamer: No, the create grade function now also updates the appropriate field in the entry itself. This is an essential step.

src/django/VLE/factory.py
```diff
@@ -431,7 +431,7 @@ def make_grade(entry, author, grade, published=False):
     """
     grade = Grade.objects.create(
         entry=entry,
-        author=author,
+        author=User.objects.filter(pk=author).first(),
```
Larspolo: why this change? its now inconsistent with the other `make_*` functions

src/django/VLE/models.py
```diff
@@ -816,6 +822,43 @@ def to_string(self, user=None):
...
+    )
+    author = models.ForeignKey(
+        'User',
+        null=True,
+        editable=False,
+        on_delete=models.SET_NULL
+    )
+
+    def save(self, *args, **kwargs):
+        self.date = timezone.now()
```
Larspolo: shouldn't this only be updated when its being created (with `if not self.pk:`)? Also, I would prefer it being called `creation_date` as that is also used elsewhere

sidenote: I noticed that creation_date and last_edited is most likely not implemented correctly on other models. Can't test this though as I am still on holidays (entry)

src/django/VLE/utils/grading.py
```diff
@@ -1,12 +1,15 @@
+from VLE import factory
 from VLE.models import Comment, Entry
 
 
-def publish_all_journal_grades(journal, published):
+def publish_all_journal_grades(journal, author):
     """publish all grades that are not None for a journal.
 
     - journal: the journal in question
-    - published: either True or False. If True show the grade to student.
```
Larspolo: now its missing author in docstring

src/django/VLE/views/entry.py
Larspolo: This can now be replaced by the is_graded method right?

```return not (self.grade is None or self.grade.grade is None)```

File: src/django/VLE/factory.py (hunk 19 above)
Comment (engelhamer) [reply src/django/VLE/factory.py]: Because User objects are not serializable.

File: src/django/VLE/models.py (hunk 20 above)
Comment (engelhamer) [reply src/django/VLE/models.py]: `creation_date` would fit in a scenario where the object can change, and consequently there should then be a `last_edited` date. Grades however are immutable and therefore carry only a single `date`. Explicitly mentioning creation would not suit here in my opinion.

File: src/django/VLE/views/entry.py (hunk 5 above)
Comment (engelhamer) [reply src/django/VLE/views/entry.py]: Yes! Thanks.

File: src/django/VLE/models.py (hunk 20 above)
Comment (mjvkeulen) [reply src/django/VLE/models.py]: Grades are indeed immutable and as such can only have a single date. However, date as a variable name is not very descriptive. The `date` represents the `creation_date` of the grade, as such using `creation_date` as a variable name would be more descriptive and more consistent in line with other naming conventions imo.

File: src/django/VLE/factory.py (hunk 19 above)
Comment (mjvkeulen) [reply src/django/VLE/factory.py]: Wrap author in an `isinstance` check? Not a big deal either way imo.

File: src/django/VLE/management/commands/preset_db.py [hunk 22]
```diff
@@ -71,7 +71,7 @@ def gen_users(self):
             },
             "Teacher": {
                 "username": "Teacher",
-                "full_name": "Xavier van Dommelen",
+                "full_name": "Engel Hamer",
```
mjvkeulen: Most important change ;)

src/vue/src/components/entry/EntryNonStudentPreview.vue
```diff
@@ -120,33 +169,45 @@ export default {
...
     watch: {
         entryNode () {
             this.completeContent = []
             this.setContent()
-            this.tempNode = this.entryNode
 
             if (this.entryNode.entry !== null) {
-                this.grade = this.entryNode.entry.grade
-                this.published = this.entryNode.entry.published
+                this.grade = {
```
Larspolo: why not change this to `this.grade = this.entryNode.entry.grade`

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -155,19 +155,19 @@ export default {
...
+            if (entry && entry.grade && entry.grade.published) {
                 return 'graded'
             } else if (!entry && this.lockDateHasPassed()) {
                 return 'failed'
             } else if (!entry && this.dueDateHasPassed()) {
                 return 'overdue'
             } else if (!entry && !this.dueDateHasPassed()) {
                 return 'empty'
-            } else if (!isGrader && entry && !entry.published) {
+            } else if (!isGrader && entry && entry.grade && !entry.grade.published) {
```
Larspolo: This doesnt work as student, as they are unable to see the grade before its published. If you want this hourglass effect, it should be changed to `            } else if (!isGrader && entry && !entry.grade && !entry.editable) {`. Same goes for other uses of `entry.grade.published`

src/django/VLE/migrations/0022_grade_history.py
```diff
@@ -0,0 +1,75 @@
...
+    for entry in Entry.objects.all():
+        if entry.grade:
+            Grade.objects.create(entry=entry, grade=entry.grade, published=entry.published, author=None)
+
+
+def link_entry_grades(apps, schema_editor):
+    Grade = apps.get_model('VLE', 'Grade')
+    Entry = apps.get_model('VLE', 'Entry')
+    for entry in Entry.objects.all():
+        entry.grade = Grade.objects.filter(entry=entry).first()
```
mjvkeulen: Does this need to be a filter / do we have floating grade objects?

src/django/VLE/serializers.py
```diff
@@ -261,10 +266,11 @@ def get_stats(self, assignment):
         # Grader stats
         if self.context['user'].has_permission('can_grade', assignment):
             stats['needs_marking'] = journal_set \
-                .filter(node__entry__isnull=False, node__entry__grade__isnull=True).count()
+                .filter(Q(node__entry__grade__grade=None) | Q(node__entry__grade=None),
```
mjvkeulen: I thought that because of the `related_name='+'` an `entry` has no backward relation to a `grade`. But this filter is clearly using this relation from entry to grade. What am I missing here?

engelhamer: You are correct in that `related_name='+'` breaks this backward relation. The relation used here concerns the `grade` attribute of an entry (which is manually updated upon each grade change, it represents the current grade of the entry). The latter could not exist in parallel with the backwards relation of the same name.

src/django/VLE/migrations/0022_grade_history.py
engelhamer: I don't think it is strictly required at this point, but from here onwards there can exist multiple grade objects related to a single entry.

src/vue/src/components/entry/EntryNonStudentPreview.vue
```diff
@@ -91,9 +98,51 @@
...
         />
+
+        <b-modal
+            id="gradeHistoryModal"
+            ref="gradeHistoryModal"
+            size="lg"
+            title="Grade history"
+            hideFooter
+        >
+            <b-card class="no-hover">
```
mjvkeulen: image

Possibly remove the outer card? To me it is already in the body of the modal and is simply adding duplicate padding.

Larspolo:
image
When grading an entry that did not have a grade before

src/django/VLE/migrations/0022_grade_history.py
mjvkeulen: Right so:

```
if entry.grade:
    entry.grade = Grade.objects.get(entry=entry)
    entry.save()
```

Would have the same result, I found the filter to be confusing although this would effectively have no different outcome.

src/vue/src/components/entry/EntryNonStudentPreview.vue
engelhamer: This is the style we practice in all modals. If we want to change that we should do it everywhere, but I vouch for keeping it this way.

PR #680, opened 2019-06-25
Deploy speedup

### LINKED ISSUE #641: Reduce deploy time by removing the replace step
The removal of the step "Set the correct API url in all javascript files" woud be a great time reduction when deploying.

This could possibly be done by reading the api url from a setting file before building the frontend.

**User story**
When deploying as developer, the "Set the correct API url in all javascript files" step takes way too long.

mjvkeulen:
## Description
- Use environment variables set from ansible to handle the front end build (e.g. the api url).
- Removed the slow find and replace step of the deploy for the api url.

Deploy split into 3 make commands:
- `run-ansible-deploy`: full deploy as we are used to.
- `run-ansible-deploy-front`: only deploy the Vue front end web structure.
`run-ansible-deploy-back`: only deploy the Django backend.

fixes #641


system_configuration_tools/roles/web/tasks/setup_frontend_file_structure.yml
```diff
@@ -4,13 +4,14 @@
...
     path: "{{ frontend_web_dir }}/"
-  tags: deploy
 
 - name: Recreate {{ frontend_web_dir }} directory
   file:
     state: directory
     path: "{{ frontend_web_dir }}/"
-  tags: deploy
+    owner: www-data
+    group: www-data
```
Larspolo: maybe this should also be extracted to a variable like gunicorn_user?

system_configuration_tools/roles/web/tasks/setup_frontend_file_structure.yml
```diff
@@ -47,34 +50,25 @@
...
-    replace: "https://api.{{ inventory_hostname }}/"
-  with_items: "{{ javascript_files.files }}"
+- name: Ensure that the frontend file permissions are set properly
+  file: path={{ frontend_web_dir }}
+        recurse=yes
+        owner=www-data
+        group=www-data
+        mode=0755
+  changed_when: false
   tags: deploy
```
Larspolo: should this also be removed?

PR #682, opened 2019-06-30
Added badge to identically named assignments

### LINKED ISSUE #674: You can have multiple assignments with the same name under 1 course
You can have multiple assignments with the same name under 1 course. This should not be possible as this makes it unclear which assignment is which 

**User interface**
Current interface:
image

Here you cannot distinguish between the different assignment

engelhamer:
## Description
Added a badge containing the due or lock date of an assignment to the assignment card if multiple assignments share an identical name in a course.

image

## Summary of changes
- Fixes #674


src/vue/src/components/assignment/AssignmentCard.vue
```diff
@@ -1,21 +1,49 @@
 <template>
     <b-card :class="$root.getBorderClass(assignment.id)">
         <slot class="float-right"/>
-        <h2 class="d-inline">
+        <h2 class="d-inline align-middle">
             {{ assignment.name }}
-        </h2><br/>
+        </h2>
+        <!-- <br/> -->
```
Larspolo: Can be removed

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <b-card :class="$root.getBorderClass(course.id)">
+    <b-card :class="$root.getBorderClass(deadline.id)">
```
Larspolo: I think the deadline should get the same colour as either the assignment or the course. Not the deadline itself

engelhamer: The `deadline` is an assignment here.

engelhamer: This todocard system could use a rewrite: stuff like this is unclear amd I reckon we should only show actual deadlines instead of just repeatong the assignments.

mjvkeulen:
Good stuff!

Couple questions:
- Do we feel there is a need to also display the time part?
- Is it possible to work with a more traditional `from` - `to` format? Be it by using a starting pointed inherited by the parent course or another solution?

engelhamer:
> Do we feel there is a need to also display the time part?

I think the full deadline would do, but if that's too long it can be shortened.

> Is it possible to work with a more traditional `from` - `to` format? Be it by using a starting pointed inherited by the parent course or another solution? And of course, do we feel this would even be a better fit than just showing due / lock dates?

It would be possible, but simply showing due and lock dates suffices for the purpose of distinguishing identically named assignments. The label doesn't appear for uniquely named assignments at the moment.

src/vue/src/components/assets/TodoCard.vue
mjvkeulen: Shall we move this to an issue and resolve the conversation?

PR #693, opened 2019-08-18
Fixed an issue where sandboxed iframes were not updated

### LINKED ISSUE #686: Comments sometimes show for the wrong entry
**Describe the bug**
When switching entries in the timeline comments are sometimes associated with the wrong entry. 

Top reproduce: create multiple entries with their own comments, switch between entries.

engelhamer:
## Description
DOM-related timing issues caused sandboxed iframes (with user-generated content) to not update properly. For example when navigating from one entry to another, where the content of the comments belonging to the previous entry sometimes remained.

## Summary of changes
- Fixes #686


Larspolo:
image
**Bug:**
1. login
2. place comment
3. make a draft for another comment
4. logout
5. log back in (could be as someone else)
6. draft appears
7. send draft (this clears the comment rich-text-edit box) 
8. logouot
9. log back in (could be as someone else)
10. see that the draft is back again.
**What should happen:**
After the draft is cleared by sending the comment, it should not re-appear when logging out and in

PR #696, opened 2019-08-31
Improved grading speed

### LINKED ISSUE #690: Speedup grading process
The grading process can be sped up. Some ideas:

- Require only one button click to both save and publish a grade
- Proceed to the next node or journal upon submitting a grade

engelhamer:
## Description
Added several measures to improve the overall speed of grading entries back to back. Where it was previously required to tick a box for publishing a grade, this is now done with a dropdown button that maintains state based on the user's preference (last selected option remains).

image

The same holds for comments. An option to publish both the currently entered grade as well as the comment itself in one click is also available here.

image

In the profile section users can now configure whether they want to automatically proceed to the next ungraded entry upon publishing a grade, or proceed to the next journal in case all entries have been graded.

## Summary of changes
- Fixes #690
- Fixes an issue where the profile page would show an alert about losing changes even when nothing was changed


src/django/VLE/models.py
```diff
@@ -270,10 +270,38 @@ class Preferences(models.Model):
...
+    PUBLISH_AND_PUBLISH_GRADE = 'g'
+    COMMENT_SEND_BUTTON_OPTIONS = (
+        (SAVE, 's'),
+        (PUBLISH, 'p'),
+        (PUBLISH_AND_PUBLISH_GRADE, 'g'),
+    )
+    comment_button_setting = models.TextField(
+        max_length=2,
+        choices=COMMENT_SEND_BUTTON_OPTIONS,
+        default=PUBLISH,
```
Larspolo: imo is save the best default option. I think its most common that a teacher will be adding a comment with the grade, not just a random extra comment

src/vue/src/components/profile/PasswordCard.vue
```diff
@@ -17,7 +17,7 @@
                 class="theme-input multi-form"
                 type="password"
                 placeholder="Current password"
-                autocomplete="current-password"
+                autocomplete="new-password"
```
Larspolo: why is this new-password?

src/vue/src/components/profile/GradingCard.vue
```diff
@@ -0,0 +1,64 @@
...
+            />
+            <h2 class="field-heading multi-form">
+                Automatically select the first ungraded entry in a journal
+            </h2><hr/>
+            <toggle-switch
+                :isActive="$store.getters['preferences/autoProceedNextJournal']"
+                class="float-right"
+                @parentActive="getAutoProceedNextJournal"
+            />
+            <h2 class="field-heading">
```
Larspolo: I dont like the styling of these h2.  When the titles are shortened, and a description is added, you get something like this:

image

src/vue/src/views/Profile.vue
```diff
@@ -2,10 +2,8 @@
     <content-single-column>
         <bread-crumb/>
         <profile-data ref="profileData"/>
-        <h4 class="mb-2 mt-4">
-            <span>Email notifications</span>
-        </h4>
         <notification-card/>
+        <grading-card v-if="$hasPermission('can_add_course')"/>
```
Larspolo: TA's should also be able to edit those settings. Why not use 'can_grade'?

src/vue/src/components/profile/PasswordCard.vue
engelhamer: I already commented on this when I opened this PR, but apparently it did not save: this prevenst e.g. chrome from autofilling the password box, triggering a changed event even though the user did not take any action.

src/vue/src/components/profile/GradingCard.vue
engelhamer: Looks better indeed.

src/vue/src/views/Profile.vue
engelhamer: `can_grade` is a course-level permission. We need something that applies for all staff here, like the `is_teacher` flag. Don't TA's also have this?

src/vue/src/components/profile/PasswordCard.vue
Larspolo: how about `autocomplete="off"`? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

src/vue/src/views/Profile.vue
Larspolo: TA's shouldnt have this, as they should not be allowed to create new courses, search for users outside current course etc

engelhamer: So what do you reckon would be a good global alternative?

src/vue/src/components/profile/PasswordCard.vue
engelhamer: I read some complaints about [chrome](https://stackoverflow.com/questions/30053167/autocomplete-off-vs-false) ignoring `off` as a value, Besides, the docs you linked state
> This may be used by the browser both to avoid accidentally filling in an existing password

so I went with `new-password` instead, as it seemed to fit fine.

src/vue/src/views/Profile.vue
Larspolo: check for all the courses if one of them is `can_grade`, if so, add this box? Or maybe jsut always display it? Or lastly (most preferred), add an `is_grader` permission in the backend that checks for any `can_grade` in all assignments?

src/vue/src/components/entry/CommentCard.vue
```diff
@@ -110,35 +110,54 @@
...
+                            class: '',
+                        },
+                    }"
+                    class="float-right mt-2"
+                    @click="addComment"
+                />
+                <b-button
+                    v-else
+                    class="float-right mt-2"
+                    @click="addComment('p')"
```
mjvkeulen: Reading `p`, `s`, or `g` is not very meaningful, possibly work with more succinctly named objects or variables as these occur in multiple files? These variables do occur in multiple files after all.

src/vue/src/components/assets/DropdownButton.vue
```diff
@@ -0,0 +1,148 @@
...
+            v-if="isOpen"
+            class="dropdown-button-options"
+            :class="{ up: up }"
+        >
+            <b-button
+                v-for="(details, option) in filteredOptions"
+                :key="option"
+                :class="details.class"
+                class="option"
+                @click="emitClick(option)"
```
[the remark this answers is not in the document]
engelhamer: I lije to keep this simple: this component is a button in the end, so a click would be the most natural event to emit. Note that this component is not used in a single dedicated context.

src/vue/src/components/entry/CommentCard.vue
```diff
@@ -110,35 +110,54 @@
...
-                        @click="addComment"
-                    >
-                        <icon name="paper-plane"/>
-                        Send
-                    </b-button>
-                </div>
+                <dropdown-button
+                    v-if="$hasPermission('can_grade') && !entryGradePublished"
+                    :up="true"
+                    :selectedOption="this.$store.getters['preferences/commentButtonSetting']"
```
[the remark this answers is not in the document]
engelhamer: Indeed.

src/vue/src/components/entry/CommentCard.vue
engelhamer: Following the same kind of naming as fields. Would longer names not be a problem?

src/vue/src/components/assets/DropdownButton.vue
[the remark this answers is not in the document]
mjvkeulen: From the perspective of being a general-purpose component sticking to the default `click` event makes sense.

src/vue/src/components/entry/CommentCard.vue
mjvkeulen: Or if more information needs to be conveyed, imported objects (as dict mapping or similar) from a helper file could clean this. You are right that field naming also is not setout for any further scaling.

PR #705, opened 2019-09-14
Fixes template order & CSP url

Larspolo:
## Description
Fixes template order
FIxes CSP url


src/django/VLE/serializers.py
```diff
@@ -359,7 +359,7 @@ class Meta:
         read_only_fields = ('id', )
 
     def get_templates(self, format):
-        return TemplateSerializer(format.template_set.filter(archived=False), many=True).data
+        return TemplateSerializer(format.template_set.filter(archived=False).order_by('name'), many=True).data
```
mjvkeulen: Upon template creation this order is not maintained front end

engelhamer: Wouldn't necessarily want templates to start shifting when changing their name anyways, so I am not really bothered by that. At last the order will remain when simply updating some fields, and will be logical upon every save.

Larspolo: Same. I would prefer how it is currently implemented

PR #706, opened 2019-09-15
Improved LTI connections

### LINKED ISSUE #685: Improve connections of multiple LMS courses to a single eJournal assignment
It is possible to connect a new LMS assignment to an existing assignment on eJournal, to allow journals that span multiple courses. Currently, grades are passed back to the first connection, which should be changed.

**User story**
Grades should be passed back to the last linked LMS assignment.

Tests for this should also be added.

Larspolo:
## Description
Improved the LTI connection. It now better allows for multiple LTI assignments to be connected to 1 eJournal assignment.

Fixes #685


src/django/VLE/lti_launch.py
```diff
@@ -132,11 +133,12 @@ def update_lti_course_if_exists(request, user, role):
     2. Add groups to the user
     """
     course_lti_id = request.get('custom_course_id', None)
-    lti_couple = Lti_ids.objects.filter(lti_id=course_lti_id, for_model=Lti_ids.COURSE)
-    if course_lti_id is None or not lti_couple.exists():
+    course = Course.objects.filter(active_lti_id=course_lti_id)
+    if course_lti_id is None or not course.exists():
```
Larspolo: course_lti_id check can be moves upwards, before the course filter

src/django/VLE/lti_launch.py
```diff
@@ -156,14 +158,16 @@ def update_lti_assignment_if_exists(request):
 
     If no course exists, return None
     If it does exist:
-    Update the published state
+        Update the published state
     """
     assign_id = request['custom_assignment_id']
```
Larspolo: why isnt the None check here as well?

src/django/VLE/migrations/0027_lti_id_handling.py
```diff
@@ -0,0 +1,64 @@
...
+
+def migrate_lti_ids(apps, schema_editor):
+    Course = apps.get_model('VLE', 'Course')
+    Assignment = apps.get_model('VLE', 'Assignment')
+    Lti_ids = apps.get_model('VLE', 'Lti_ids')
+
+    for course in Course.objects.all():
+        lti_ids = Lti_ids.objects.filter(for_model=Lti_ids.COURSE, course=course)
+        if lti_ids.exists():
+            course.active_lti_id = lti_ids.last()
```
Larspolo: should be changed to: `lti_ids.last().lti_id`. This is definitely never checked. Do this before deploy.

src/django/VLE/migrations/0027_lti_id_handling.py
```diff
@@ -0,0 +1,64 @@
...
+            course.save()
+
+    for assignment in Assignment.objects.all():
+        lti_ids = Lti_ids.objects.filter(for_model=Lti_ids.ASSIGNMENT, assignment=assignment)
+        if lti_ids.exists():
+            assignment.active_lti_id = lti_ids.last()
```
Larspolo: same here

src/django/VLE/migrations/0027_lti_id_handling.py
```diff
@@ -0,0 +1,64 @@
...
+            assignment.lti_id_set = [lti_id for lti_id in lti_ids]
```
Larspolo: and here. Maybe also filter out the duplicates if there are any

src/django/VLE/models.py
```diff
@@ -552,14 +574,61 @@ class Assignment(models.Model):
...
+                lti_id_set__contains=[self.active_lti_id]).exclude(pk=self.pk)
+            if other_assignments_with_lti_id_set.exists():
+                raise ValidationError(
+                    "A lti_id should be unique, and only part of a single assignment's lti_id_set.")
+
         return super(Assignment, self).save(*args, **kwargs)
 
+    def get_active_course(self):
+        """"Query for retrieving the course which matches the active lti id of the assignment."""
+        return self.courses.filter(assignment_lti_id_set__contains=[self.active_lti_id])
```
Larspolo: maybe add a `.first()`? as it should only return 1 right? `.first()` returns None is there are no items in the query result.

src/django/VLE/serializers.py
```diff
@@ -289,6 +298,23 @@ def get_courses(self, assignment):
             return None
         return CourseSerializer(assignment.courses, many=True).data
 
+    def active_lti_course(self, assignment):
+        if 'user' in self.context and self.context['user'] and \
+           self.context['user'].has_permission('can_edit_assignment', assignment):
+            c = assignment.get_active_course()
+            if c.exists():
+                c = c.first()
+                return {'cID': c.pk, 'name': c.name}
```
Larspolo: preferably use `'course_id'` we use it almost everywhere but LTI

Larspolo: Furthermore, a serializer (maybe a new `CourseDetailSerializer`)would be nicer. Not it's inconsistent with the one used below.

src/django/VLE/views/assignment.py
```diff
@@ -62,7 +61,7 @@ def list(self, request):
 
         data = serializer.data
         for i, assignment in enumerate(data):
-            data[i]['lti_couples'] = len(Lti_ids.objects.filter(assignment=assignment['id']))
+            data[i]['lti_couples'] = len(Assignment.objects.get(id=assignment['id']).lti_id_set)
```
Larspolo: why isnt this added in the serializer itself?

src/django/VLE/serializers.py
engelhamer: It's qute convenient ths way since this is only ever directly in the front-end and not updated as such.

PR #711, opened 2019-09-17
Removed a condition which could create an infinite loop under certain coniditions for the template field set

Larspolo:
## Description
Removed initite loop of template field set


mjvkeulen:
Added templates are shuffled after saving, aligning the ordering with the backend. Best would be to keep these two in similar, so the user experiences no disturbance?

PR #722, opened 2019-09-19
Added datetime field

### LINKED ISSUE #687: Datetime field in template
Possibility to chose a datetime as option from an template field type
Currently, only a date field is supported

Larspolo:
## Description
Added datetime field
Also improved entry validation and tests

## Summary of changes
- Fixes #687


src/django/VLE/factory.py
```diff
@@ -21,7 +21,7 @@ def make_instance(allow_standalone_registration=None):
     return instance
 
 
-def make_user(username, password, email, lti_id=None, profile_picture=None,
+def make_user(username, password, email, lti_id=None, profile_picture='/unknown-profile.png',
```
engelhamer: What does all the user stuff have to do with this?

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -28,7 +28,12 @@
                 v-model="completeContent[i].data"
                 class="theme-input full-width"
             />
-
+            <flat-pickr
+                v-if="field.type == 'dt'"
+                v-model="completeContent[i].data"
+                class="theme-input full-width"
+                :config="$root.flatPickrTimeConfig"
```
engelhamer: This config includes an alternative display format (e.g. `Mon 5 jan 2019`). Now this field is inconsistent with a regular date field, so that should get the same display style too.

src/django/VLE/factory.py
Larspolo: I got irritated by this function, it's not even really that useful anymore

src/django/VLE/validators.py
```diff
@@ -64,15 +55,15 @@ def validate_entry_content(data, field):
...
         url_validate = URLValidator(schemes=('http', 'https', 'ftp', 'ftps'))
         url_validate(data)
 
-    if field.type == SELECTION:
+    if field.type == Field.SELECTION:
         if data not in json.loads(field.options):
             raise ValidationError("Selected option is not in the given options")
 
-    if field.type == DATE:
+    if field.type == Field.DATE:
```
mjvkeulen: Would a similar check be desired for datetime?

src/django/VLE/utils/entry_utils.py
```diff
@@ -35,9 +37,23 @@ def get_node_index(journal, node, user):
...
 
-def check_required_fields(template, content):
+def check_fields(template, content_list):
+    """Check if the supplied content list is a valid for the given template"""
+    received_ids = []
+
+    # Check if all the content is valid
+    for content in content_list:
+        data, field_id = utils.required_params(content, 'data', 'id')
+        if data is not None and data != '':
```
mjvkeulen: Should we expand this to none, null etc (whatever js makes this)

Larspolo: not in here, that would mean that those texts cannot be posted. `"null"` != `null`, but if we check for it here, it would check for `"null"`

PR #723, opened 2019-09-20
Fixed a bug which prevents file fields from being edited

### LINKED ISSUE #691: Changing an uploaded image gives a not found error
**Describe the bug**
If an uploaded picture is changed after submitting an entry, clicking the title gives a 'not found' error toast.

**To Reproduce**
1. Post an entry with an image as a student
2. Click on 'edit'
3. Select another image on your computer for the image field
4. Save the change
5. Click on the name of your new image in the field

**Expected behavior**
The image to be succesfully saved.

**Screenshots**
image

**Additional context**
This issue was found on the development version. It is important to note that the updated image also does not appear in the media folder.

mjvkeulen:
## Description
Temporary files were deleted too early, resulting in a 404 when a user edits a file in an entry.

- Added some `user_file` tests
- Updated maintenance page deployment logic

fixes #691


src/django/VLE/views/user.py
```diff
@@ -376,6 +376,7 @@ def upload(self, request):
             except Content.DoesNotExist:
                 return response.bad_request('Content with id {:s} was not found.'.format(content_id))
 
+            request.user.check_can_edit(content.entry)
```
Larspolo: Should this check not also be added in `entry.py:partial_update()`

mjvkeulen: Entry patch has much more fine-grained response information. A restructure could be made where all edit logic is moved to `can_edit`, into a format which also handles the response information. 

I think this could be made quite clean overall, but not without significant refracture.

src/django/VLE/views/user.py
```diff
@@ -320,7 +320,12 @@ def download(self, request, pk):
         On success:
             success -- a zip file of all the userdata with all their files
         """
-        if int(pk) == 0:
+        try:
+            pk = int(pk)
```
Larspolo: how about using the `utils.required_typed_params` function? Then it would response with a propper invalid value error

src/django/test/test_file_handling.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # def test_req_factory(self):
+    #     req = self.reqFactory.post(path=self.upload_url, data={
+    #         'file': self.video,
+    #         'assignment_id': self.journal.assignment,
+    #         'content_id': 'null'
+    #     })
+    #     req.user = self.student
+    #     response = UserView.as_view({'post': 'upload'})(req)
+    #     self.assertEqual(response.status_code, 200)
```
Larspolo: can be removed if not used

src/django/VLE/permissions.py
```diff
@@ -118,6 +118,30 @@ def is_user_supervisor_of(supervisor, user):
     return False
 
 
+def can_edit(user, obj):
+    # QUESTION: Do we still want to enable superusers?
```
Larspolo: Not entries, we also exclude superusers from having journals

src/django/VLE/views/entry.py
```diff
@@ -159,7 +159,7 @@ def partial_update(self, request, *args, **kwargs):
             else:
                 factory.make_content(entry, data, field)
 
-            file_handling.remove_temp_user_files(request.user)
+        file_handling.remove_temp_user_files(request.user)
```
engelhamer: That's a sneaky error.

system_configuration_tools/roles/web/tasks/activate_maintenance_page.yml
```diff
@@ -0,0 +1,21 @@
+---
```
engelhamer: Good addition!

PR #725, opened 2019-09-22
Fixed an issue where the assignment editor's save button would appear without making changes

### LINKED ISSUE #719: Format editor save button appears when clicking around on entrynodes
**Describe the bug**
Format editor save button appears when clicking around on entrynodes

**To Reproduce**
1. Go to an assignment editor page
2. Click some entrynodes without editing them
4. See save button appear

**Expected behavior**
No save button when you are not editing

Larspolo:
## Description
Fixes format editor save button appearing
Improved handling of target points

## Summary of changes
- Fixes #719


src/django/VLE/migrations/0030_tager_to_float.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.7 on 2019-09-22 12:14
```
mjvkeulen: tager_to_float?

src/django/VLE/serializers.py
```diff
@@ -421,15 +421,15 @@ class Meta:
 
     def get_unlock_date(self, node):
         if node.type == Node.ENTRYDEADLINE and node.unlock_date is not None:
-            return node.unlock_date.strftime('%Y-%m-%d %H:%M')
+            return node.unlock_date.strftime('%Y-%m-%dT%H:%M')
```
engelhamer: How about we do this for all dates back end, so this data will always be uniform?

src/django/VLE/utils/generic_utils.py
```diff
@@ -229,7 +232,12 @@ def update_presets(assignment, presets, new_ids):
         preset_node.lock_date = lock_date if lock_date else None
 
         if preset_node.type == Node.PROGRESS:
-            preset_node.target = target
+            if target > 0 and target <= assignment.points_possible:
+                preset_node.target = target
+            else:
+                raise VLEBadRequest(
+                    'Progress goal needs to be between 0 and the maximum amount for the assignment: {}'
```
engelhamer: This is not enforced in the front-end. Would be nice to have a limit on the input field there.

src/django/VLE/serializers.py
Larspolo: This is for all dates using the serializer. Default is already does it this way. Did I miss it anywhere?

engelhamer: Yes, course start / end dates and assignment unlock, due and lock.

src/vue/src/components/format/FormatPresetNodeCard.vue
```diff
@@ -149,6 +149,7 @@
                 class="theme-input"
                 placeholder="Amount of points"
                 min="1"
+                :max="assignmentDetails.points_possible"
```
engelhamer: Maybe also add a check to the format validation to indicate that some preset nodes have a goal that is too high?

src/django/VLE/serializers.py
Larspolo: image

And course start/end is only a date, so thats in the form of `%Y-%m-%d`
image

mjvkeulen:
I like the split of the todo-square in needs marking and publish. I don't like seeing `0 | X` tho, would a single box with `X` not be prefered for this scenario?

PR #726, opened 2019-09-22
Improved deadline deck

### LINKED ISSUE #584: Points to achieve in Student ToDo card
As a student, I would like to see how many points I still need to achieve before the next deadline. Now it just displays when the upcoming deadline is, but an extra amount would be nice

Edit: this is only possible for deadlines with as goal an amount of points. For other deadlines, an indication on what kind  of deadline it is, would also be nice

**User Interface**
image

### LINKED ISSUE #659: Progress bar should be capped at 100%
**Describe the bug**
The progress of a student compared to others is displayed in a progress bar. If the average grade for an assignment is over 100% (so on average students have gathered more points than required) the progress bar will scale along with that. This is confusing since having a perfect score for the journal would then not fill up the bar entirely.

**Expected behavior**
A full progress bar should represent 100% of the possible points. Also, the percentage (in text) should maybe be capped to 100% upon achieving enough points.

**Screenshots**
image

### LINKED ISSUE #683: Improve to do section
The to do section now simply lists all assignments, whereas it was intended for upcoming deadlines that are not yet met. This results in an awkward situation in for example a course page, where there's essentially two identical lists of the course's assignments.

**User story**
As a teacher I want to see the amount of entries that need grading in a journal in a to do card. By clicking it I want to be taken directly to the first ungraded entry. If there is no work to grade no card should appear.

As a student I want to see the specific upcoming preset in a to do card. For example '10 points' (a progress goal) or 'Final reflection report' (a preset entry), rather than the name of the assignment itself. If all goals have been met and there are no preset entries available (note the unlock date of these: if that is in the future we should show another preset in the card) there should be no card.

### LINKED ISSUE #714: LTI assignment linked to two courses appears twice in to do section
**Describe the bug**
When an assignment a user participates in is linked to two LTI courses, it appears twice in the to do section. If the user does not participate in the other course, a 403 is displayed when the to do card is clicked on.

**To Reproduce**
1. Set up a course that is linked to two LTI assignments
2. Add another user (teacher for simplicity) to one of the courses
3. Log in as the added user
4. As the user click on the to do link that states the abbreviation of the course they do *not* participate in

**Expected behavior**
The assignment should appear only once, for the course the user participates in.

**Additional context**
Related to #683

Larspolo:
## Description
Improved deadline deck & progress bar issue
Student:
image

Teacher:
image
image

Progressbar fix:
image

## Summary of changes
- Fixes #714
- Fixes #683
- Fixes #584
- Fixes #659


src/django/test/test_assignment.py
```diff
@@ -38,6 +39,43 @@ def test_rest(self):
...
+        assert resp['journals'] is not None, 'Response should include teacher serializer'
+
+    def test_list(self):
+        course2 = factory.Course(author=self.teacher)
+        factory.Assignment(courses=[self.course])
+        factory.Assignment(courses=[self.course, course2])
+        assignment = factory.Assignment()
+
+        resp = api.get(self, 'assignments', params={'course_id': self.course.pk}, user=self.teacher)['assignments']
+        [print(a['id']) for a in resp]
```
mjvkeulen: Some leftovers

src/vue/src/components/assets/DeadlineDeck.vue
```diff
@@ -1,8 +1,11 @@
 <template>
-    <div>
+    <div v-if="computedDeadlines.length > 0">
+        <h3 slot="right-content-column">
+            To Do
+        </h3>
         <!-- TODO: This seems like an inappropriate permission check. Will have to be reconsidered in the rework. -->
```
engelhamer: We can now use a similar technique as we do for checking if the grading settings should be displayed in the profile page here.

src/vue/src/views/Assignment.vue
```diff
@@ -77,7 +77,8 @@
                     @click="publishGradesAssignment"
                 >
                     <icon name="upload"/>
-                    Publish all grades
+                    {{ assignment.journals.length === filteredJournals.length ?
+                        "Publish all grades" : "Publish grades of filtered journals" }}
```
engelhamer: This text is too long and screws up the alignment of the buttons when an LTI assignment has multiple links.

Larspolo: Sooo: "Publish grades"?

mjvkeulen: How many chars can we fit? I like the distinction between filtered and not. Maybe Publish all grades and Publish filtered grades?

src/vue/src/components/assets/TodoSquare.vue
```diff
@@ -1,21 +1,59 @@
...
+            {{ num2 ? Math.round(num2 * 100) / 100 : 0 }}
+        </b-badge>
+    </span>
 </template>
 
 <script>
 export default {
-    props: ['num'],
+    props: {
+        num: {
```
engelhamer: Not very descriptive naming.

src/vue/src/components/assignment/StudentCard.vue
```diff
@@ -77,13 +82,31 @@ export default {
...
+            } else if (this.markingNeeded > 1) {
+                info.push(`${this.markingNeeded} entries need marking`)
+            }
+            if (this.unpublished === 1) {
+                info.push('an entry needs to be published')
+            } else if (this.unpublished > 1) {
+                info.push(`${this.unpublished} grades need to be published`)
+            }
+
+            return info.join(' and ')
```
engelhamer: I don't like how the sentence starts lowercase. Maybe capitalise it here too?

PR #733, opened 2019-09-28
Moved actions for assignment to sidebar

Larspolo:
## Description

Moved action buttons to the sidebar 
image

Fixed bug where manage LTI popup selected wrong active LTI course


mjvkeulen:
I feel we're not consistent in our action placement:

- Create and edit actions are positioned in the main deck (course, assignment, group, permission etc).
- The format editor has many actions which are also displayed in the main deck e.g. (copy other assignment, delete assignment). 
- Profile page logic e.g. download data.

Because of this we should first settle on an overall strategy to improve our overall consistency of action placements, before we make these changes.

Larspolo:
Need to be revisited later

PR #736, opened 2019-09-28
Added fancier loading spinner

### LINKED ISSUE #712: Better looking spinner
The current spinner when loading (journals, from LTI to eJournal etc.) is not that good looking. It would be nice if users would be greeted with a nicer loading icon

engelhamer:
## Description
Added a loading spinner for tasks that can usually take a while.

load-spinner

## Summary of changes
- Fixes #712


src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -302,7 +319,12 @@ export default {
                 })
                     .then(() => {
                         journalAPI.getNodes(this.jID)
-                            .then((nodes) => { this.nodes = nodes })
+                            .then((nodes) => {
+                                this.nodes = nodes
+                                setTimeout(() => {
```
mjvkeulen: It would make sense to move the "minimum display time" logic to the spinner component itself. I am not sure om the cleanest way to go about this, ideally the parent component needs no additional state to accomplish this. Ideas?

src/vue/src/components/assets/LoadSpinner.vue
```diff
@@ -2,26 +2,46 @@
     <b-card
         class="large-load-spinner no-hover"
     >
-        <icon
-            name="circle-o-notch"
-            spin
-            scale="4"
-        />
+        <img src="/ejournal-book.svg"/>
+        <img src="/ejournal-book.svg"/>
```
mjvkeulen: Possible duplicate?

src/vue/src/views/Assignment.vue
```diff
@@ -247,12 +250,13 @@ import { mapGetters, mapMutations } from 'vuex'
 export default {
     name: 'Assignment',
     components: {
-        contentColumns,
```
mjvkeulen: So many additional lines flagged in this PR. For a single addition, or a wrap in a transition. Can we influence this some how?

src/vue/src/components/assets/LoadSpinner.vue
engelhamer: Nope, there's two books alternating eqch other in the animation (one has not yet left the scene when the other appears).

src/vue/src/views/Assignment.vue
engelhamer: Will try that with a dedicated loading (wrapper) component as discussed.

src/vue/src/views/Assignment.vue
```diff
@@ -176,48 +176,43 @@
...
+                    <student-card
+                        :listView="true"
+                        :student="journal.student"
+                        :stats="journal.stats"
+                    />
+                </b-link>
+            </div>
+            <main-card
+                v-if="assignmentJournals.length === 0"
+                :line1="'No participants with a journal'"
```
Larspolo: this can be just `line1="No participants with a journal"` right? same as below

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -192,6 +196,7 @@ export default {
             assignmentJournals: [],
             assignment: {},
             journal: null,
+            loadingNodes: true,
```
mjvkeulen: `!nodes` would work as well?

Larspolo: For this case yes, as you always have a start node. However, it wont work always, as for example with the assignments, it could also return an empty list of assignments

mjvkeulen: In javascript, an empty list is a non-truthy value I think

engelhamer: Yes, so that would mean the spinner stays indefinitely if there are no assignments in a course.

mjvkeulen: Fair enough, another reason to always initialize with `null` or equiv

Larspolo: image
Something like this works. I say we apply this as from then on we can differentiate between unloaded and loaded without the use on an extra variable

src/vue/src/views/AssignmentsOverview.vue
```diff
@@ -1,85 +1,87 @@
...
+                <b-button
+                    v-if="order"
+                    class="button multi-form"
+                    @click.stop
+                    @click="setOrder(!order)"
+                >
+                    <icon name="long-arrow-up"/>
+                    Descending
+                </b-button>
+            </div>
```
Larspolo: Should these input options already be loaded or not? Because now it is already loaded in Assignment.vue, but not in AssignmentOverview.vue. We should at least keep it consistent. Personally I would suggest already having the filter inputs in place

engelhamer: Fine by me :+1:

src/vue/src/components/journal/JournalNonStudent.vue
engelhamer: True, but that would eliminate the possibility to always check for e.g. `.length`.

src/vue/src/views/Assignment.vue
```diff
@@ -204,7 +204,7 @@
             </div>
             <main-card
                 v-if="assignmentJournals.length === 0"
-                :line1="'No participants with a journal'"
+                line1="'No participants with a journal'"
```
mjvkeulen: Double quotes no longer needed

PR #737, opened 2019-09-29
Improved LTI group names

### LINKED ISSUE #730: More sensible LTI section names
While we retrieve some LTI section names via a separate API for our current instance, some section names cannot be found. This includes the Canvas 'staff' section. The default section name on eJournal is the Canvas ID, so it often happens that a course contains a group with an odd name like `99834`. This is confusing for users that see this name appear in group filters. Groups should thus get more sensible names by default.

Larspolo:
## Description
Improved LTI group names

## Summary of changes
- Fixes #730


src/django/VLE/migrations/0032_sensible_lti_group_names.py
```diff
@@ -0,0 +1,28 @@
+# Set all latest (highest key) lti ids as the active lti id
```
mjvkeulen: Bit off topic I would say

PR #749, opened 2019-10-04
Fixed a permission issue which would sometimes prevent displaying a correct error page

### LINKED ISSUE #724: front-end permission system crashes when going to unauthorized assignment by changing the URL
**Describe the bug**
Going to an assignment a user does not have access to, results in JS crashing

**To Reproduce**
0. Login as Teacher
1. Create a course and assignment without any students
2. Login as a Student
3. go to that assignment by changing the URL (http://localhost:8080/Home/Course/1/Assignment/3 for example)
4. See error in console log

**Expected behavior**
Clear error page that you do not have access

**Screenshots**
image

**Additional context**
It could be that when loggin in the student didnt have access to the assignment, but while still in the assignment, it obtained that access. it would be nice that the student did not have to log out&in before it can access that newly obtained assignment

mjvkeulen:
## Description
Added thorough check for front end permission access, any unknown permissions are now simply false (instead of an error).

image

The question is whether this approach will hide relevant errors from our logging. Returning false is equal to silently failing the check as far as our logging goes front end, apart from side effects such as 403 responses. Thoughts?

- Fixes #724


src/vue/src/store/modules/permissions.js
```diff
@@ -1,6 +1,16 @@
...
 import router from '@/router'
 
+function getPermission (permissions, levelAndID, permission) {
+    if (levelAndID in permissions && permission in permissions[levelAndID]) {
+        return permissions[levelAndID][permission]
+    }
+
+    /* Unknown permission combination.
+     * QUESTION: Is throwing a (custom) error not still desired so sentry will register the event? */
+    return false
```
Larspolo: imo this shouldn't return false. It should try and update the locally stored permissions with an api request. After that request, and it still does not have the permission, it can return false

mjvkeulen: Whereas this certainly can be done, ask yourself under what conditions this would actually help?

Larspolo: This would help when a teacher adds a student to a new assignment while the student is still logged in. I think the new assignment will show up for the student, however, when clicking on that assignment, this function will return `false` as it doesn't have those new permissions set yet.

With my suggestion, it would also need to extend to work for superusers, as they do have all the permissions (except `can_have_journal`, so it also needs to pass this.

engelhamer: I agree with Lars' comment, since this would even cover the scenario's where a user is added to a course while they are logged in.

Regarding the question: I don't reckon this is of much interest here, since the error only occurs when manually navigating to a page you shouldn't be on?

Larspolo: > Is throwing a (custom) error not still desired so sentry will register the event?

I wouldnt say so, this should not really be considered anything more than a `403`

mjvkeulen: Updated, let me know

src/vue/src/store/modules/permissions.js
```diff
@@ -1,19 +1,46 @@
...
+
+    return (!store.getters['permissions/lastPermissionUpdate']
+            || ((date - store.getters['permissions/lastPermissionUpdate']) > fiveMins))
+}
+
+/* Attempt to update the permissions, throw error if this cannot be done and is not underway. */
+function handleError (error = null) {
+    if (canRefreshPermissions()) {
+        store.dispatch('permissions/updatePermissions')
+    } else if (error && !store.getters['permissions/permissioUpdateInFlight']) {
```
Larspolo: permissionUpdateInFlight*

src/vue/src/store/modules/permissions.js
```diff
@@ -1,19 +1,46 @@
...
+
     return false
 }
 
 const getters = {
+    lastPermissionUpdate: state => state.lastPermissionUpdate,
+    permissioUpdateInFlight: state => state.permissioUpdateInFlight,
     // eslint-disable-next-line
-    hasPermission: (state, getters, rootState, rootGetters) => (permission, givenKeyLevel = null, id = null) => {
+    hasPermission: (state, getters, rootState, rootGetters ) => (permission, givenKeyLevel = null, id = null) => {
```
Larspolo: why add that space after `rootGetters`?

src/vue/src/store/modules/permissions.js
```diff
@@ -30,11 +57,11 @@ const getters = {
...
         if (givenKeyLevel && id) {
             if (permissions[givenKeyLevel + id] === undefined) {
-                throw Error(`Permission type input error, no match for the given key: ${givenKeyLevel}${id}.`)
+                handleError(Error(`Permission type input error, no match for the given key: ${givenKeyLevel}${id}.`))
             }
-            if (permissions[givenKeyLevel + id][permission] === undefined) {
-                throw Error(
-                    `Permission ${permission} is not a valid permission for the context: ${givenKeyLevel}${id}.`)
+            if (permissions[givenKeyLevel + id] && permissions[givenKeyLevel + id][permission] === undefined) {
+                handleError(Error(
```
Larspolo: I think this function should also return what `handleError` returns

src/vue/src/store/modules/permissions.js
```diff
@@ -1,19 +1,46 @@
...
+/* Attempt to update the permissions, throw error if this cannot be done and is not underway. */
+function handleError (error = null) {
+    if (canRefreshPermissions()) {
+        store.dispatch('permissions/updatePermissions')
+    } else if (error && !store.getters['permissions/permissioUpdateInFlight']) {
+        throw error
+    }
+}
 
 function getPermission (permissions, levelAndID, permission) {
```
Larspolo: This still needs an `if (issuperuser && not permission is can_have_journal) return true`

src/vue/src/store/modules/permissions.js
```diff
@@ -1,19 +1,46 @@
+import Vue from 'vue'
 import * as types from '../constants/permission-types.js'
+import * as mutationTypes from '../constants/mutation-types.js'
 import router from '@/router'
+import store from '@/store'
+import connection from '@/api/connection.js'
+
+/* If no update has been performed in the last five minutes, update permissions */
```
Larspolo: Why 5 minutes? If a student is not connected to a course, then asks the teacher about it & the teacher adds the student, then the student needs to wait 5 minutes, or relog. Why not just every permission error?

src/vue/src/store/modules/permissions.js
mjvkeulen: Which returns nothing? This is a getter, and as such cannot work with asynchronous results directly.

src/vue/src/store/modules/permissions.js
mjvkeulen: Clicking around would spam the server, somewhat mitigated by the in-flight check already. I will reduce this to 30 seconds and let's leave it at that?

src/vue/src/store/modules/permissions.js
mjvkeulen: We currently don't even serialize the `is_superuser` flag, superusers are given all permissions backend.

Larspolo: Yeah, that's the issue then right. Because now superusers still cannot do anything because this will just forward them to the forbidden page

src/vue/src/store/modules/permissions.js
Larspolo: That will do

PR #752, opened 2019-10-06
Support for Canvas staff student view mode

### LINKED ISSUE #718: Student preview
A method for a teacher to view an assignment as if they were a student. This would give a clearer view of how the assignment is set-up, and check if it's correct.

**User story**
Login as teacher,
go to an assignment edit page,
press a button that says "student view"
be able to view the assignment as a student

mjvkeulen:
- Added support for staff-student view mode
- Expansion of LTI tests

fixes #718


mjvkeulen:
The student username of user view test students can get rather lengthy:

image

@engelhamer any chance you feel like looking at this?

src/django/VLE/migrations/0032_student_view.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='email',
+            field=models.EmailField(blank=True, max_length=254, null=True, unique=True),
+        ),
+        migrations.AddField(
+            model_name='user',
+            name='student_view_user',
```
engelhamer: I'd prefer `is_test_student`, consistent with the superuser and teacher style.

src/django/VLE/views/user.py
```diff
@@ -105,6 +105,7 @@ def create(self, request):
 
         lti_id, user_image, full_name, email, is_teacher = get_lti_params(
             request, 'user_id', 'custom_user_image', 'custom_user_full_name', 'custom_user_email')
+        is_test_student = bool(lti_id) and not bool(email)
```
Larspolo: `lti_id and email is None` would be clearer imo

src/django/VLE/views/user.py
```diff
@@ -129,14 +130,10 @@ def create(self, request):
...
-        user = User(username=username, email=email, lti_id=lti_id, full_name=full_name,
-                    is_teacher=is_teacher, verified_email=bool(lti_id),
-                    profile_picture=user_image if user_image else '/unknown-profile.png')
-        user.set_password(password)
-
-        user.full_clean()
-        user.save()
+        user = factory.make_user(username=username, email=email, lti_id=lti_id, full_name=full_name,
+                                 is_teacher=is_teacher, verified_email=bool(lti_id) and bool(email), password=password,
+                                 profile_picture=user_image if user_image else '/unknown-profile.png',
```
Larspolo: In the `factory.make_user` itself, it already defaults to `unknown-profile.png`

src/django/test/factory/params.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class JWTParamsFactory(factory.Factory):
+    class Meta:
+        model = dict
+
+    user_id = factory.Sequence(lambda x: 'LMS_user_id{}'.format(x))
+    custom_user_full_name = 'full name of LMS user'
+    custom_user_email = factory.Sequence(lambda x: 'valid_LMS_email{}@address.com'.format(x))
+    custom_user_image = 'https://LMS_user_profile_image_link.com'
```
Larspolo: Maybe use a `.png` extension for images?

src/django/test/test_permissions.py
```diff
@@ -16,7 +16,7 @@
 
 class PermissionTests(TestCase):
     def setUp(self):
-        self.user = factory.make_user('Username', 'Password', email='some@email.address', full_name='Test User')
+        self.user = factory.make_user('Username', 'Pass123!', email='some@email.address', full_name='Test User')
```
Larspolo: This should use the variable set in `test/factory/user` `DEFAULT_PASSWORD`, same as below

src/django/VLE/views/user.py
mjvkeulen: image

src/django/VLE/views/user.py
mjvkeulen: If zero can be a legitimate id, `bool(lti_id)` is required. Email can be `""` and `None`, but the bool is indeed not required.

mjvkeulen: image

I don't mind the bool tbh

engelhamer:
Too long usernames are now clipped:

image

image

src/django/test/factory/user.py
```diff
@@ -33,6 +33,10 @@ class TeacherFactory(UserFactory):
     is_teacher = True
 
 
+class LtiTeacherFactory(TeacherFactory):
+    lti_id = factory.Sequence(lambda x: "id{}".format(x))
```
Larspolo: Student and teacher should get different `lti_id`'s. Now if you create 1 `LtiStudent` and 1 `LtiTeacher` they will have the same `lti_id`

mjvkeulen: If so that sucks tbh, I would expect factories which inherit to share `sequence`. I had a similar problem with the params where I needed to split classes but share the same `sequence`.

mjvkeulen: As expected factory boy is not stupid

image

src/django/test/test_lti_launch.py
```diff
@@ -190,6 +190,26 @@ def test_lti_launch_key_error(self):
             delete_field='custom_username',
         )
 
+    def test_lti_launch_create_test_user(self):
+        next_lti_id = User.objects.all().order_by("-lti_id")[0]
```
Larspolo: `.first()` also works. This returns `None` when it crashes iso throws an index error

mjvkeulen:
> Account creation should be skipped when logging in as a test student.
> 
> As a teacher of a course, it would also be nice if you could delete (all) test students connected to that course

There can only be one test user per course, and this user can be removed along the normal path like all other users.

src/django/VLE/factory.py
```diff
@@ -44,6 +43,7 @@ def make_user(username, password, email=None, lti_id=None, profile_picture='/unk
     if is_test_student:
         user.set_unusable_password()
     else:
+        validators.validate_password(password)
```
Larspolo: User should only be created if the password is valid. Now it creates the user, and only after that is crashes.

src/django/test/utils/lti.py
```diff
@@ -50,3 +51,8 @@ def gen_jwt_params(params={}, user=None):
         req['oauth_signature'] = get_signature(req)
 
     return {'jwt_params': lti_view.encode_lti_params(req)}
+
+
+def get_new_lti_id():
+    user = User.objects.exclude(lti_id=None).latest('lti_id')
+    return '1' if not user.lti_id else '{}{}'.format(user.lti_id, 1)
```
Larspolo: `return '1' if not (user or user.lti_id) else '{}{}'.format(user.lti_id, 1)`

src/django/VLE/models.py
```diff
@@ -242,6 +243,8 @@ def to_string(self, user=None):
     def save(self, *args, **kwargs):
         if not self.email and not self.is_test_student:
             raise ValidationError('A legitimate user requires an email adress.')
+        if self.is_test_student and settings.LTI_TEST_STUDENT_FULL_NAME not in self.full_name:
+            raise ValidationError('Test user\'s full name deviates.')
```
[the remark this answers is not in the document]
mjvkeulen: I forced it onto the save for consistency, but this can be removed to allow a test user to change his name without consequences.

PR #757, opened 2019-10-12
Improved support accessibility

### LINKED ISSUE #743: Feedback modal is behind header
**Describe the bug**
Feedback modal is behind header (on Firefox)

### LINKED ISSUE #744: Make feedback option more accessible
Users have trouble finding the feedback button when they need it. It should be clearer and the hassle of giving feedback should be minimized.

**User interface**
Probably a larger chat icon and move to the bottom right.

engelhamer:
## Description
Moved the support button to the bottom right corner for increased visibility. Also renamed from 'feedback' to 'support' for the users, to make it more obvious that they can ask for help or submit bugs here as well.

image

The message that is sent to the user is now also sent to the support inbox, so a direct reply to that is possible.

## Summary of changes
- Fixes #743 
- Fixes #744


mjvkeulen:
I feel the chat icon implies chat rather strongly, would there be a more suitable alternative?
image

Some ideas:
image
image

I believe `from the options below` is superfluous, and very much implied from the button context.

Another point is the other button/category. I believe this will only be used for feedback (or do you guys foresee some other use?), and as such, it makes sense to make this explicit by naming it feedback.

On the point of bug reports, I believe it could be valuable to link the feedback to the most recent sentry event as we do on our error page.

engelhamer:
Icon can be changed and other updated. However, I feel the bug should not be linked to the sentry event per se: bugs that are reported this way have not caused a front end error and might just be behaviour the user doesn't expect, but also does not cause an error.

engelhamer:
@mjvkeulen what icon from https://fontawesome.com/icons?d=gallery would fit
your description?

mjvkeulen:
> @mjvkeulen what icon from https://fontawesome.com/icons?d=gallery would fit your description?

I could not find an appropriate icon there, how about this one? https://www.flaticon.com/free-icon/support_102365

engelhamer:
> I could not find an appropriate icon there, how about this one? https://www.flaticon.com/free-icon/support_102365

That's totally out of style with the rest of the (vue-/font awesome) icons 😬

mjvkeulen:
Why do you feel it is such a mismatch style wise?

image
image
image

engelhamer:
Now looks as follows:
image

PR #760, opened 2019-10-17
Group are now allowed to have the same name.

Larspolo:
## Description
Group are now allowed to have the same name.


src/django/test/test_group.py
```diff
@@ -106,3 +110,8 @@ def test_members(self):
         api.create(self, 'members',
                    params={'group_id': self.group.pk, 'user_id': self.teacher.pk}, user=factory.Student(), status=403)
         api.get(self, 'members', params={'group_id': self.group.pk}, user=factory.Student(), status=403)
+
+    def test_group_api(self):
+        course = factory.Course(active_lti_id='6068')
+        nfac.make_lti_groups(course)
```
mjvkeulen: nfac?

PR #763, opened 2019-10-18
Case for when attempting to link an account with an already taken email address

### LINKED ISSUE #511: Server error (duplicate email) when connecting real account LTI account
**Describe the bug**
I get a duplicate email error when trying to link my account from datanose to the ejournal tool. This is caused by another account that already has this email, but I dont want to connect it to that account.

**To Reproduce**
1. Create on eJournal two accounts where 1 has an email that is the same on DataNose/Canvas
2. Go from DN/Canvas to eJournal, and try to link it to the account that does not have the same email. 

**Expected behavior**
either a message "there is already another account witht he same email" or remove that email from that other account and notify the user, or lastly (and preferred) immediately give the option to connect to that account with the same email.

**Screenshots**
image

**Desktop:**
 - OS: ubuntu
 - Browser firefox

mjvkeulen:
## Description
When attempting to link an account from a LMS to an eJournal account it is possible the email address is already taken by a different account on the eJournal side. This scenario will no longer yield a 500 error,  and inform the user of the situation. 
As the user supposedly has access to the claimed email address, he or she should be able to link to that specific account instead.

## Summary of changes
- Fixes #511


src/django/VLE/views/user.py
```diff
@@ -185,6 +185,10 @@ def partial_update(self, request, *args, **kwargs):
         if user_image is not None:
             user.profile_picture = user_image
         if user_email:
+            if User.objects.filter(email=user_email).exclude(pk=user.pk).exists():
+                return response.bad_request(
+                    '{} is claimed by another user, please link to that account or contact support.'.format(user_email))
```
engelhamer: I'd go with `is taken` rather than `is claimed`?

engelhamer: Also: by another `account` instead of `user`, since we speak of an account in the second half of the sentence (could be separated by a period as well).

PR #771, opened 2019-10-20
Added the ability to copy an assignment and shift deadlines

### LINKED ISSUE #710: Possibility to shift all the dates in an assignment by a year
The Possibility to shift all the dates in an assignment by a year.
This is useful when copying from an old assignment. It would be nice if all the dates would translate to the current year.

### LINKED ISSUE #778: Copy assignment causes 500 error for existing journals
**Describe the bug**
Upon changing an assignment  by copying its structure from another assignment, loading an existing journal for the updated assignment causes a 500 error (traced back to the `get_nodes` function).

**To Reproduce**
1. Fill the preset database
2. As the `Teacher`, copy the assignment structure from 'Logboek' into 'Colloquium'
3. Try to load a student journal

**Expected behavior**
Existing journals should obviously remain functional.

Maybe we should not allow copying once an assignment has been created.

mjvkeulen:
## Description
Expanded the format copy logic to assignment copy functionality. The copy functionality is no longer positioned in the assignment edit page but is now moved to the course edit page.

Copy logic is accessible from the Course page:
image

As well as from the lti launch screen:
image

image

When copying an assignment format, it is possible to set how many months all deadlines should be moved forward in time. This process attempts to conserve the weekdays while matching the requested offsets as closely (in full or more) as possible.

After a successful copy the user is directed to the format edit of the copied assignment.

## Summary of changes
- Fixes #710 
- Fixed #778 
- A bug which breaks LTI launch on link to existing course


src/vue/src/views/FormatEdit.vue
```diff
@@ -131,15 +131,42 @@
...
                     ref="copyFormatModal"
                     size="lg"
                     title="Select an assignment to copy"
-                    hideFooter
                 >
                     <format-copy-selection
                         :copyableFormats="copyableFormats"
                         @copyFormat="copyFormat"
                     />
+                    <template v-slot:modal-footer>
```
engelhamer: I don't like the way this looks: it should be moved to the selection card above, and only show in case an asssignment has been selected (adding a button 'Copy assignment' to complete that would then also be nice).

src/django/VLE/views/format.py
```diff
@@ -12,6 +13,29 @@
 from VLE.serializers import AssignmentDetailsSerializer, FormatSerializer
 
 
+def day_neutral_datetime_increment(date, years=0, months=0):
+    return date + relativedelta(years=years, months=months, weekday=date.weekday())
+
+
+def set_assignment_dates(from_assignment, to_assignment, years=0, months=0):
```
engelhamer: I find it unintuitive that a teacher has to first create an assignment with ssome dates, which are then shifted anyways by this offset based on the original assignment. It would make more sense to ad a 'copy existing assignment' option to the LTI setup / course screen (if we are going to stick to the original deadlines anyways). This copy would then also include the description etcetera.

engelhamer:
Also, copying an assignment format can break existing journals in some cases: this results in 500 errors.

mjvkeulen:
> Changing the dates for the assignment that is being copied to makes setting these dates not logical in the first place. We should come up with a more intuitive solution for this.

The dates are based on the source, not the assignment where we are copying into? Could you explain why this is not logical?

engelhamer:
If I want to copy an existing assignment, I now have to first create one with arbitrary dates which are then overwritten on copy.

mjvkeulen:
I can get behind copying an assignment and making a dedicated trigger. I would place this in the course edit page. The shift logic is still obvious for the use case setting up an assignment for next round of fresh students.

src/vue/src/views/FormatEdit.vue
mjvkeulen: Can you explain what you dislike?

engelhamer:
>
> This does not adhere to the style we have across the website: modal
contents should always be inside of a card. Besides, I find it illogical
that clicking an assignment triggers the copy action while the input fields
below influence the result of the copy (in terms of dates).

>
>
>

src/vue/src/views/Course.vue
```diff
@@ -165,3 +180,11 @@ export default {
     },
 }
 </script>
+
+<style lang="sass">
+@import '~sass/modules/breakpoints.sass'
+
+@include max-width(439)
```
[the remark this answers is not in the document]
mjvkeulen: Its the exact breakpoint where the two add and copy buttons no longer fit on the same horizontal row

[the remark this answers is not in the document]
engelhamer: This will not work, as we make use of relative sizes (em) for both the padding and font of the button.

mjvkeulen:
The highlighted selection color is too subtle on some screens:
image

What would you find a workable alternative?

PR #794, opened 2019-11-15
Fixed issue where LTI wouldn't load with too large parameters

Larspolo:
## Description
Fixed issue where LTI wouldn't load when jwt_params where too long (4k+). There now is a max length on the url of 8k (was 4k). Also, all requests urls now dont include crazy long parameters anymore, they are all moved to the request body.


Larspolo:
>Any specific reason for picking 8190 as the new max length? Is it unreasonable to assume any longer requests? (Otherwise looks fine)

From the docs: "Value is a number from 0 (unlimited) to 8190."
This is the maximum (that is not unlimited) value

system_configuration_tools/roles/web/templates/gunicorn_start.j2
```diff
@@ -35,4 +35,5 @@ exec gunicorn \
     --user $USER --group $GROUP \
     --log-level debug \
     --bind unix:$SOCKFILE \
+    --limit-request-line 8190 \
```
mjvkeulen: Great find! I would prefer this to be moved to a configuration option. Most sense would be the defaults found here: 
`system_configuration_tools/roles/web/defaults/main.yml`

engelhamer:
The line limit was not moved to the config file?

PR #798, opened 2019-11-28
Fixed an issue where two entries were posted when double-clicking the entry post button

### LINKED ISSUE #777: Disable entry post button while request is being processed
Requests are currently sent twice after double clicking the post button. It should be disabled after clicking once, and enabled after the completion toast is shown.

engelhamer:
## Description
Fixed an issue where two entries were posted when double-clicking the entry post button.

## Summary of changes
- Fixes #777 
- Also fixed for comment send and edit


src/vue/src/components/entry/CommentCard.vue
```diff
@@ -252,10 +258,14 @@ export default {
         editComment (cID, index) {
             this.$set(this.commentObject[index], 'text', this.editCommentTemp[index])
             this.$set(this.editCommentStatus, index, false)
+            this.saveRequestInFlight = true
             commentAPI.update(cID, {
                 text: this.editCommentTemp[index],
             })
-                .then((comment) => { this.$set(this.commentObject, index, comment) })
+                .then((comment) => {
+                    this.saveRequestInFlight = false
```
Larspolo: This should also be in the catch right? Now when a post fails, the page needs to be reloaded to post again

engelhamer: That's right. Will add a catch on all occurences.

Issue #782, opened 2019-11-01
Images cause rich text fields to partially hide content


Larspolo:
**Describe the bug**
![image](https://user-images.githubusercontent.com/11822478/68026440-0895f600-fcb0-11e9-8d24-d9281ce9146a.png)

The image is not fully shown as the iframe cuts it off. This is most likely a loading issue where the hight of the images is not yet properly set.

engelhamer:
Interesting one: the iframe onLoad should trigger a resize, but from some testing it looks as if the images are not included here. Additional downside is that in this context the iframe does not support javascript, since we fill it with user-generated content. Otherwise it would be possible to add an onload to all image tags.

engelhamer:
This can be generalised to all rich text fields (iframes), not just entries.

engelhamer:
Fixed in #799.

PR #799, opened 2019-11-30
Fixed an issue where images in rich text fields caused some content to be hidden

engelhamer:
## Description
Fixed an issue where images in rich text fields caused some content to be hidden. This is now solved using `srcdoc` for the sandboxed iframes, however this functionality is not supported by Internet Explorer and Edge. Global browser support is [92.5%](https://caniuse.com/#feat=iframe-srcdoc) of users at the time of writing.

Users with unsupported browsers will not be affected by this change, and can still experience the issue in some occasions. 

## Summary of changes
- Fixes (to the extent which is currently feasible) #782 
- Removed the slight gray border around FABs (since it should not be displayed there)


Larspolo:
Should we alert edge / IE users that some functionallity might not fully work?

src/vue/src/components/assets/SandboxedIframe.vue
```diff
@@ -1,6 +1,8 @@
 <template>
     <iframe
+        :srcdoc="content"
         sandbox="allow-same-origin allow-popups"
+        height="0"
```
mjvkeulen: Is this added to force rescaling?

engelhamer: This is added so the iframes only increase in height upon content injection, and never decrease. The default height of an iframe caused not-so-fancy resizes when loading (e.g. first several lines height, then when content was injected the height decreased to only one line).

PR #800, opened 2019-12-02
Made usernames and email addresses case insensitive

### LINKED ISSUE #610: Ignore capitalisation in usernames
It makes no sense that different capitalisation can yield different unique usernames. This is also quite dangerous, since e.g. a capital letter i could be used to imposter a lowercase L: 'I' versus 'l'. All usernames should be made lowercase.

Larspolo:
## Description
It now sets every mail / username to lowercase and filters case 
independent.

## Summary of changes
- Fixes #610


src/django/VLE/models.py
```diff
@@ -256,6 +255,10 @@ def save(self, *args, **kwargs):
         if self.email == '':
             self.email = None
 
+        if isinstance(self.email, str):
+            self.email = self.email.lower()
+            self.username = self.username.lower()
```
mjvkeulen: To be a safe a username update should also always be made lowercase.

src/django/VLE/migrations/0036_mail_username_to_lower_case.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 2.1.7 on 2019-10-04 11:31
+
+from django.db import migrations, models
+
+
+def mail_username_to_lower(apps, schema_editor):
+    User = apps.get_model('VLE', 'User')
+    for user in User.objects.all():
+        u = User.objects.filter(email__iexact=user.email)
+        if u.count() == 1 and isinstance(user.email, str):
```
engelhamer: Should there be any conflicts here (which I don't expect, but still) then we should not forget manually updating them.

Larspolo: We indeed need to remember this. And there is a conflict with an mail of a test-student.

src/django/VLE/models.py
Larspolo: Save is triggered on update also right? So this already happend?

src/django/VLE/migrations/0036_mail_username_to_lower_case.py
mjvkeulen: Good practice would be to add a logger, to make sure we capture all conflicts? (Next to checking manually)

src/django/VLE/models.py
mjvkeulen: Because email can be `null` and `blank`, I am not entirely sure. I believe it goes emailfield -> charfield which has `''` as an empty value, and is thus never represented as None on a db level. However, in memory we can use None which would mean username would now not update?

mjvkeulen:
This also impacts the archive path of GDPR requests, we might be left with additional archives (minor impact).

src/django/VLE/migrations/0036_mail_username_to_lower_case.py
engelhamer: Indeed

engelhamer:
> This also impacts the archive path of GDPR requests, we might be left with additional archives (minor impact).

No problem at all, existing archives can simply be removed as a new archive will be created upon request anyways.

src/django/VLE/migrations/0036_mail_username_to_lower_case.py
Larspolo: It is just this one case whereafter it will most likely never happen again

mjvkeulen: There is a built-in logger right, so this would be 3 lines?

src/django/VLE/views/user.py
```diff
@@ -36,6 +36,7 @@ def get_lti_params(request, *keys):
 
 class LoginView(TokenObtainPairView):
     def post(self, request):
+        request.data['username'] = request.data['username'].lower()
```
engelhamer: Is this still required? If so, why only here and not for password reset?

mjvkeulen: Forgot password needs this as well on both username and email address

src/django/VLE/migrations/0036_mail_username_to_lower_case.py
mjvkeulen: Handling is still required for cases where there in fact exists emails or usernames which are equal without capitalization

src/django/test/test_user.py
```diff
@@ -244,6 +244,12 @@ def test_login(self):
         api.login(self, user, password="wrong", status=401)
         assert old_last_login == User.objects.get(pk=user.pk).last_login, 'Last login should not be updated'
 
+        # Check that login should laos be possible with different capitalizations
```
mjvkeulen: laos?

PR #801, opened 2019-12-04
Improved email handling

mjvkeulen:
## Description
- Moved to Mailgun as our email processor in order to be better equipped with the increase in volume we have seen lately.
- Rework of our deploy inventory structure.
- Improved email test case (can be extended now if we wish)


system_configuration_tools/roles/web/templates/maintenance_off.html
```diff
@@ -17,7 +17,7 @@
     <h1><span>e</span>Journal</h1>
     <h2>We&apos;ll be back soon!</h2>
     <div>
-        <p>Sorry for the inconvenience but we&apos;re performing some maintenance at the moment. If you need to you can always <a href="mailto:contact@ejourn.al">contact us</a>, otherwise we&apos;ll be back online shortly!</p>
+        <p>Sorry for the inconvenience but we&apos;re performing some maintenance at the moment. If you need to you can always <a href="mailto:contact@ejournal.app">contact us</a>, otherwise we&apos;ll be back online shortly!</p>
```
engelhamer: Would be more logical to use the support address here, no?

src/django/test/test_email.py
```diff
@@ -24,6 +27,10 @@ def test_forgot_password(self):
         resp = api.post(self, 'forgot_password', params={'username': self.student.username})
         assert 'An email was sent' in resp['description'], \
             'You should be able to get forgot password mail with only a username'
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].to, [self.student.email])
```
Larspolo: why the switch to assert equal?

PR #809, opened 2019-12-12
Flag the postgres database user as superuser during migrations

Larspolo:
## Description
With this change we can now add extensions to the DB using the migrations without needing to manually alter the role that handles the migrations.


system_configuration_tools/roles/web/tasks/setup_django_app.yml
```diff
@@ -42,6 +53,17 @@
...
+- name: Remove superuser privileges from db user
+  become: true
+  become_user: postgres
+  postgresql_user: db={{ db_name }}
+                   name={{ db_user }}
+                   password={{ db_password }}
+                   encrypted=yes
+                   priv=ALL
+                   role_attr_flags=SUPERUSER
+                   state=present
```
mjvkeulen: absent?

Makefile
```diff
@@ -146,6 +146,7 @@ postgres-init-development:
 	-c \"ALTER ROLE $(postgres_dev_user) SET default_transaction_isolation TO 'read committed'\" \
 	-c \"ALTER ROLE $(postgres_dev_user) SET timezone TO 'CET'\" \
 	-c \"GRANT ALL PRIVILEGES ON DATABASE $(postgres_db) TO $(postgres_dev_user)\" \
+	-c \"alter role $(postgres_dev_user) superuser\" \
```
engelhamer: Do we still need this with the latest additions?

Larspolo: For preset-db it does need superuser rights, so yes

PR #832, opened 2020-02-04
Fixed an issue where notifications wouldn't be sent to some users

Larspolo:
## Description
Fixed an issue where notifications wouldn't be send to all users. This was caused because it crashed when trying to retrieve a journal using the new queryset

## Summary of changes
- Fixes sentry 1491830387


engelhamer:
This also contains #830

PR #842, opened 2020-02-19
Added a warning message when using unsupported or outdated browsers

### LINKED ISSUE #802: Warning dialogue for unsupported browsers
We use functionality (the `srcdoc` attribute on iframes, see #799) that is not fully supported by browsers like Edge. A warning indicating preferred browsers for an optimal viewing experience would be a welcome addition.

engelhamer:
## Description
Added a warning for outdated browser versions when using eJournal.

image

The alert shows up every time the user is routed to a different page, except for when they click dismiss. Then it stays away until the app as a whole is refreshed. This makes it sufficiently annoying to stress that an update should take place, but does not completely obstruct user workflow if they just so happen to be unable to update at the time.

Internet explorer is completely blocked via a redirect from the index.

image

Browsers for which no version is whitelisted are allowed to be used, as to allow new browsers or obscure mobile stuff.

## Summary of changes
- Fixed #802
- Renamed `system_configuration_tools` folder to `config`: would like to move all config related matter (maybe even instance settings such as registration) there, more centralised
- Supported browsers can be configured in the `config/supported-browsers.yml` file


src/vue/public/index.html
```diff
@@ -13,6 +13,12 @@
         <meta name="msapplication-config" content="/favicon/browserconfig.xml?v=3">
         <meta name="theme-color" content="#252c39">
         <title>eJournal</title>
+        <script>
+            // TODO: Check if not triggered by anything other than IE.
```
engelhamer: Any thoughts on this?

Larspolo: Do we really need to block everything on IE? It still mostly works right? I find it personally a bit too extreme and a banner will do

Larspolo: Or a "continue anyway" button

engelhamer: I'm strongly against allowing IE anymore: rich text entries show no content at all, the timeline is destroyed and even the logo is cut off for 70%, to name a few. Besides, e.g. Canvas also blocks IE entirely (and so many users would need a different browser in the first place).

Larspolo: Solid.

Canvas uses the same, so it will probably be alright.
https://github.com/instructure/canvas-lms/blob/66d9d792e2e3e92c485a96e2b4b083996d3a21b7/app/views/layouts/_head.html.erb#L25

engelhamer: Yes, but they use an extra wrapper for apparently some windows apps require IE for oauth. We have to check if this  relevant for us.

PR #844, opened 2020-02-24
Changed timezone

### LINKED ISSUE #488: eJournal should use timezones
entry.py:68:0: `datetime.now()` should only be called once and during the start of the request, which should be saved on some request level global. Also, dates should be stored in UTC so that there are never any timezone issues with the server (changing servers with different timezone settings will have no impact), conversion to a local time should be done on the client (which can be in a different timezone than the server).

Larspolo:
## Description
Changed timezone to CET. This way all dates will be displayed at CET. Even when users are outside CET, times will be shown as CET. This may be changed in the future, but for now this is enough imo.

NOTE:
After deployment the following command needs to be executed to keep the datetimes correct.

```
from datetime import timedelta
from django.db.models import F
from VLE.models import *

Assignment.objects.all().update(due_date=F('due_date') - timedelta(hours=1), unlock_date=F('unlock_date') - timedelta(hours=1), lock_date=F('lock_date') - timedelta(hours=1))
PresetNode.objects.all().update(due_date=F('due_date') - timedelta(hours=1), unlock_date=F('unlock_date') - timedelta(hours=1), lock_date=F('lock_date') - timedelta(hours=1))
```

## Summary of changes
- Does not fully fiks #488 as that would be a full feature thing, and this is mostly just a bugfix


engelhamer:
*Previous closing was due to misclick on mobile*

I think this should simply be configured on the instance itself and not in the github repo.

PR #855, opened 2020-02-27
Improved error messages during forgot password and account creation

Larspolo:
## Description
Improved error when supplying wrong email/username in forgot password
Improved email / full_name creation check

## Summary of changes
- Fixes Sentry 1366149171
- Fixes Sentry 1360068976


src/vue/src/components/account/LoginForm.vue
```diff
@@ -118,16 +116,7 @@ export default {
     },
     methods: {
         handleForgotPassword () {
-            let username = ''
-            let emailAdress = ''
-
-            if (validation.validateEmail(this.usernameEmail, false)) {
```
engelhamer: `validateEmail` seems to be also used for login, otherwise it could have been deleted from the validation file too.

PR #861, opened 2020-03-04
Improved user file storage

### LINKED ISSUE #788: Uploading an image through the rich text editor's browse button does not work
**Describe the bug**
The rich text editor contains a button in the toolbar to upload an image from the user's device. Upon clicking it the file browser appears and a file can be selected, however the selected image is not inserted into the rich text field. Drag-and-dropping from e.g. Nautilus does seem to work.

**To Reproduce**
1. Create a new entry with a rich text field
2. Click on the image button at the top of the editor
3. Click on the browse icon
4. Select an image
5. Observe that it is not inserted

**Expected behavior**
The image should be inserted into the rich text field.

### LINKED ISSUE #807: Files can sometimes get stored with the wrong filename
**Describe the bug**
Files get stored with the wrong filename. When using sites like webp.net to resize the image, the filename sometimes does not get changed propperly. This causes our system to store them wrongly.

**To Reproduce**
1. Resize on image on webp.net
2. Upload it (with the weird filename from webp)
3. See that you are not able to download it

### LINKED ISSUE #853: Files with same name can get swapped during entry creation
**Describe the bug**
Files with same name can get swapped

**To Reproduce**
1. Create 2 different files on your computer
2. Give them the same name
3. Upload them to eJournal in the same entry but different fields (during creation of a new entry)
4. Notice that the files may be swapped

**Expected behavior**
Files to stay in their correct field

**Desktop (please complete the following information):**
 - OS: Debian
 - Browser: Firefox

mjvkeulen:
## Description

- Updated TinyMCE to v5
- Fixed a bug which prevented modal functionality of TinyMCE itself to be compatible from being launched in a modal
- Reworked file handling towards a more complete solution, we no longer embed any base64 encoded files anywhere.

Fixes #788, fixes #807, and fixes #853


engelhamer:
Sorry but have to comment up front: what's with all the `theme-TAG` spam? This does not seem like an appropriate style of working.

engelhamer:
Also there still seem to be some traces of CKeditor in the package lock etc.

mjvkeulen:
> Also there still seem to be some traces of CKeditor in the package lock etc.

Do we care about package lock? Package.json is what matters here?

mjvkeulen:
> Sorry but have to comment up front: what's with all the `theme-TAG` spam? This does not seem like an appropriate style of working.

Previously we were only required to apply some of the theme isolation to headers. With the possibility of moving towards a WYSIWYG editor that does not function from an Iframe, our own theme style was leaking into the editor. 
The theme prefix is now more consistent, and separates our own css style from that of external elements, allowing for more friendly integration with external modules.

src/vue/public/tinymce/content.css
```diff
@@ -0,0 +1,6 @@
+@import url('https://fonts.googleapis.com/css?family=Roboto+Condensed|Roboto:400,700');
+
+body {
+    font-family: 'Roboto', sans-serif;
+    font-size: 16px !important;
```
engelhamer: Why do we have a pixel value here?

src/vue/src/sass/partials/headers.sass
```diff
@@ -1,4 +1,4 @@
-h1
+.theme-h1
```
engelhamer: Would really like to see this different, but it's alright to tackle that in the future (#866).

src/vue/src/Header.vue
```diff
@@ -24,7 +24,10 @@
             :to="{ name: 'Home' }"
             class="brand-name text-shadow"
         >
-            <img src="/ejournal-logo-white.svg"/>
+            <img
+                src="/ejournal-logo-white.svg"
+                class="theme-img"
```
engelhamer: Is this class stil relevant here?

src/vue/src/Header.vue
```diff
@@ -120,7 +123,10 @@
             :to="{ name: 'Guest' }"
             class="brand-name"
         >
-            <img src="/ejournal-logo-white.svg"/>
+            <img
+                src="/ejournal-logo-white.svg"
+                class="theme-img"
```
engelhamer: Dito.

src/vue/src/components/assets/BreadCrumb.vue
```diff
@@ -19,7 +19,7 @@
         </b-button>
         <div>
             <div v-if="crumbs.length > 1">
-                <h4>
+                <h4 class="theme-h4">
```
engelhamer: We could also scope this more to the breadcrumb itself, as I reckon this style is not used much elsewhere? (maybe in #866)

src/vue/src/components/assets/file_handling/FileDownloadButton.vue
```diff
@@ -4,18 +4,17 @@
         @click="fileDownload"
     >
         <icon name="download"/>
-        <i><span>{{ fileName }}</span></i>
+        <i><span>{{ file.file_name }}</span></i>
```
engelhamer: Is `file` not a String itself?

src/vue/src/components/profile/ProfileData.vue
```diff
@@ -116,9 +120,9 @@ export default {
         },
         fileHandler (dataURL) {
             userAPI.updateProfilePictureBase64(dataURL, { customSuccessToast: 'Profile picture updated.' })
```
engelhamer: Maybe B64 output from the cropper can change to png instead.

src/django/VLE/views/entry.py
```diff
@@ -141,27 +157,51 @@ def partial_update(self, request, *args, **kwargs):
...
+                if field.type in field.FILE_TYPES:  # Image, file or PDF
+                    if field.required and not data:
+                        raise FileContext.DoesNotExist
+                    if data:
+                        files_to_establish.append(
+                            (FileContext.objects.get(pk=int(data)), old_content))
+
+                # Establish all files in the rich text editor
+                if field.type == Field.RICH_TEXT:
+                    files_to_establish += [(f, old_content) for f in file_handling.get_files_from_rich_text(data)]
```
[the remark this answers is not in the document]
Larspolo: They are created on upload, after posting they only need to be established: meaning they dont get deleted when purging old / unestablished files

src/vue/src/components/assets/file_handling/FileDownloadButton.vue
Larspolo: It is an object with download-url and name for example

engelhamer: But then why is the prop defined as String below?

engelhamer:
Also: which issues does this PR close? I reckon #788, #807, maybe #853, something else?

src/django/VLE/migrations/0038_file_rework.py
```diff
@@ -0,0 +1,334 @@
...
+# Expects a string containing a single base64 file
+def base64ToContentFile(string, filename):
+    matches = re.findall(r'data:(.*);base64,(.*)', string)[0]
+    mimetype = matches[0]
+    extension = guess_extension(mimetype)
+    return ContentFile(base64.b64decode(matches[1]), name='{}{}'.format(filename, extension))
+
+
+def fileToEmbdeddedImageLink(file):
+    return '<img src="{}/files/{}/access_id/"/>'.format(settings.API_URL, file.access_id)
```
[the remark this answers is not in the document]
Larspolo: 'files/<int>' requests ask for an integer (the primary key). Access_id is not an int, and therefor needs its own url. `/files/{}/access_id/` could be changed to `/files/{}/whatever/`, but imo not necessary. Users should never see/use it like this anyway

src/django/VLE/models.py
```diff
@@ -35,79 +37,107 @@ def to_string(self, user=None):
         return self.name
 
 
-class UserFile(models.Model):
-    """UserFile.
+# https://stackoverflow.com/a/2257449
```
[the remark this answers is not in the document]
Larspolo: I could swear I copied that one. Well, good catch!

src/django/VLE/migrations/0038_file_rework.py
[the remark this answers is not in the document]
engelhamer: Yes, but would it not make more sense to include it as a query param?

src/vue/src/components/assets/file_handling/FileDownloadButton.vue
Larspolo:
```        file: {
            required: true,
        },```
Not anymore?

File: src/django/VLE/views/user.py [hunk 11]
```diff
@@ -419,14 +340,27 @@ def set_profile_picture(self, request):
...
-
-        request.user.profile_picture = request.data['file']
+        def base64ToContentFile(string, filename):
+            matches = re.findall(r'data:(.*);base64,(.*)', string)[0]
+            mimetype = matches[0]
+            extension = guess_extension(mimetype)
+            return ContentFile(base64.b64decode(matches[1]), name='{}{}'.format(filename, extension))
+
+        file_data, = utils.required_params(request.data, 'file')
+        content_file = base64ToContentFile(file_data, 'profile_picture')
```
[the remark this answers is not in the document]
Larspolo: B64 is a representation of the bytes of any file in ascii. The type can be either jpg or png. I now changed it to represent an PNG image as that allows for transparency, but it is still b64

src/vue/src/Header.vue
Larspolo: Not now, but if we want to change the style of all theme-img in the future, it will update automatically. Do you still suggest removing it?

engelhamer: Fine

src/vue/src/components/assets/BreadCrumb.vue
Larspolo: image
Couple cases actually, but might be good to take a look at it in #866.

src/django/VLE/models.py
```diff
@@ -35,79 +37,107 @@ def to_string(self, user=None):
...
     """
     file = models.FileField(
         null=False,
-        upload_to=get_path
+        upload_to=file_handling.get_file_path
+    )
+    in_rich_text = models.BooleanField(
+        default=False
+    )
+    access_id = models.CharField(
```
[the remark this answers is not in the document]
mjvkeulen: Hm, should we then also account for this on save? If we assume duplicates will (realistically) be drawn, we should also make sure these cause no actual errors

PR #872, opened 2020-03-10
Added rich text support for preset descriptions

Larspolo:
## Description
Added rich text to progress node description


Larspolo:
Will add this to entry presents as well

Issue #606, opened 2019-03-08
Custom TinyMCE skin


mjvkeulen:
**Describe the solution you'd like**
Implementation of a custom TinyMCE skin more in line with the rest of the theme of the project. See:
http://skin.tinymce.com/ (The editor currently only supports v4, which at this time is also our TinyMCE version)

PR #878, opened 2020-03-16
Fixed various user interface bugs

### LINKED ISSUE #783: Date time picker gets cleared when set equal to max date
**Describe the bug**
When e.g. a due date for a progress goal is set to the maximum possible date (in this case the assignment due date) the `flatpickr` date time picker clears its contents.

**To Reproduce**
1. Create an assignment with X points possible and a due date
2. Go to the assignment editor
3. Click on the auto-created progress goal with the 'X' points
4. See that the due date picker appears empty, though the due date for the goal is still visible in the timeline

**Expected behavior**
The max date should be inclusive. According to the [flatpickr docs](https://flatpickr.js.org/options/) it should be, but it looks like this option might clash with the custom date format option.

**Screenshots**
image

engelhamer:
## Description
Updated the flatpickr date picker and fixed a major issue. Also restyled it a bit, it now looks as follows:

image

## Summary of changes
- Fixed #783
- Fixed some issues with rich text editors showing content of other fields in the template editor
- Improved the styling of several elements:
image
image
- Made the rich text editor menu bar more consistent with our style, however a custom skin (#606) would be needed to gain a bit deeper control of the UI elements such as dropdowns and modals inside of the editor.
- Cleaned up sass and classes (got rid of some `important` spam 🎉)


src/vue/src/sass/partials/inputs.sass
```diff
@@ -35,11 +35,39 @@
...
+    margin-bottom: 10px
+    border-radius: 5px
+    cursor: pointer
+
+.flatpickr-calendar
+    @extend .shadow
+    font-family: 'Roboto Condensed', sans-serif
+    border: 1px solid $theme-medium-grey
+    border-radius: 5px
+    // overflow: hidden
```
mjvkeulen: Left over

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -87,9 +93,9 @@
                             <text-editor
                                 v-if="showEditors"
                                 :id="`rich-text-editor-field-${template.id}-${field.location}`"
+                                :key="`rich-text-editor-field-${template.id}-${field.location}`"
                                 v-model="field.description"
                                 :basic="true"
-                                :displayInline="true"
```
mjvkeulen: What was the motivation behind this change?

PR #880, opened 2020-03-16
Added a tooltip indicating password requirements

### LINKED ISSUE #827: Display password requirements in tooltip
It would also be nice to have a tooltip next to (all?) password fields that state the rules a password need to follow.

engelhamer:
## Description
Added a tooltip indicating password requirements.

## Summary of changes
- Fixes #827


engelhamer:
> Missing tooltip at login, on purpose?

Yes, as the user simply needs to know their own password there. The password requirements are only relevant for setting a new password.

> Also, maybe this is now better in a component as it is copied all over

I think this is too small to make into a component. We could move all password creation stuff to a component (since these two fields always come in a pair) and include the checks there, then pass back only a single value (the password) making it a bit more appropriate to split. Do you value that?

PR #881, opened 2020-03-17
Fixed a bug which could crash rich text editor instances on Firefox

### LINKED ISSUE #884: Sandboxed iframes do not size images correctly on Firefox
**Describe the bug**
When submitting an image as the content of a sandboxed iframe, the first render is clipped vertically.

**To Reproduce**
1. Use Firefox
2. Go to the format editor
3. Select a template
4. Pass an image into a rich text field
5. Click on preview
6. Observe vertically clipped render of image
7. Toggle edit and preview, observe correctly loaded image

**Screenshots**
First render:
image

Second render:
image

**Desktop**
 - OS: Ubuntu 18.04
 - Browser: Firefox
 - Version 74.0

mjvkeulen:
## Description
Moved the destroy hook of TinyMCE over to call remove instead of the destroy API of TinyMCE. This should fix the user interruptions we are experiencing on firefox.

This is still not a clean solution, as the error still occurs (in the TinyMCE logic), however as we do not reuse TinyMCE editor instances, this should fix the problems the user encounters. 

Fixes #884


src/vue/src/components/assets/TextEditor.vue
```diff
@@ -195,7 +195,12 @@ export default {
         tinymce.init(this.config)
     },
     beforeDestroy () {
-        if (this.editor) { this.editor.destroy() }
+        try {
+            if (this.editor) { tinymce.remove(this.editor) }
+        } catch (NS_ERROR_UNEXPECTED) {
```
engelhamer: Why is this catch still needed?

mjvkeulen: Sadly as my note says I have no idea atm

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -19,7 +19,7 @@
             </b-button>
         </div>
         <hr/>
-        <div v-if="mode === 'edit'">
+        <div v-show="mode === 'edit'">
```
engelhamer: Why this?

Larspolo: reduces time to switch between edit and preview. Makes it way better to "quickly check if it is alright"

Larspolo: Can take multiple seconds on non desktops without this

PR #893, opened 2020-03-22
Moved the deactivation of the maintenance page to a handler

mjvkeulen:
## Description
We want the deactivation of the maintenance page to be (one of) the last executed actions on a deployment action.


mjvkeulen:
I overlooked the include was not tagged, reworked to work for all call methods.

PR #895, opened 2020-03-23
Speedup in permission checking

Larspolo:
## Description
Speedup in permission checking
Practical result:
image

Side effect: has_permission now no longer checks if user is participant.
This practically means quicker a permission error iso a participation
error. However, we always check for participation beforehand anyway.

(also added more testing)


src/django/test/test_permissions.py
```diff
@@ -255,30 +279,44 @@ def test_serialize(self):
...
+
+        high1 = factory.make_role_default_no_perms(
+            "HIGH", self.course1, can_view_course_users=True, can_view_all_journals=True)
+        factory.make_participation(high_user, self.course1, high1)
+
+        high2 = factory.make_role_default_no_perms(
+            "HIGH", self.course2, can_view_course_users=True, can_view_all_journals=True)
+        factory.make_participation(high_user, self.course2, high2)
+
+        middel1 = factory.make_role_default_no_perms("MIDDEL", self.course1, can_view_course_users=True)
```
mjvkeulen: Dutch?

PR #902, opened 2020-03-27
Fixed an issue where members of a group journal could get multiple deadline emails

Larspolo:
## Description
- Fixed bug where users in group assignments would get multiple emails 
and
- Deadline notifications are now better tested


src/django/VLE/tasks/beats/notifications.py
```diff
@@ -55,7 +55,7 @@ def _send_deadline_mails(deadline_query):
     no_submissions = Q(type=Node.ENTRYDEADLINE, node__entry__isnull=True) | Q(type=Node.PROGRESS)
     notifications_enabled = Q(node__journal__authors__user__preferences__upcoming_deadline_notifications=True)
     verified_email = Q(node__journal__authors__user__verified_email=True)
-    deadlines = deadline_query.filter(notifications_enabled & verified_email & no_submissions)\
+    deadlines = deadline_query.filter(notifications_enabled & verified_email & no_submissions).distinct()\
```
engelhamer: This query is still not suitable for group journals: if _some_ journal member has the deadline reminders active and _some_ user has a verified email address, **all** members will receive the reminder regardless of their preference or verification status.

engelhamer: Might actually be an interesting test too.

Issue #904, opened 2020-03-27
More fine grained errors


engelhamer:
**Describe the solution you'd like**
We are not very reluctant in showing an error page to the user. This is done for any 403, 404 or 5XX request really. This can cause some confusing scenarios, e.g. when some element in the sidebar causes a 404 due to a bug. In this case, the entire page would not be usable anymore.

Naturally, if the user navigates to a course that does not exist (thus also triggering a 404) one would expect a redirect to the error page. The distinction between the two situations described here is not clear in our current error handling logic, and should be made more explicit.

PR #905, opened 2020-03-27
Fixed an issue where simultaneous failed requests caused multiple redirects

### LINKED ISSUE #903: NavigationDuplicated errors to login and error page
**Describe the bug**
Vue throws a `NavigationDuplicated` error when some redirects to the error and login page occur.

engelhamer:
## Description
In many occasions multiple async requests are fired upon loading a page. If for some reason multiple of these would fail, we would re-route to the error page multiple times as well. This caused `NavigationDuplicated` errors.

An example of when this happened was when a user navigates to the course page of a non existing course. Load attempts were made for both the upcoming deadlines section for that course, the course data itself and the assignments for the course, all of which failed and pushed a 404 paged rapidly after another. The same holds when a user was not logged in.

In general, it would be nice to have these errors more fine-grained (see #904).

## Summary of changes
- Fixes #903


src/vue/src/api/auth.js
```diff
@@ -104,7 +104,7 @@ function handleError (error, connArgs) {
                 description: response.data.description,
             },
         })
-    } else {
+    } else if (!connArgs.redirect) {
```
engelhamer: Prevents a flood of toasts if a re-route to the login or error page has taken place already.

PR #906, opened 2020-03-29
Addition of computed fields and a performance profiler (Silk) to the local and development environments

Larspolo:
## Description
Added silk to local and development for easier optimalization development.

After `make setup`, visit localhost:8000/silk to see


requirements/ci.txt
```diff
@@ -17,3 +17,4 @@ pytest==5.2.0 # Testing library
 pytest-cov==2.7.1 # Pytest
 pytest-django==3.5.1 # Pytest
 pytest-xdist==1.30.0 # Run Pytest in parallel
+django-silk==4.0.1 # Optimazation development
```
mjvkeulen: Spelling and not really what Silk does?

src/django/VLE/settings/development.py
```diff
@@ -28,9 +28,18 @@
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 SECURE_SSL_REDIRECT = True
 
-MIDDLEWARE = ['corsheaders.middleware.CorsMiddleware'] + MIDDLEWARE
```
mjvkeulen: I prefer keeping CSP on all environments, and for local (maybe development?) allow inline css for Silk

requirements/ci.txt
```diff
@@ -17,3 +17,4 @@ pytest==5.2.0 # Testing library
 pytest-cov==2.7.1 # Pytest
 pytest-django==3.5.1 # Pytest
 pytest-xdist==1.30.0 # Run Pytest in parallel
+django-silk==4.0.1 # Profiling / Optimisation
```
mjvkeulen, line 20: Why does the CI env require Silk?

src/django/VLE/settings/base.py
```diff
@@ -125,7 +125,6 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'VLE.utils.error_handling.ErrorMiddleware',
-    'csp.middleware.CSPMiddleware',
```
mjvkeulen, line 128: Currently breaks, e.g. on the development env.

mjvkeulen:
Non-local environment require CSP exceptions to handle the inline CSS of Silk

requirements/ci.txt
Larspolo: Think CI uses local settings

src/django/VLE/models.py
```diff
@@ -1014,34 +1015,65 @@ class Journal(models.Model):
...
+    @computed(models.TextField(null=True), depends=['authors', 'authors#user', 'authors.user#full_name'])
+    def name(self):
+        if self.stored_name:
+            return self.stored_name
+        return ', '.join(self.authors.values_list('user__full_name', flat=True))
 
-    def get_image(self):
-        if self.image:
-            return self.image
+    @computed(models.TextField(null=True), depends=['authors', 'authors#user', 'authors.user#profile_picture'])
```
Larspolo, line 1046: Mostly check if these dependencies are set correctly

src/django/VLE/urls.py
```diff
@@ -49,4 +49,6 @@
     path('update_lti_groups/', lti.update_lti_groups, name='update_lti_groups'),
 
     path('names/<int:course_id>/<int:assignment_id>/<int:journal_id>/', common.names, name='names'),
+
+    url(r'^silk/', include('silk.urls', namespace='silk'))
```
mjvkeulen, line 53: This will not work for all environments

mjvkeulen: Bump, should we just make this a conditional registration?

Larspolo: Yes, but was prioritizing notfications over this PR, as it is mostly needed for expansion as we discussed

.travis.yml
```diff
@@ -31,7 +31,7 @@ env:
     - BROKER_URL="amqp://ejournal:password@localhost/ejournal"
     - LTI_SECRET="4339900ae5861f3086861ea492772864"
     - LTI_KEY="0cd500938a8e7414ccd31899710c98ce"
-    - DJANGO_SETTINGS_MODULE="VLE.settings.local"
```
mjvkeulen, line 34: We have now opted to work with an additional settings module dedicated to travis. This configuration differs from local primarily by excluding silk, right? Does this then warrant a full settings file? Does our Travis config differ further?

Larspolo: Does this then warrant a full settings file? Think so

Does our Travis config differ further? No

engelhamer:
@Larspolo @mjvkeulen have you decided on this yet?

Larspolo:
Development now also has silk: https://github.com/eJourn-al/eJournal/pull/906/commits/2aba43afeee6b6bc51e899b467641e80e7cefed8

PR #908, opened 2020-03-31
Improved the load times of the assignment page

Larspolo:
## Description
Speedup journal serializer

authors are now requested in a seperate request. This is only needed for group assignments and when it is expanded (adding / removing members)


src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -126,46 +127,20 @@ export default {
...
-            return this.journal.stats ? this.journal.stats.submitted - this.journal.stats.graded : 0
-        },
-        unpublished () {
-            return this.journal.stats ? this.journal.stats.graded - this.journal.stats.published : 0
+            return this.journal.groups.join(', ')
         },
         squareInfo () {
             const info = []
-            if (this.markingNeeded === 1) {
+            if (this.journal.stats.marking_needed === 1) {
```
engelhamer: We previously checked if stats existed in the first place, should still be needed right?

src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -126,46 +127,20 @@ export default {
...
                 info.push('an entry needs marking')
-            } else if (this.markingNeeded > 1) {
-                info.push(`${this.markingNeeded} entries need marking`)
+            } else if (this.journal.stats.marking_needed > 1) {
+                info.push(`${this.journal.stats.marking_needed} entries need marking`)
             }
-            if (this.unpublished === 1) {
+            if (this.journal.stats.unpublished === 1) {
                 info.push('an entry needs to be published')
```
engelhamer: A grade needs to be published instead of entry?

src/vue/src/components/journal/JournalDetails.vue
```diff
@@ -22,7 +22,7 @@
             {{ journal.name }}
         </span>
         <progress-bar
-            :currentPoints="journal.stats.acquired_points"
+            :currentPoints="journal.grade"
```
engelhamer: I like this name way better by the way 👍

src/vue/src/Store.vue
```diff
@@ -18,43 +18,31 @@ export default {
...
         }
 
         function searchFilter (journal) {
             return journal.name.toLowerCase().includes(searchValue.toLowerCase())
-                || journal.authors.some(
-                    author => author.user.username.toLowerCase().includes(searchValue.toLowerCase()))
-                || journal.authors.some(
-                    author => author.user.full_name.toLowerCase().includes(searchValue.toLowerCase()))
+                || journal.usernames.toLowerCase().includes(searchValue.toLowerCase())
+                || journal.full_names.toLowerCase().includes(searchValue.toLowerCase())
```
engelhamer: `full_names` are not part of the journal serializer so this breaks the search filter. Should be added.

src/vue/src/components/assignment/JournalCard.vue
Larspolo: This was outdated. Now stats always exists if the journal exists. Everyone who can serialize the journal, should also be able to see the journal

engelhamer: Yes but this is a front end matter: if `journal` is not yet loaded, `stats` will be undefined and thus reading any properties from it will cause errors right when loading the page.

Larspolo: if journal is not yet loaded `this.journal.stats` would also crash. Thing journal-card is really only loaded when the journal is also loaded. Other calculations also assume and assumed that stuff like authors and stats were already loaded

PR #910, opened 2020-04-01
Fixed an issue where empty optional rich text fields would cause an error

Larspolo:
## Description
Fixed bug where empty optional rich text fields would cause an error

Fixes Sentry#17


src/django/VLE/utils/file_handling.py
```diff
@@ -124,13 +124,13 @@ def establish_file(author, identifier, course=None, assignment=None, journal=Non
 
 
 def get_files_from_rich_text(rich_text):
+    if rich_text is None or len(rich_text) < 128:
```
engelhamer, line 127: What's the 128 for?

Larspolo: Because we use IDs longer than 128, it will  never have results that are less than 128, no might as well skip the regex

PR #911, opened 2020-04-01
Fixed issue where user would be send to login page while already being there

Larspolo:
## Description
Fixed issue where user would be send to login page while already being there

- Fixes Sentry#19


engelhamer:
..so it turns out `auth` was not the only redirect to login, should have been fixed in #905.

PR #915, opened 2020-04-06
Fixed an issue where a bad request response message was not shown

Larspolo:
## Description
Fixed issue where bad request response message was not showing

Also now sending an empty comment will return the proper response.

## Summary of changes
- Fixes Sentry#29


src/django/VLE/views/comment.py
```diff
@@ -146,14 +148,15 @@ def partial_update(self, request, *args, **kwargs):
             return response.forbidden('You are not allowed to edit this comment.')
 
         comment.last_edited_by = request.user
+
+        if not text:
+            return response.bad_request('Comment cannot be empty.')
+
+        comment.text = text
+        comment.published = published or not request.user.has_permission('can_grade', assignment)
```
engelhamer, line 156: I don't think that we offer the ability to unpublish a comment via any UI option. This can therefore be removed.

Larspolo: I see it being added in the future though. Why not already put the possibility  in the backend?

engelhamer: Because we want to take it one step at a time?

engelhamer: Now it only adds confusion as this is not possible through normal behaviour, and also unpublishing comments feels pretty unnatural to me.

Larspolo: This was already possible, only thing I did was restrict that students cannot create unpublished comments

src/vue/src/api/auth.js
```diff
@@ -104,7 +106,8 @@ function handleError (error, connArgs) {
                 description: response.data.description,
             },
         })
-    } else if (!connArgs.redirect) {
+    } else if (router.app.$route.name !== 'ErrorPage' && (!connArgs.redirect || !ERRORS_TO_REDIRECT.has(status))) {
+        // If page is not the error page, nor requires a redirect, display toast
```
mjvkeulen, line 110: Opinions on non-temp comments in single vs multi-line format?

PR #916, opened 2020-04-06
Config improvements

mjvkeulen:
## Description
Improved on some of our config settings. Working towards making the `web` role an `ejournal` role, without tight coupling with external roles. This will eventually allow us to merge with other plays, such as those of our monitoring stack.

The following roles are improved and are now no longer tightly coupled with `web`:
- Cerbot
- Nginx

The Nginx base config is now also captured in the code as a template, this allowed for some cleanup in the eJournal Nginx config template.

Fixes:
- Authorized SSH keys
- Lower maintenance time


config/roles/nginx/tasks/main.yml
```diff
@@ -3,38 +3,32 @@
 - name: Install Nginx
   apt: name=nginx update_cache={{ update_apt_cache }} state=present
   tags: packages
+  become: yes
```
Larspolo, line 6: What is changed now these are added? Will this now also work is provision is run after is already ran without becoming "yes".

config/roles/nginx/templates/nginx.conf.j2
```diff
@@ -0,0 +1,65 @@
...
+
+    # server_names_hash_bucket_size 64;
+    # server_name_in_redirect off;
+
+    include /etc/nginx/mime.types;
+    default_type application/octet-stream;
+
+    ########## SSL Settings ##########
+
+    # Following the continually updated strong cipher list: https://cipherli.st/
```
Larspolo, line 33: image

config/roles/web/tasks/build_frontend.yml
```diff
@@ -0,0 +1,32 @@
...
+#   tags:
+#     - packages
+#
+# TODO can we get rid of this command, global is bad afterall
+# - name: npm install npm@latest -g
+#   command: npm install npm@latest -g
+
+# - name: Install node dependencies
+#   npm:
+#     path: "{{ tmp_build.path }}/src/vue/"
```
Larspolo, line 17: Why arent the packages installed as well? Now we have to be careful that those are actually installed correctly.

config/hosts.yml
```diff
@@ -1,20 +1,27 @@
 all:
+  vars:
+    base_domain: ejournal.app
```
engelhamer, line 3: Would it not make sens to move this to some other config file then?

config/hosts.yml
```diff
@@ -1,20 +1,27 @@
...
     production:
       vars:
         env: production
         server_user: mel
       hosts:
         uva.ejournal.app:
-          baselink: https://uva.ejournal.app
-          api_url: https://api.uva.ejournal.app
-          mailgun_sender_domain: uva.ejournal.app
+          ejournal_frontend_domain: uva.{{ base_domain }}
```
engelhamer, line 23: Why make the base domain a variable yet keep hardcoded subdomains?

config/roles/nginx/templates/nginx.conf.j2
engelhamer: It's that strong!

config/roles/security/vars/main.yml
```diff
@@ -1,6 +1,109 @@
 ---
 user_public_keys:
-  - ~/.ssh/id_rsa.pub
+  - !vault |
```
engelhamer, line 3: Can you explain why this has changed?

mjvkeulen: We now include a few of our own public keys, guarantees access to all of us instead of the one running the script.

config/roles/web/tasks/build_frontend.yml
mjvkeulen: Why should these be installed if we build locally?

config/hosts.yml
mjvkeulen: Could be done, go for it if you wish, however, the host file is meant to be shared across plays so it is not a strange place for such a shared variable.

config/hosts.yml
mjvkeulen: The base domain is shared across our tenants, the subdomains are static per tenant?

config/roles/nginx/tasks/main.yml
mjvkeulen: We currently run the entire play as root, this should be reworked. As I reused some of the plays from the monitoring stack, become is present on some tasks. I can rework the full play to this structure, but as is these have no effect.

config/roles/nginx/templates/nginx.conf.j2
mjvkeulen: Appears to have been discontinued end of Jan: https://raymii.org/s/blog/Cancellation_notice_for_cipherlist_ssldecoder_and_certificatemonitor.html

config/roles/certbot/defaults/main.yml
```diff
@@ -1,6 +1,23 @@
...
-certbot_admin_email: admin@ejournal.app
-certbot_script: /opt/certbot-auto
+certbot_script: certbot # Available due to isntallation via package
 certbot_output_dir: "/etc/letsencrypt/live"
+letsencrypt_cert_filename: fullchain.pem
+letsencrypt_privkey_filename: privkey.pem
+
+# Certbot auto-renew cron job configuration (for certificate renewals).
+certbot_auto_renew: true
+certbot_auto_renew_user: "{{ ansible_user | default(lookup('env', 'USER')) }}"
```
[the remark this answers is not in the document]
mjvkeulen: Updated, should be root as the user needs access to both the lets encrypt files and the permission to stop and reload the nginx service.

config/roles/nginx/templates/nginx.conf.j2
mjvkeulen: Our config seems to hold:
image

mjvkeulen: Found a respectable alternative: https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.4

config/roles/web/tasks/build_frontend.yml
Larspolo: I ment locally, the installs are needed before building, else the build would crash

mjvkeulen: Currently, our local build is handled via Makefile, are you requesting a rework to Ansible?

PR #917, opened 2020-04-08
Additional 1.3 bug fixes

engelhamer:


engelhamer:
> Dont we also want to add the new spreadsheet export for example?

That will be another version.

PR #918, opened 2020-04-08
Added the ability to export assignment results

### LINKED ISSUE #834: Export assignment results to spreadsheet
The possibility to export a spreadsheet containing e.g. username / full name, grade and who gave the grade for every student participating in an assignment. Useful for further processing elsewhere. Users should be abke to choose what specific things to export.

engelhamer:
## Description
Teachers can now export assingment results to a spreadsheet. This can be done from the assignment page:

image

A modal pops up to specify the fields:

image

It is possible to either export all journals, or just those currently in view (to which the filter applies). The spreadsheet (in `.xlsx`) is downloaded upon clicking the export button:

image

## Summary of changes
- Fixes #834


src/vue/src/main.js
```diff
@@ -165,6 +167,32 @@ new Vue({
 
             return route
         },
+        exportExcel (name, data) {
```
engelhamer, line 170: This will allow us to easily export other types of data into a spreadsheet as well: just provide an object containing as its properties the desired columns.

mjvkeulen: Since we only use this logic in one location, should this not be moved to a helper file or a similar alternative?

src/vue/src/views/Assignment.vue
```diff
@@ -139,6 +148,98 @@
                 </b-card>
             </b-modal>
 
+            <b-modal
```
mjvkeulen, line 151: This nearly adds 90 lines to the file, could this be moved to a component?

engelhamer: Yes will do.

engelhamer: Together with all related logic.

src/vue/src/views/Assignment.vue
```diff
@@ -95,12 +95,20 @@
...
                 <b-button
                     v-if="$hasPermission('can_publish_grades') && assignmentJournals.length > 0"
-                    class="change-button multi-form flex-grow-1"
+                    class="change-button multi-form fill-form-width"
                     @click="showModal('bonusPointsModal')"
                 >
                     <icon name="star"/>
                     Import bonus points
                 </b-button>
+                <b-button
```
mjvkeulen, line 104: We move towards an action section when we add the next action ^^

src/vue/src/views/Assignment.vue
```diff
@@ -582,6 +699,52 @@ export default {
...
+            }
+            if (this.exportSelection.entry_points) {
+                data.entry_points = journalsToExport.map(
+                    journal => (journal.grade - journal.bonus_points))
+            }
+            if (this.exportSelection.bonus_points) {
+                data.bonus_points = journalsToExport.map(journal => journal.bonus_points)
+            }
+
+            this.$root.exportExcel(`eJournal export ${this.assignment.name}`, data)
```
Larspolo, line 734: would prefer csv as that is more platform independant (with ;)

mjvkeulen:
The modal could include a preview of the sheet layout based on which toggles are active, opinions?

src/vue/src/views/Assignment.vue
```diff
@@ -582,6 +699,52 @@ export default {
...
+            const data = {}
+
+            this.exportInProgress = true
+
+            if (this.exportSelection.name) {
+                data.name = journalsToExport.map(journal => journal.name)
+            }
+            if (this.exportSelection.full_name) {
+                data.full_name = journalsToExport.map(
+                    journal => journal.authors.map(author => author.user.full_name).join(','))
```
Larspolo, line 713: journal doesnt have "authors" anymore. should be journal.full_names and journal.usernames

src/vue/src/views/Assignment.vue
```diff
@@ -139,6 +148,98 @@
...
+                        />
+                    </b-form-checkbox>
+                    <b-form-checkbox
+                        v-model="exportSelection.bonus_points"
+                        class="multi-form"
+                    >
+                        Bonus points
+                        <tooltip
+                            class="ml-1"
+                            tip="Amount of bonus points for a journal"
```
Larspolo, line 213: *of the journal (same goes for the others)

src/vue/src/views/Assignment.vue
```diff
@@ -582,6 +699,52 @@ export default {
...
+            }
+            if (this.exportSelection.bonus_points) {
+                data.bonus_points = journalsToExport.map(journal => journal.bonus_points)
+            }
+
+            this.$root.exportExcel(`eJournal export ${this.assignment.name}`, data)
+                .then(() => {
+                    this.exportInProgress = false
+                    this.hideModal('assignmentExportSpreadsheetModal')
+                    this.$toasted.success('Spreadsheet export completed.')
```
Larspolo, line 738: "Successfully exported to spreadsheet" would be more aligned with other messages

src/vue/src/views/Assignment.vue
engelhamer: I would not since that would make it harder to worknwith for most of our users, and _Excel_ functionality specifically keeps getting requested.

Larspolo: Iris herself uses a mac, idk if that entails the mac sheets of excel, but safari users are not small

engelhamer: They will open the file in Excel anyways, that is browser independent and available on Mac too. Linux users could open the file in Libreoffice or the like, since there really is no complicated proprietary Excel magic going on in this use case 😅

engelhamer:
> The modal could include a preview of the sheet layout based on which toggles are active, opinions?

We could do that, however I don't think anyone would benefit much. The fields that will be downloaded are clear already (so for that we don't need a preview) and we don't want people editing the Excel file in browser already. Furthermore all data is readily available in the application itself to view.

src/vue/src/views/Assignment.vue
engelhamer: I see, forgot to pull newest develop before making this PR.

src/vue/src/main.js
engelhamer: It can be moved to a helper, but the idea is that we will use this in multiple instances. If so, why do we have things like breakpoints and the like here?

src/vue/src/sass/partials/forms.sass
```diff
@@ -1,6 +1,6 @@
 .fill-form-width
     @include md
-        margin-right: 8px
+        margin: 0px 5px
```
engelhamer, line 3: Gotta check where else this is used before merge, appears to be a bit too much margin between some menu items.

src/vue/src/main.js
mjvkeulen: I believe breakpoints see much wider use. If you really foresee this logic being used all over the place this is indeed the right place. However, to me that seems unlikely?

engelhamer: E.g. exporting  course participants, groups, list of courses, journal entries and so on. But if you don't like it here I'm fine with moving it elsewhere. What would you suggest then? Importing the helper from a component?

mjvkeulen: If you feel those are reasonable scenarios leave it here and I will resolve this conversation.

src/vue/src/components/assignment/AssignmentSpreadsheetExport.vue
```diff
@@ -0,0 +1,167 @@
...
+            />
+        </b-form-checkbox>
+        <b-form-checkbox
+            v-model="exportSelection.bonus_points"
+            class="multi-form"
+        >
+            Bonus points
+            <tooltip
+                class="ml-1"
+                tip="Amount of bonus points of a journal"
```
mjvkeulen, line 52: Total number of journal bonus points

src/vue/src/components/assignment/AssignmentSpreadsheetExport.vue
```diff
@@ -0,0 +1,167 @@
...
+            <tooltip
+                class="ml-1"
+                tip="Total points as a percentage of the assignment's possible points"
+            />
+        </b-form-checkbox>
+        <b-form-checkbox v-model="exportSelection.entry_points">
+            Entry points
+            <tooltip
+                class="ml-1"
+                tip="Amount of points of a journal achieved from entries"
```
mjvkeulen, line 42: Number of journal points acquired by entry submissions

src/vue/src/components/assignment/AssignmentSpreadsheetExport.vue
```diff
@@ -0,0 +1,167 @@
...
+            <tooltip
+                class="ml-1"
+                tip="Total amount of points of a journal"
+            />
+        </b-form-checkbox>
+        <b-form-checkbox v-model="exportSelection.progress_percentage">
+            Progress percentage
+            <tooltip
+                class="ml-1"
+                tip="Total points as a percentage of the assignment's possible points"
```
mjvkeulen, line 35: assignment's possible points?

src/vue/src/components/assignment/AssignmentSpreadsheetExport.vue
```diff
@@ -0,0 +1,167 @@
...
+            <tooltip
+                class="ml-1"
+                tip="If a journal has multiple authors, usernames will be combined"
+            />
+        </b-form-checkbox>
+        <b-form-checkbox v-model="exportSelection.total_points">
+            Total points
+            <tooltip
+                class="ml-1"
+                tip="Total amount of points of a journal"
```
mjvkeulen, line 28: Total points vs possible points can be confusing

mjvkeulen:
> Could we beautify the column headers? E.g. progress_percentage to Progress percentage?

This is not yet addressed or answered?

PR #919, opened 2020-04-12
Increased max filename length and created proper response when passed

Larspolo:
## Description
Increased max filename length and created proper response when passed

Also fixed issue in keep_testing.sh where changes in media folder were also considered valid changes

## Summary of changes
- Fixes [Sentry#36](https://sentry.ejournal.app/organizations/ejournal/issues/36/)


engelhamer:
> If you upload an image with a name of length > 255 into a rich text field the error is toasted but upload still seems to succeed (image appears and also persists upon refresh of the page). Try e.g. posting a comment.

Moved to #924..

PR #922, opened 2020-04-12
Added the ability to import templates from other assignments

### LINKED ISSUE #310: Copy template from other assignment
# Feature request

## Description
For a teacher it would be nice to have an option to reuse templates, this makes it easier to not have to recreate older templates.

## User story
As a teacher you would go to the assignment edit page and upon clicking 'add template' I want to be able to select a template from another assignment that I participate in.

Larspolo:
## Description
Added ability to import templates from other assignments

## Summary of changes
- Fixes #310


src/django/VLE/views/assignment.py
```diff
@@ -432,6 +432,34 @@ def importable(self, request):
...
+        pk -- assignment to import to
+        """
+        assignment = Assignment.objects.get(pk=pk)
+        request.user.check_permission('can_edit_assignment', assignment)
+
+        template_id, = utils.required_typed_params(request.data, (int, 'template_id'))
+        template = Template.objects.get(pk=template_id)
+        request.user.check_permission('can_edit_assignment', template.format.assignment)
+
+        from_template_id = template.pk
```
mjvkeulen, line 451: Copy action has a source, might be a more apt name than `from`?

src/vue/src/components/template/ImportTemplateModal.vue
```diff
@@ -0,0 +1,140 @@
...
+
+                <theme-select
+                    v-if="selectedAssignment"
+                    v-model="selectedTemplate"
+                    label="name"
+                    trackBy="id"
+                    :options="templates"
+                    :multiple="false"
+                    :searchable="true"
+                    placeholder="Select an assignment"
```
mjvkeulen, line 58: Select a template

src/vue/src/components/template/ImportTemplateModal.vue
```diff
@@ -0,0 +1,140 @@
...
+            <div v-if="importableTemplates && importableTemplates.length > 0">
+                <h2 class="theme-h2 multi-form">
+                    Select a template to import
+                </h2>
+                <p>
+                    This action will create a new template in the current assignment that is identical to the template
+                    of your choice. Any changes made will only affect the current assignment.
+                </p>
+
+                <theme-select
```
mjvkeulen, line 20: So we make use of the theme multi-select so we can allow searching? Do we always intend to make use of the theme select even when there is only a single select scenario?

mjvkeulen, line 20: I would expect headers, Course, Assignment, Template respectively:
image
Can become a bit more ambiguous when the placeholder is not visible.

src/vue/src/components/template/ImportTemplateModal.vue
```diff
@@ -0,0 +1,140 @@
...
+                    label="name"
+                    trackBy="id"
+                    :options="templates"
+                    :multiple="false"
+                    :searchable="true"
+                    placeholder="Select an assignment"
+                    class="multi-form"
+                />
+                <hr/>
+
```
mjvkeulen, line 62: This would be a good position to insert template preview functionality. As teachers have a lot of templates this prevents switching to the respective course and assignment to verify whether the selected template is the correct template.

src/vue/src/components/template/ImportTemplateModal.vue
engelhamer: I would say that depends on the scenario, but here we indeed use it because of the search. It is also possible to replace all select instances without search requirements, but this is not really necessary.

src/vue/src/components/template/ImportTemplateModal.vue
engelhamer: Good idea!

src/vue/src/components/template/ImportTemplateModal.vue
engelhamer: I personally like this better, especially since a field only become visible once the previous has been selected. Placeholders are quite self-explanatory.

src/vue/src/components/template/ImportTemplateModal.vue
engelhamer: Has been added.

mjvkeulen:
Multiselect bug on drop-down?
image

src/django/VLE/views/assignment.py
```diff
@@ -432,6 +432,46 @@ def importable(self, request):
...
+        template_data.pop('id')
+        template_data.pop('archived')
+        template_data.pop('format')
+        template.delete()
+
+        for field in template_data['field_set']:
+            field.pop('id')
+            field.pop('template')
+
+        print(template_data)
```
mjvkeulen, line 471: left over

engelhamer:
> Multiselect bug on drop-down?
> image

Should be fixed now..

engelhamer:
> Why is the first letter capitalized of the multi-select?
> image
> 
> 
> This does not reflect the actual name?

All button and selection fields are capitalized.

engelhamer:
> > Why is the first letter capitalized of the multi-select?
> > image
> > This does not reflect the actual name?
> 
> All button and selection fields are capitalized (mainly placeholders).

Fixed now.

PR #925, opened 2020-04-14
Bumped version to 1.3.1

engelhamer:


engelhamer:
For now, versioning approach will change soon.

PR #935, opened 2020-04-21
Added a group filter to the to do section and assignment overview

### LINKED ISSUE #934: A "nothing todo" display when all todo's are done
A "nothing todo" display when all todo's are done. Even better would be to also add a section of "global" stats: stats about entries not graded in your group specifically 

Something like this in the ToDo section if everything is graded / filled in / whatever other todo stuff there is
image

engelhamer:
## Description
Added the ability to toggle filtering on a teacher's own course groups to the to do section and assignment overview. This allows teachers to view what is to be done for the groups they are responsible for at a glance, instead of having to first visit the assignment page and then filtering on their own groups there.

Also added a message when all to do's are done.

image

image

## Summary of changes
- Fixes #934
- Added persistent filter in local storage


engelhamer:
Tests will need to be added still (for the updated serializer data)

src/vue/src/components/assets/DeadlineDeck.vue
```diff
@@ -1,40 +1,80 @@
...
+                <icon
+                    name="check"
+                    scale="4"
+                    class="fill-green mb-2 text-shadow"
+                /><br/>
+                <b class="field-heading">
+                    All done!
+                </b>
+            </div>
+            You do not have any {{ isTeacher ? 'entries to grade' : 'upcoming deadlines' }} at this moment.
```
Larspolo, line 76: preferably also an indicator that it is only for the group that are being filtered, and maybe even if there are more in other groups, also state that other groups sitll have some stuff to do

src/vue/src/components/assets/DeadlineDeck.vue
```diff
@@ -93,7 +156,7 @@ export default {
 
             return deadlines
         },
-        showSortBy () {
+        isTeacher () {
             return Object.entries(this.$store.getters['user/permissions']).some(
                 ([key, value]) => ((key.indexOf('assignment') >= 0) && value.can_grade))
```
Larspolo, line 161: What is this index of check with checking if someone is a teacher? Either a comment, or a change in function name would be preferred

src/vue/src/views/AssignmentsOverview.vue
```diff
@@ -1,5 +1,28 @@
...
+                <b>own groups</b>
+            </span>
+            <span
+                v-else
+                v-b-tooltip.hover
+                title="Showing to do items for all groups"
+                @click="selectedFilterOwnGroups = true"
+            >
+                Showing to do:
+                <b>all</b>
```
Larspolo, line 23: This is now duplicate code of elsewhere right? If so, would prefer their own component to reduce code clones.

src/django/VLE/serializers.py
```diff
@@ -363,21 +363,23 @@ def get_stats(self, assignment):
...
-        if participation.exists():
-            participation = participation.first()
-            if participation.groups and users.filter(participation__groups=participation.groups.first()).exists():
-                users = users.filter(participation__groups=participation.groups.first())
+        own_participation = Participation.objects.filter(user=self.context['user'], course=course)
+        if own_participation.exists():
+            # Get all users that are in any of the request user's groups.
+            own_participation = own_participation.first()
+            own_group_users = all_users.filter(participation__groups__in=own_participation.groups.all())
+            print('\n\n\n', own_group_users.values(), own_participation.groups.all())
```
Larspolo, line 379: :(

engelhamer: Whoops forgot to push removal commit

src/vue/src/views/AssignmentsOverview.vue
engelhamer: This is merely a very specific clickable text so I do not really think that we would benefit much from making it into a separate component.

src/django/VLE/serializers.py
```diff
@@ -363,21 +363,22 @@ def get_stats(self, assignment):
 
         course = self.context.get('course', None)
         if course is None:
+            # Get the stats from only the course that its linked to, when no courses are supplied.
```
mjvkeulen, line 366: it's

src/vue/src/components/assets/DeadlineDeck.vue
```diff
@@ -1,40 +1,83 @@
...
+            v-if="$root.canGradeForSomeCourse"
+            class="text-grey float-right unselectable cursor-pointer"
         >
-            <b-link
-                :to="$root.assignmentRoute(d)"
-                tag="b-button"
+            <span
+                v-if="filterOwnGroups"
+                v-b-tooltip.hover
+                title="Showing to do items only for groups which you are a member of"
```
mjvkeulen, line 16: Drop only mid-sentence or move to front: `Only showing to do items for groups which you are a member of`

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -22,7 +22,9 @@
         <br/>
         <span v-if="deadline.deadline.date">
             <!-- Teacher deadline shows last submitted entry date  -->
-            <span v-if="deadline.stats.needs_marking + deadline.stats.unpublished > 0">
+            <span
+                v-if="unpublishedOrNeedsMarking"
```
mjvkeulen, line 26: Should single property tags be forced on a single line?

src/vue/src/components/assets/DeadlineDeck.vue
engelhamer: Matter of personal preference

PR #955, opened 2020-05-11
Stricter server access

mjvkeulen:
## Description
Enforce use of a password in combination with an ssh-key server side.


Makefile
```diff
@@ -23,6 +17,8 @@ ifndef host
 host=staging
 endif
 
+ansible_play_default_flags = --ask-pass --ask-vault-pass
```
Larspolo, line 20: why remove the file detection? Will be really useful when we separate staging from production key

mjvkeulen: This check promotes keeping keys (including production-related keys) in plaintext, this should be discouraged.

Larspolo: Then "become" should now also be removed

engelhamer, line 20: > @Larspolo : become should be removed

Larspolo: Also sshpass needs to get installed in `make setup`

mjvkeulen: It is in the PR already

Larspolo: Totally missed that, mb

PR #959, opened 2020-05-21
Fixed an issue where assignments were not deleted upon course deletion

### LINKED ISSUE #931: Dangling assignments after course deletion
**Describe the bug**
Assignments are not deleted when a course is deleted.

**To Reproduce**
1. Create a new assignment
2. Go to the course page
3. Delete the course
4. Check the database to see that the assignment still exists

Especially for LTI assignments this bug is noticeable, since no new assignment can be created with the same LTI id.

**Expected behavior**
The assignment should be deleted once the last connected course is deleted.

Larspolo:
## Description
Fixed issue that created dangling assignments after course deletion. When course is deleted, assignment switched to most recent lti course, if there are no other lti courses, lti_id gets set to nothing. If no other courses are connected, check if user can delete assignment, if so, also delete assignment.

## Summary of changes
- Fixes eJourn-al/eJournal#931


Larspolo:
Created issue #965

PR #960, opened 2020-05-21
Ignore default LMS profile picture

### LINKED ISSUE #805: Ignore default LMS profile picture
Most LMS users have not set a custom profile picture. As a result, we often receive the default LMS profile picture URL via LTI. Since this URL is always the same, it would be nice to filter it and instead keep our own default picture (which looks better since it is in the same style as the rest of the site).

**User interface**
Our own default picture:
image

Canvas default picture:
image

Larspolo:
## Description
Ignore default LMS profile picture.

Only checks when entering eJournal through an LMS instance. Also migrates every existing default icon to the eJournal standard

## Summary of changes
- Fixes eJourn-al/eJournal#805
- Added first migration test


src/django/VLE/lti_launch.py
```diff
@@ -70,7 +70,8 @@ def get_user_lti(request):
     users = User.objects.filter(lti_id=lti_user_id)
     if users.exists():
         user = users.first()
-        if 'custom_user_image' in request:
+        if 'custom_user_image' in request and \
+           request['custom_user_image'] != 'https://canvas.instructure.com/images/messages/avatar-50.png':
```
engelhamer, line 74: Would it not make more sense to make this an instance variable?

Larspolo: done

src/django/VLE/models.py
```diff
@@ -32,6 +32,9 @@ class Instance(models.Model):
     name = models.TextField(
         default='eJournal'
     )
+    default_lms_profile_picture = models.TextField(
+        default='https://canvas.instructure.com/images/messages/avatar-50.png'
```
mjvkeulen, line 36: I would expect this default to come from our server configuration. So adding this as an env variable set at the hosts.yml level in our config folder.

src/django/test/test_lti_launch.py
```diff
@@ -614,6 +614,28 @@ def test_get_lti_params_from_jwt_key_Error(self):
...
             assert_msg='When missing keys, it should return a keyerror')
 
+    def test_get_lti_params_update_prifile_picture(self):
+        lti_launch(
+            request_body={
+                'user_id': self.teacher.lti_id,
+                'custom_username': self.teacher.username,
+                'custom_user_image': Instance.objects.get_or_create(pk=1)[0].default_lms_profile_picture,
+            },
+            response_value=1,
```
mjvkeulen, line 624: The 1 should make use of `LTI_STATES` enum

src/django/VLE/models.py
engelhamer: Agreed would be nice!

Makefile
```diff
@@ -103,7 +103,7 @@ setup-venv:
 		source ./venv/bin/activate && \
 		pip install -r requirements/$(requirements_file) && \
 		isort -rc src/django/ && \
-		ansible-playbook ./config/provision-local.yml --ask-become-pass --ask-vault-pass && \
```
mjvkeulen, line 106: How come this can be removed?

Larspolo: cuz sudo is not used in the local playbook

Issue #570, opened 2019-02-04
Dynamic filetype as field


Larspolo:
**Describe the solution you'd like**
As a teacher I would like to specify which files can be submitted on which fields.

engelhamer:
This would mean a single file field which can display e.g. images and PDFs in the browser as well.

PR #964, opened 2020-05-23
Added support for file feedback to entries

### LINKED ISSUE #939: File feedback attached to comment
Being able to submit (mutliple) files as feedback on an entry. This is most useful when wanted to share your annotated PDF.

PDF (or other file) as comment

Larspolo:
## Description
Added support for file feedback to entries

See in image: editing a comment, comment with mutliple files, comment with text & 1 file, new comment pending with 1 file
image

## Summary of changes
- Fixes eJourn-al/eJournal#939


src/django/VLE/views/comment.py
```diff
@@ -149,11 +163,19 @@ def partial_update(self, request, *args, **kwargs):
 
         comment.last_edited_by = request.user
 
-        if not text:
-            return response.bad_request('Comment cannot be empty.')
-
         comment.text = text
         comment.published = published or not request.user.has_permission('can_grade', assignment)
+        # Add new files
```
mjvkeulen, line 168: Two duplicate blocks right?

src/django/VLE/views/comment.py
```diff
@@ -88,6 +90,16 @@ def create(self, request):
         # By default a comment will be published, only users who can grade can delay publishing.
         published = published or not request.user.has_permission('can_grade', assignment)
         comment = factory.make_comment(entry, request.user, text, published)
+        # Add new files
+        for file_id in files:
+            file = FileContext.objects.get(pk=int(file_id))
+            if not comment.files.filter(pk=file.pk).exists():
+                comment.files.add(file)
+                file_handling.establish_file(request.user, file.access_id, comment=comment)
+        # Remove old files
```
mjvkeulen, line 99: This workflow throws an error: tab 1 create a comment with file don't send. tab 2 create a comment with file don't send. Send tab 1 comment. Send tab 2 comment -> FileContext not found error. I think this workflow is obscure and is allowed to crash. Do we want a more specific error?

src/django/VLE/views/comment.py
```diff
@@ -88,6 +90,16 @@ def create(self, request):
         # By default a comment will be published, only users who can grade can delay publishing.
         published = published or not request.user.has_permission('can_grade', assignment)
         comment = factory.make_comment(entry, request.user, text, published)
+        # Add new files
```
mjvkeulen, line 93: Each small 'logic' block here is no longer separated by horizontal space. Is this your (new?) preferred style or not intentional.

src/django/VLE/views/comment.py
```diff
@@ -149,11 +163,19 @@ def partial_update(self, request, *args, **kwargs):
...
         comment.text = text
         comment.published = published or not request.user.has_permission('can_grade', assignment)
+        # Add new files
+        for file_id in files:
+            file = FileContext.objects.get(pk=int(file_id))
+            if not comment.files.filter(pk=file.pk).exists():
+                comment.files.add(file)
+                file_handling.establish_file(request.user, file.access_id, comment=comment)
+        # Remove old files
+        for file in comment.files.all():
```
mjvkeulen, line 175: comment.files.exclude(pk__in=files).delete()?

src/django/VLE/views/comment.py
Larspolo: I get that, but that is for a different issue right? This is a problem for all things. Entries being posted like this, will lose there files as well

src/django/VLE/views/comment.py
Larspolo: not intentional. Think at first I added the empty lines, and later filled them with comments. Will add another mepty line

engelhamer:
Would be nice to display attached PDFs or images in the comment itself (based on file extension). Also:  should a button be added like 'Attachments' to not always show the (large) upload field for every comment?

engelhamer:
File display field should be done in #570, so that can then be used here too.

engelhamer:
I don't really like that the file upload field is so large and that the name of the uploaded file remains until you select a new one (this is more suitable for when an explicit 'upload' or 'send' button is used, like when posting an entry). How about something like this?

image

Would keep it way more compact. With download buttons straight away such that you can check _what_ you uploaded

engelhamer:
image

Button now opens file picker.

Issue #875, opened 2020-03-13
Improve file field


Larspolo:
**Description**
Users with outdated browsers are able to upload non image files to image fields. This leads to those files not being able to be displayed correctly.

A solution would be to check in the backed if the files are actually valid.

Further improvements to the file field can also be done to reduce the three different types (img/pdf/file) to 1 more inclusive and customizable field.

Also: automatically detect file type, and based on that file type, show a corresponding output, e.g. automatically show the image on .jpg files.

PR #968, opened 2020-06-03
Improved configurability and display of file upload fields

Larspolo:
## Description
Allowed for easier customization and display in file field. Also allows for easy expansion into also supporting different file view methods (video's for example)

Automatic file extension detection:
image

More combustibility for teacher with supported prefixes:
image

Indication in allowed upload file types:
image

## Summary of changes
- Fixes eJourn-al/eJournal#570
- Fixes eJourn-al/eJournal#875


src/django/VLE/migrations/0044_dynamic_filename.py
```diff
@@ -0,0 +1,33 @@
+
+import logging
+
+from django.db import migrations, models
+
+
+def convert_to_dynamic_file_types(apps, schema_editor):
+    logger = logging.getLogger(__name__)
```
mjvkeulen, line 8: Unused

src/django/VLE/validators.py
```diff
@@ -82,3 +82,8 @@ def validate_entry_content(data, field):
             datetime.strptime(data, '%Y-%m-%dT%H:%M:%S')
         except ValueError as e:
             raise ValidationError(str(e))
+
+    if field.type == Field.FILE:
+        if field.options:
+            validator = FileExtensionValidator(field.options.split(', '))
```
mjvkeulen, line 88: Ah good use of a build-in!

src/vue/src/components/assets/file_handling/FileDisplay.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+    <image-file-display
+        v-if="type == 'img'"
+        :file="file"
+    />
+    <file-download-button
+        v-else-if="type == 'file'"
+        :file="file"
+    />
+
```
mjvkeulen, line 10: hspace

src/vue/src/components/assets/file_handling/FileDisplay.vue
```diff
@@ -0,0 +1,50 @@
...
+            required: true,
+        },
+    },
+    computed: {
+        type () {
+            if (!this.file) {
+                return null
+            }
+            const arr = this.file.file_name.split('.')
+            const extension = arr[arr.length - 1]
```
mjvkeulen, line 39: this.file.file_name.split('.').pop()

src/vue/src/components/template/TemplatePreview.vue
```diff
@@ -25,15 +25,6 @@
...
-                :acceptedFiletype="'image/*'"
-                :maxSizeBytes="$root.maxFileSizeBytes"
-                :autoUpload="false"
-                :aID="$route.params.aID"
-                :nID="'1'"
-                class="input-disabled"
-            />
             <file-upload-input
                 v-else-if="field.type == 'f'"
                 :acceptedFiletype="'*/*'"
```
mjvkeulen, line 30: This requires configuration here as well right?

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -130,6 +130,20 @@
...
+                                <b-select
+                                    v-model="field.fileExtensions"
+                                    :options="fileExtensions"
+                                    placeholder="Custom"
+                                    class="theme-select multi-form mr-2"
+                                    @change="field.fileExtensions === ' ' ? null : field.options = field.fileExtensions"
+                                />
+                                <b-input
+                                    v-model="field.options"
+                                    placeholder="Allowed extensions"
```
mjvkeulen, line 143: Possibly include a description of the ',' separated format?

src/django/VLE/utils/error_handling.py
```diff
@@ -71,7 +71,7 @@ def process_exception(self, request, exception):
         elif isinstance(exception, ObjectDoesNotExist):
             return response.not_found('{0} does not exist.'.format(str(exception).split()[0]), exception=exception)
         elif isinstance(exception, ValidationError):
-            return response.validation_error(str(exception), exception=exception)
+            return response.validation_error(exception.messages[0], exception=exception)
```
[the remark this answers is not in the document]
Larspolo: The response previously would poorly handle an array response with just a string representation of the array. This is now changed to only displaying one error at the time, but actually displaying it in a correct format

src/vue/src/components/template/TemplatePreview.vue
Larspolo: It is readonly, therefor I left it like this. Reduces complexity of the component

mjvkeulen: Right, we don't allow upload here anyway, so what file types we accept does not matter.

mjvkeulen:
Apart from not being able to select certain file types, how is it communicated to the students what files types are allowed?
E.g. I have an image which is not supported, how can I find out what types are allowed apart from trial and error?

src/vue/src/components/template/Field.vue
```diff
@@ -0,0 +1,204 @@
+<template>
```
engelhamer, line 1: In cases like these (with only 1 usage of this code) I am not really in favor of making a component, for it introduces more juggling around with data, less clear event handling and generally more steps to trace back where logic comes from.

Larspolo: In this case I would disagree. Splitting is done to reduce the long component size (350+ lines), and makes it easier to handle a specific field that needs to store information outside of the retrieved information from the API (in this instance it was the need for which option is selected, all, img, pdf, custom). The event handling happens only when fields are removed, all else will stay inside the component.

src/django/test/test_file_field.py
```diff
@@ -0,0 +1,57 @@
+import test.factory as factory
+from test.utils import api
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test import TestCase
+
+from VLE.models import Field
+
+BOUNDARY = 'BoUnDaRyStRiNg'
+MULTIPART_CONTENT = 'multipart/form-data; boundary=%s' % BOUNDARY
```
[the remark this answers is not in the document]
Larspolo: changed, but originally implemented by @mjvkeulen

engelhamer:
Reduced the vertical height of the cards by omitting the input field for predefined settings. Also made the text more verbose.

image


image

PR #970, opened 2020-06-12
Disabled issues for the repository

engelhamer:
## Description
Bug reports or suggestions for new features can be directed to support@ejournal.app instead.


Larspolo:
Issue templates should also be removed

PR #971, opened 2020-06-15
Changed the permission system to allow students viewing other journals when the respective permission is set

Larspolo:
## Description
can_view_all_journals is now also available for students, so students
can view eachothers journals when given this permission
supervisor is now determined by the can_grade permission, meaning
observers are no longer considered supervisors
you are still not able to have can_grade and can_have_journal in one
assignment (even with multiple courses)

Also added the "is_own" api call where you can check if the journal is
your own journal

## Summary of changes
- Fixes Z#1045

Time Till PR: 1:40


src/vue/src/views/Journal.vue
```diff
@@ -26,18 +29,25 @@ export default {
...
+        }
+    },
+    created () {
+        journalAPI.isOwnJournal(this.jID).then((isOwn) => {
+            this.isOwn = isOwn
+        })
+    },
     beforeRouteLeave (to, from, next) {
-        if (this.$hasPermission('can_have_journal') && !this.$refs['journal-student-ref'].discardChanges()) {
+        // TODO: change to authon in journal
```
mjvkeulen, line 43: Authon?

engelhamer: What about this?

Larspolo:
> The first ungraded assignment is selected when a student views a peer journal (even though they obviously cannot grade)
Fixed
> Students can see stats about unpublished grades in the overview
Fixed
> Observers (who _are_ faculty but not students) can now have a journal which is undesirable
Only if you give them the "can_have_journal"
> Students will now pass the `is_user_supervisor_of` check which does not make sense (as they are not a supervisor of one another)
They are not, only people that can grade are not supervisor

> I would like to postpone this feature to when peer feedback is introduced.
We wanted this early because someone requested it, right?

src/django/VLE/permissions.py
```diff
@@ -87,7 +87,7 @@ def is_user_supervisor_of(supervisor, user):
     can_view_course_users or where the supervisor is linked to the user through an assignment where the supervisor
```
engelhamer, line 87: Outdated documentation.

src/django/VLE/serializers.py
```diff
@@ -519,6 +521,22 @@ def get_usernames(self, journal):
 
         return None
 
+    def get_unpublished(self, journal):
+        # If annotated that it is allowed, immediatly get it, else check for can_view_all_journals
```
engelhamer, line 525: We don't check for `can_view_all_journals` anymore now right?

engelhamer: Kind of confusing sentence.

engelhamer:
The title of a journal bugs when loading it subsequent times (only first time is correct)

example5

engelhamer:
Once you reach your own journal the next/previous quick navigation  buttons disappear:

example6

engelhamer:
> > I would like to postpone this feature to when peer feedback is introduced.
> 
> We wanted this early because someone requested it, right?

Could also be implemented (and would require for the particular purpose) sharing a journal by URL. This alone would be easier I guess.

Larspolo:
[L] can_comment & can_view_assignment_journals zou nice zijn ook als assignment permission -> Zube issue van maken?

Larspolo:
Closing in favour of a full fletched system

Larspolo:
Closing in favour of a full fletched system

PR #973, opened 2020-06-20
Fixed an issue where it was possible to post an entry before a file upload finished

### LINKED ISSUE #889: Comment does not send after switching nodes
**Describe the bug**
Comment card:
- type a comment
- switch between nodes
- click "send" -> This will not send anything

- type an extra letter
- click "send -> now it works

Preferably the comment content will be kept in memory so teachers will be able to click around and use other nodes as well to give proper feedback

Larspolo:
## Description

Create/edit entry/comment is no longer possible while files are being uploaded

Also reworked the CommentCard to reduce complicated array logic

Also fixed #889

Also added comment file upload test


src/vue/src/components/assets/TextEditor.vue
```diff
@@ -273,10 +274,15 @@ export default {
...
                 formData.append('file', file)
-
+                this.uploading += 1
+                this.$emit('startedUploading')
                 auth.uploadFile('files', formData)
                     .then((response) => { success(response.data.download_url) })
                     .catch(() => { failure('File upload failed') })
+                    .finally(() => {
+                        this.uploading -= 1
+                        this.$emit('finishedUploading')
```
mjvkeulen, line 284: Would finished not require a uploading == 0 check? Possibly by watcher?

src/vue/src/components/entry/EntryNode.vue
```diff
@@ -33,6 +33,8 @@
                 :displayMode="false"
                 :nodeID="entryNode.nID"
                 :entryID="entryNode.entry.id"
+                @uploadingFile="uploadingFiles ++"
```
mjvkeulen, line 36: Enforce no space for unary operators?

src/vue/src/components/assets/TextEditor.vue
Larspolo: moved to outside component indeed, for every successful file you get a notification

PR #976, opened 2020-07-12
Added the ability to provide additional instructions (no-submission fields) to templates

### LINKED ISSUE #471: Info-only template field
The ability to add a field to an entry template where the user does not need to do anything at all. Just for additional instructions etc.

**User story**
The option 'Info only' or something better-named in the template editor, which can be used for placing extra text in a template (in the description of the field).

**User interface**
Just an extra option.

engelhamer:
## Description
Added a 'no-submission' field which can be used to e.g. provide additional instructions within a template:

image

## Summary of changes
- Fixes #471


src/django/test/test_entry.py
```diff
@@ -102,33 +102,52 @@ def test_create_entry(self):
...
         }
-        create_params = {
-            'journal_id': self.journal.pk,
-            'template_id': template.pk,
-            'content': [{
-                'id': f.pk,
-                'data': entries[f.type][0]
-            } for f in fields if f.type in entries]
+        invalid_content = {
+            # Field.VIDEO: 'https://www.test.com/watch?v=dQw4w9WgXcQ', TODO: This should fail.
```
mjvkeulen, line 121: Left over todo

engelhamer: It's not left over but intentionally left: this should be enabled after Z#1095.

src/vue/src/components/template/Field.vue
```diff
@@ -102,6 +103,13 @@
                         </b-button>
                     </div>
                 </div>
+                <small
+                    v-else-if="field.type == 'n'"
```
mjvkeulen, line 107: Let's use a strict equality operator here as well for consistency?

PR #978, opened 2020-07-19
Moved privacy policy to external source

engelhamer:
## Description
eJournal's privacy policy is now not included in this repository anymore. Custom privacy policies require simply changing the URL.

## Summary of changes
- Fixes Z#1091


engelhamer:
> Some text blocks in the PDF have a different color background:
> 
> image

Will fix, but the PDF is not part of this PR.

PR #979, opened 2020-07-23
Moved the assignment actions (such as exporting results) to sidebar

engelhamer:
## Description
To make room for further additions. Possibly these could be put into a kebab menu in a later stage (e.g. to better support iframes).

image

Also made the journal page more consistent with the other pages (removed card wrappers) and fixed some issues with Bootstrap shadow styles overriding the theme.

image


Larspolo:
Although I do agree with these changes, prev time I tried this, it got closed
https://github.com/eJourn-al/eJournal/pull/733

engelhamer:
> Although I do agree with these changes, prev time I tried this, it got closed
> https://github.com/eJourn-al/eJournal/pull/733
> 

There's now more buttons, and the course edit and journal page follow this approach already.

Larspolo:
In Journal, should it not also be an "Insight" and "Action" split, where the prev/next btns also are in the action section?

Larspolo:
image
Toast now has a different colour green

engelhamer:
> In Journal, should it not also be an "Insight" and "Action" split, where the prev/next btns also are in the action section?

I think next and prev are best tied to the details (next and prev journal).

PR #982, opened 2020-08-06
Fixed a couple small issues with the LTI grade passback system

Larspolo:
## Description
Fixed couple LTI grading bugs:
Fixed bug where creating new course wouldnt work
Grade now also gets send back when giving an unpublished grade
Submission date is now correctly set

## Summary of changes
- Fixes Z#1117
- Fixes Z#1085
- Fixes Z#1084
- Fixes Z#1143


src/vue/src/views/LtiLaunch.vue
```diff
@@ -252,8 +252,9 @@ export default {
                     this.linkableAssignments = assignments.slice()
                     for (let i = 0; i < this.linkableAssignments.length; i++) {
                         this.linkableAssignments[i].assignments = this.linkableAssignments[i].assignments.filter(
-                            assignment => assignment.active_lti_course === null
-                            || assignment.active_lti_course.cID !== this.page.cID)
+                            assignment => typeof assignment.active_lti_course === 'object'
+                            && assignment.active_lti_course !== null
```
engelhamer, line 256: If the `typeof` check indeed shows that it is an `assignment.active_lti_course` is an object, then it is never `null`?

engelhamer: This check is incorrect: it does not show assignments to link even though they exist (different course, or the same course with no active LTI link yet).

engelhamer:
```
assignment => !assignment.active_lti_course
|| assignment.active_lti_course.cID !== this.page.cID
```

Larspolo: > If the typeof check indeed shows that it is an assignment.active_lti_course is an object, then it is never null?
image
If only JavaScript would make sense.

> This check is incorrect: it does not show assignments to link even though they exist (different course, or the same course with no active LTI link yet).

Will check this now

engelhamer (approved):

engelhamer (changes requested):
Would like to see Z#1143 also fixed.

mjvkeulen (approved):

Issue #949, opened 2020-05-05
Use uppercase first letter for all component imports


engelhamer:
**Describe the solution you'd like**
The current style for component imports is something like

`import textEditor from '@/components/assets/TextEditor.vue'`

then

`components: { textEditor }`

I would prefer consistent capitalisation with file names everywhere for easier editing.

`import TextEditor from '@/components/assets/TextEditor.vue'`

Larspolo:
Please dont do this manually

mjvkeulen:
Linter rule would be welcome

PR #983, opened 2020-08-09
Improved front-end entry handling

engelhamer:
## Description
Previously, several different components were used to display, create or edit entries. They had a lot of logic in common, which was duplicated across the files. This has now been merged mostly into the `Entry` component (except for grading tasks).

Entry content is now an object with the corresponding field IDs as keys. This makes lookups easier and less prone to programming errors (which the `completeContent` stuff was). For example, an entry for a template with fields 'Title' (id 33), 'Summary' (id 34) and 'Rating' (id 35) would now be represented as:

```
content: {
    '33': 'AwesomeIT',
    '34': 'Lorem ipsum dolor ...',
    '35': 5,
}
```

## Summary of changes
- Fixes Z#1109
- Fixes Z#1110


src/vue/src/components/entry/EntryFields.vue
```diff
@@ -203,11 +200,11 @@ export default {
             return options
         },
         checkChanges () {
-            for (let i = 0; i < this.completeContent.length; i++) {
-                if (this.completeContent[i].data !== null && this.completeContent[i].data !== '') {
-                    return true
-                }
-            }
+            // for (let i = 0; i < this.content.length; i++) {
```
Larspolo, line 203: TODO

src/django/VLE/serializers.py
```diff
@@ -598,13 +598,19 @@ def get_author(self, entry):
...
     def get_template(self, entry):
         return TemplateSerializer(entry.template).data
 
     def get_content(self, entry):
-        return ContentSerializer(entry.content_set.all(), many=True).data
+        content_dict = {}
+
+        for content in entry.content_set.all():
+            # Only include the actual content (so e.g. text).
+            content_dict[content.field.id] = ContentSerializer(content).data['data']
```
Larspolo, line 611: This is the only use of a contentserializer, seems a bit too much. Would personally reduce it

src/django/test/test_entry.py
```diff
@@ -177,25 +167,19 @@ def test_required_and_optional(self):
...
-                'content': [{
-                    'id': field.pk,
-                    'data': 'filled' if field.required else empty_value
-                } for field in fields]
+                'content': {field.pk: 'filled' if field.required else empty_value for field in fields}
             }
             resp = api.update(self, 'entries', params=params.copy(), user=self.student)['entry']
-            assert len(resp['content']) == 3, 'Response should have emptied the optional fields, not deleted'
-            assert any([c['data'] is None for c in resp['content']]), \
+            assert len(resp['content'].keys()) == 3, 'Response should have emptied the optional fields, not deleted'
```
Larspolo, line 173: `.keys()` is not necessary:
```
>>> len({'a':6})
1
```

src/django/test/test_entry.py
```diff
@@ -47,10 +47,9 @@ def test_create_entry(self):
 
         # Check if students cannot update journals without required parts filled in
         create_params = self.valid_create_params.copy()
-        create_params['content'] = [{
-            'data': 'test title',
-            'id': self.valid_create_params['content'][0]['id']
-        }]
+        create_params['content'] = {
+            list(self.valid_create_params['content'].keys())[0]: 'only one field filled',
```
Larspolo, line 51: `.keys()` is not necessary:
```
>>> list({'a':6})
['a']
```

src/django/test/test_file_handling.py
```diff
@@ -170,7 +170,7 @@ def test_remove_unused_files_content(self):
         content_new_rt2 = api.post(
             self, 'files', params={'file': self.image, 'in_rich_text': True},
             user=self.student, content_type=MULTIPART_CONTENT, status=201)
-        patch['content'][1]['data'] = "<p>hello!<img src='{}' /><img src='{}' /></p>".format(
+        patch['content'][list(patch['content'].keys())[1]] = "<p>hello!<img src='{}' /><img src='{}' /></p>".format(
```
Larspolo, line 173: Dict order is preserved from Python3.6 onward. Before this PR we also supported Python3.5, do we want to ditch that?

src/vue/src/components/entry/Entry.vue
```diff
@@ -0,0 +1,223 @@
...
+    },
+    methods: {
+        saveChanges () {
+            if (this.checkRequiredFields()) {
+                entryAPI.update(this.node.entry.id, { content: this.newEntryContent },
+                    { customSuccessToast: 'Entry successfully updated.' })
+                    .then((entry) => {
+                        this.node.entry = entry
+                        this.edit = false
+                        this.requestInFlight = false
```
Larspolo, line 175: `this.requestInFlight = false` is never set to true (also indeleteEntry)

src/vue/src/components/journal/JournalStudent.vue
```diff
@@ -50,82 +50,65 @@
...
-                                :nodes="nodes"
-                                :bonusPoints="journal.bonus_points"
-                            />
-                        </div>
-                    </div>
+                        <option
+                            :value="null"
+                            disabled
+                        >
+                            New entry: select a template
```
Larspolo, line 63: I would prefer sth like this:
image

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -185,14 +185,14 @@
...
-import journalEndCard from '@/components/journal/JournalEndCard.vue'
-import progressNode from '@/components/entry/ProgressNode.vue'
+import EntryNonStudent from '@/components/entry/EntryNonStudent.vue'
+import Timeline from '@/components/timeline/Timeline.vue'
+import JournalDetails from '@/components/journal/JournalDetails.vue'
+import BreadCrumb from '@/components/assets/BreadCrumb.vue'
+import LoadWrapper from '@/components/loading/LoadWrapper.vue'
+import JournalStartCard from '@/components/journal/JournalStartCard.vue'
+import JournalEndCard from '@/components/journal/JournalEndCard.vue'
+import ProgressNode from '@/components/entry/ProgressNode.vue'
```
Larspolo, line 195: Capitalization is inconsistent throughout

Without capitalization at the start:
- Profile.vue
- JournalStudent.vue
- AssignmentOverview.vue
- Home.vue

With:
- Entry.vue
- TemplateEdit.vue
- JournalNonStudent.vue (now)

src/vue/src/views/Journal.vue
```diff
@@ -27,7 +27,7 @@ export default {
     },
     props: ['cID', 'aID', 'jID'],
     beforeRouteLeave (to, from, next) {
-        if (this.$hasPermission('can_have_journal') && !this.$refs['journal-student-ref'].discardChanges()) {
+        if (this.$hasPermission('can_have_journal') && !this.$refs['journal-student-ref'].safeToLeave()) {
```
Larspolo, line 30: Why is this renamed? And not renamed everywhere?

src/vue/src/components/journal/JournalNonStudent.vue
engelhamer: #949 mentions this change to capitalize the vue component imports (consistent with their actual filename). Just like some other changes we've made in the past I think it is no problem to do this on the go for files that are hit in a PR, if all imports in the file are addressed.

src/vue/src/views/Journal.vue
engelhamer: `discardChanges` implies that changes will be discarded once called. In reality this was just a check _if_ any unsaved changes were present and if so, the user would be _asked_ to discard them or not. This was a bit confusing. `safeToLeave` is easy to understand: if true, you can leave. Otherwise you can't.

engelhamer (commented):
Requested changes fixed

src/django/test/test_file_handling.py
engelhamer: Changed so dict order is not required anymore.

src/vue/src/views/Journal.vue
engelhamer: Also renamed for the other occurence now.

PR #986, opened 2020-08-19
Ensured the explicit naming of serializer fields

Larspolo:
## Description
Removed all __all__ fields in serializers 

## Summary of changes
- Fixes Z#1108


src/django/VLE/serializers.py
```diff
@@ -692,4 +688,4 @@ def get_download_url(self, file):
 class FieldSerializer(serializers.ModelSerializer):
     class Meta:
         model = Field
-        fields = '__all__'
+        fields = ('type', 'title', 'description', 'options', 'location', 'required', )
```
engelhamer, line 691: Don't we need an ID for fields?

engelhamer (changes requested):

mjvkeulen (approved):

PR #987, opened 2020-08-25
Added support for students requesting to continue a journal in another assignment

### LINKED ISSUE #217: Continue from old journal
Students that have to redo a course continue from the previous year. This will copy all entries from an older journal

mjvkeulen:
## Description
What did you change and why?

## Summary of changes
- Added the ability to create journal import requests (**JIR**s) Fixes #217. These can be created in the target journal view, by a student. Once created these requests become visible for the teachers of the relevant assignment with the new permission `can_manage_journal_import_requests`. The teacher can process the request in the target journal, if approved, the source journal and all entries will be copied to the target journal. Any relevant files or comments are also copied. There are 4 options when it comes to processing:
     - Approve including the existing grades, ungraded entries can be edited.
     - Approve excluding the existing grades, all entries can be edited.
     - Approve and lock the entries, all entries will be graded as `0`.
     - Simply decline the request.

    JIRs are kept as a history of earlier import requests and are only deleted when a pending JIR's source journal is deleted. JIR 
    statistics and ToDos show when relevant.
- Testing factories reworked to always present an accurate state of the underlying database model, no more additional or dangling testing instances. Each model can be initiated in isolation. Because each instantiated model will be linked in a sensible way according to our expected DB model, test will execute a lot swifter if instances upwards in the chain are reused. For example, creating entries in a loop without specifying the entry will generate a new journal, assignment, and course instance and users. Do not initiate formats in isolation, initiate a format via an assignment. Addition of tests for the factories themselves.
- Minor changes related to the database layer, the addition of some validation and stricter conditions as preparation for Z#1122
- Extention of the entry content validation rules, the addition of existing file checks.
- Added a script to drop any testing database generated by a Pytest thread in the case of an unsuccessful termination. Expects to be executed by the `postgres` user.
- Addition of helper function `equal_models` to compare dictionaries or model instances, including any nested lists or dictionaries. Makes use of DeepDiff library, can be passed top-level keys, e.g. `id` which will be directly ignored. Or can be passed regex to exclude nested paths from the comparison, e.g. something to catch `root['entry']['content']['<index>']['id']` value.
- Presetdb is reworked to only make use of our testing factories, longterm this should increase the accuracy of our tests. If anything is amiss in our local testing environment we now have to fix the factories to match the desired outcome. Fixes Z#1146
- Removed no longer needed VLE.factories, more can be removed as we move more existing tests to the new factories.
- Addition of a more scalable number badge display component Fixes Z#1135.

image

image


Larspolo (changes requested):
- Missing visual indication of pending import journal requests for student

- Would prefer using the journal component when getting the preview of the old journal (over the current iframe approach). Preferably with only the stuff that is necessary: timeline, content, grades (so no grading box, next/prev buttons etc)

- This dropdown can be non existing when there are no import requests
image

- Would prefer the button to be hidden if there are no open JIR's
image

- Badge keeps indicating TODO on JIR's even after approving all JIR's
image

- You can rerequest importing a journal after reloading the page (only does stuff in frontend though)

- A visual indication on which entries are imported would be nice

engelhamer:
> - Would prefer using the journal component when getting the preview of the old journal (over the current iframe approach). Preferably with only the stuff that is necessary: timeline, content, grades (so no grading box, next/prev buttons etc)

Maybe just omit the preview alltogether?


> - A visual indication on which entries are imported would be nice

I can add one now.

Also: instead of just the title of the assignment / journal in the teacher dropdown, it would be better to include the course and start/end year as well in the same dropdown line

src/django/VLE/factory.py
```diff
@@ -172,6 +171,7 @@ def make_lti_groups(course):
                 continue
 
 
+# TODO: Remove the chain of not used factories (or only used in tests)
```
engelhamer, line 174: When should this be done?

src/django/VLE/models.py
```diff
@@ -1604,7 +1646,7 @@ class Entry(CreateUpdateModel):
         null=True,
     )
 
-    last_edited = models.DateTimeField()
+    last_edited = models.DateTimeField(auto_now_add=True)
```
Larspolo, line 1649: This should not happen. This means that entry will change its last_edited state when the grade is updated (or author or whatever) while we use this field ONLY for checking if a user changed its content

src/django/VLE/models.py
```diff
@@ -1617,6 +1659,13 @@ class Entry(CreateUpdateModel):
         choices=TYPES,
     )
 
+    jir = models.ForeignKey(
```
Larspolo, line 1662: This is unclear to me, what does this field do? Maybe a comment or a different name

src/django/VLE/models.py
```diff
@@ -1628,9 +1677,6 @@ def is_graded(self):
 
     def save(self, *args, **kwargs):
         is_new = not self.pk
-        if is_new:
-            self.last_edited = timezone.now()
```
Larspolo, line 1632: Needs to be added back again

src/django/VLE/management/commands/preset_db.py
```diff
@@ -25,484 +22,256 @@ class Command(BaseCommand):
...
-                "abbr": "PAV1",
-                "author": self.users["Teacher"],
-                "students": [self.users[s] for s in ["Student", "Student2", "Student3", "Student4", "Student5",
-                                                     "TestUser"]],
-                "teachers": [self.users["Teacher"]],
-                "tas": [self.users["TA"]],
-                "start_date": faker.date("2018-09-01"),
-                "end_date": faker.date("2021-07-31"),
-                "student_group_names": ["Cobol", "Smalltalk"]
+                'instance': factory.Course(
```
engelhamer, line 96: Mixed single and double quotes (occurs often throughout all files in this PR).

src/django/VLE/models.py
```diff
@@ -1729,11 +1775,20 @@ def to_string(self, user=None):
...
+
+
 class Field(CreateUpdateModel):
     """Field.
 
     Defines the fields of an Template
     """
+    ALLOWED_URL_SCHEMES = ('http', 'https', 'ftp', 'ftps')
+    ALLOWED_DATE_FORMAT = '%Y-%m-%d'
+    ALLOWED_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S'
```
Larspolo, line 1791: LTI uses a different standard, so will have to be updated then :upside_down_face:

src/django/VLE/models.py
```diff
@@ -1793,17 +1848,18 @@ class Content(CreateUpdateModel):
...
-        on_delete=models.SET_NULL,
-        null=True
+        on_delete=models.CASCADE,
     )
     data = models.TextField(
         null=True
     )
 
     def save(self, *args, **kwargs):
+        if Content.objects.filter(field=self.field, entry=self.entry).count() > 1:
```
Larspolo, line 1860: Can this also be done as a unique together? Would like to use that instead if so

src/django/VLE/models.py
```diff
@@ -567,6 +567,8 @@ class Notification(CreateUpdateModel):
         NEW_COMMENT: 'entry',
     }
 
+    # QUESTION: WHy is this an integer field? If one is to debug a Notifaction, we have to look at the integer
```
engelhamer, line 570: Remove question or discuss elsewhere? This is not really the place.

src/django/VLE/models.py
```diff
@@ -1398,10 +1432,16 @@ def reset(self):
             Node.objects.create(type=preset_node.type, journal=self, preset=preset_node)
 
     def save(self, *args, **kwargs):
+        if not self.author_limit == self.UNLIMITED and self.authors.count() > self.author_limit:
+            raise ValidationError('Journal users exceed author limit.')
+        # QUESTION: Can a journal belong to a non group assignment with an author limit > 1?
```
engelhamer, line 1437: Is the author limit always 1 for non-group journals?

src/django/VLE/models.py
```diff
@@ -1426,6 +1466,8 @@ def to_string(self, user=None):
         return self.get_name()
 
 
+# You can only have node of a PresetNode for the same journal
+# TODO JIR: Unique together (journal, preset)
```
engelhamer, line 1470: To do left.

src/django/VLE/models.py
```diff
@@ -1885,3 +1941,120 @@ def save(self, *args, **kwargs):
...
+
+        if source_id:
+            source = Journal.objects.get(pk=source_id)
+            if not Entry.objects.filter(node__journal=source).exists():
+                raise ValidationError('You cannot create an import request whose source is empty.')
+
+        # QUESTION: Could even move the permission checks here, e.g. need to be author of both source and target journal
+        if target_id and source_id:
+            if source.assignment.pk == target.assignment.pk:
+                raise ValidationError('You cannot import a journal into itself.')
```
engelhamer, line 2043: `You cannot import a journal from within the same assignment.` would make more sense maybe?

engelhamer (changes requested):
49/88 viewed.

engelhamer:
I would like fewer journals with import requests in the preset db. This now seems oversampled (5 out of 7 whereas in reality even just 1 would be a rare occurence). Also the button for creating a request is too 'in your face' maybe? Only required once by a select few at most, but every time you visit a journal it's always there for everyone, permanently in sight.

engelhamer:
Added `IMPORTED` badge (like `LATE` badge) but noticed that even for imported entries `jir` is `null` (so it is not currently displayed)

image

mjvkeulen:
> Added `IMPORTED` badge (like `LATE` badge) but noticed that even for imported entries `jir` is `null` (so it is not currently displayed)
> 
> image

Would we not also expect an indication in the timeline itself? 

The serialization of an entry with corresponding JIR is tested here: `assert target_node['entry']['jir']['source']['assignment']['name'] == source_assignment.name` in `test_jir_import_result_via_serialization`. Are you sure the node data structure might not simply vary front end?

src/django/test/test_journal_import_request.py
```diff
@@ -0,0 +1,455 @@
...
+
+class JournalImportRequestTest(TestCase):
+    def test_journal_import_request_factory(self):
+        jir_c = JournalImportRequest.objects.count()
+        jir = factory.JournalImportRequest(author=factory.Student())
+
+        assert JournalImportRequest.objects.count() == jir_c + 1, 'A single journal import request is created'
+        assert jir.target.pk != jir.source.pk, 'A unique journal is generated for both the source and target'
+
+        # A jir generates only a single ap for both its source and target
```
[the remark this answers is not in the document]
mjvkeulen: The JIR factory will generate both a source and target journal unless provided, this is a test for the factory after all?

src/django/test/test_gradepassback.py
```diff
@@ -26,17 +26,17 @@ def setUp(self):
...
-            user=self.student, assignment=self.assignment, sourcedid='f6d552',
-            grade_url='https://uvadlo-tes.instructure.com/api/lti/v1/tools/267/grade_passback')
-        self.journal = factory.Journal(assignment=self.assignment)
-        self.journal.add_author(ap)
-        self.journal.save()
+        self.lti_journal = factory.LtiJournal(
+            assignment=self.assignment,
+            ap__user=self.student,
+            ap__sourcedid='f6d552',
+            ap__grade_url='https://uvadlo-tes.instructure.com/api/lti/v1/tools/267/grade_passback'
```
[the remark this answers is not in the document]
mjvkeulen: Agreed but not something introduced in this PR

src/django/test/test_entry.py
```diff
@@ -1,41 +1,309 @@
...
-            'journal_id': self.journal.pk,
+            'journal_id': self.group_journal.pk,
             'template_id': self.template.pk,
             'content': {}
         }
         fields = Field.objects.filter(template=self.template)
         self.valid_create_params['content'] = {field.id: 'test data' for field in fields}
 
+    def test_entry_factory(self):
+        course_c = Course.objects.count()
```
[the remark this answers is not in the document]
mjvkeulen: You don't think the variable names are quite clear given the structure? I am all for descriptive variable names but this was quite repetitive.

engelhamer:
> > Added `IMPORTED` badge (like `LATE` badge) but noticed that even for imported entries `jir` is `null` (so it is not currently displayed)
> > 
> > image
> 
> Would we not also expect an indication in the timeline itself? 
> 
> The serialization of an entry with corresponding JIR is tested here: `assert target_node['entry']['jir']['source']['assignment']['name'] == source_assignment.name` in `test_jir_import_result_via_serialization`. Are you sure the node data structure might not simply vary front end?

`jir` was `null` here although the entry was imported. I haven't yet traced why.

There is no need for an indication in the timeline itself in my opinion, maybe in some other PR where progress colours or the timeline in general are also addressed?

src/django/VLE/migrations/0053_entry_last_edited_auto_now_add.py
```diff
@@ -0,0 +1,16 @@
...
+        ('VLE', '0052_content_field_cascade'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='entry',
+            name='last_edited',
+            field=models.DateTimeField(auto_now_add=True),
+        ),
+    ]
```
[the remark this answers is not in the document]
mjvkeulen: I opted for separate as these changes have no connection to journal import requests

mjvkeulen:
> * Would prefer the button to be hidden if there are no open JIR's

It is currently conditional to `v-if="journal.import_requests"`, guess thats broken then?

mjvkeulen:
> * Would prefer using the journal component when getting the preview of the old journal (over the current iframe approach). Preferably with only the stuff that is necessary: timeline, content, grades (so no grading box, next/prev buttons etc)
> *
The components you are referring to are not modular

mjvkeulen:
> * This dropdown can be non existing when there are no import requests

I don't quite follow, the manage import request button should not show if there are no import requests, and the modal auto closes if there are no more import requests to manage?

engelhamer:
> > * Would prefer the button to be hidden if there are no open JIR's
> 
> It is currently conditional to `v-if="journal.import_requests"`, guess thats broken then?

Empty list is truthy. Check on length larger than 0.

src/django/VLE/models.py
mjvkeulen: It links to the JIR which approved its import. Could be `journal_import_request` but its quite long and `jir` is used quite often as shorthand. I think you will get used to it before you know it?

engelhamer:
> > * Would prefer using the journal component when getting the preview of the old journal (over the current iframe approach). Preferably with only the stuff that is necessary: timeline, content, grades (so no grading box, next/prev buttons etc)
> > *
> The components you are referring to are not modular

Omit the preview. Teacher can open journal (probably force new tab) by clicking on the card.

src/django/VLE/models.py
mjvkeulen: Fair enough, will revert.

src/django/VLE/management/commands/preset_db.py
mjvkeulen: The double quotes are leftovers from the previous preset_db, I should not be mixing quotes. Please point them out in the PR.

src/django/VLE/models.py
mjvkeulen: Since this is not live yet, I propose we change this more in line with the other choice fields.

src/django/VLE/models.py
mjvkeulen: I think so, would be another validation rule. @Larspolo

src/django/VLE/models.py
Larspolo: Cool, that's fair. Keep it then.

mjvkeulen:
> > > * Would prefer the button to be hidden if there are no open JIR's
> > 
> > 
> > It is currently conditional to `v-if="journal.import_requests"`, guess thats broken then?
> 
> Empty array is truthy. Check on length larger than 0?

I believe the `import_requests` are simply an int, the real JIRs are fetched once the manager is opened to speed up the assignment overview page. Not quite sure why the check wouldn't work, will check properly tomorrow.

src/vue/src/components/entry/Entry.vue
```diff
@@ -95,10 +95,18 @@
...
                         v-if="node.due_date
                             && new Date(node.due_date) < new Date(node.entry.last_edited)"
+                        v-b-tooltip:hover="'This entry was submitted after the due date'"
                         class="late-submission-badge"
                     >
                         LATE
                     </b-badge>
+                    <b-badge
+                        v-if="node.entry.jir"
+                        v-b-tooltip:hover="'This entry was imported from another journal'"
```
mjvkeulen, line 105: Would we not expect a bit more info? For example where the import came from?

src/django/VLE/serializers.py
```diff
@@ -614,12 +628,14 @@ def get_template(self, entry):
...
 
-        for content in entry.content_set.all():
+        for content in entry.content_set.all().order_by('field__location'):
             # Only include the actual content (so e.g. text).
             if content.field.type == Field.FILE:
                 try:
                     content_dict[content.field.id] = FileSerializer(FileContext.objects.get(pk=content.data)).data
                 except FileContext.DoesNotExist:
+                    capture_message(
+                        f'FILE content {content.pk} refers to unknown file in data: {content.data}', level='error')
```
Larspolo, line 638: Aperantly I started using this notation in the Notification PR as well. However, this is not Python 3.5. This will drop support for that version. Do we want that?

src/django/VLE/serializers.py
```diff
@@ -644,6 +660,16 @@ def get_comments(self, entry):
             return None
         return CommentSerializer(Comment.objects.filter(entry=entry), many=True).data
 
+    def get_jir(self, entry):
+        if entry.jir:
+            return {
+                'source': {
+                    'assignment': AssignmentSerializer(entry.jir.source.assignment, context=self.context).data,
```
Larspolo, line 667: `AssignmentSerializer` shouldnt be necessary here. This massively increased the data needed. At least use SmallAssignmentSerializer, or just what is needed

src/django/VLE/serializers.py
```diff
@@ -644,6 +660,16 @@ def get_comments(self, entry):
...
+                },
+                'processor': UserSerializer(entry.jir.processor, context=self.context).data
```
Larspolo, line 669: Also here, is all of this necessary? Doesnt even seem to be used anywhere

src/django/VLE/serializers.py
```diff
@@ -695,3 +721,36 @@ class FieldSerializer(serializers.ModelSerializer):
...
+    def get_processor(self, jir):
+        return UserSerializer(jir.processor, context=self.context).data
+
+    def get_source(self, jir):
+        if jir.source is None:
+            return 'Source journal no longer exists.'
+
+        return {
+            'journal': JournalSerializer(jir.source, context=self.context).data,
+            'assignment': AssignmentSerializer(jir.source.assignment, context=self.context).data,
```
Larspolo, line 749: This is an expensive one, here again Small is preferred. Dont think we need the full one.

src/django/VLE/utils/file_handling.py
```diff
@@ -123,15 +123,23 @@ def establish_file(author, identifier, course=None, assignment=None, journal=Non
...
 def get_files_from_rich_text(rich_text):
     if rich_text is None or len(rich_text) < 128:
-        return []
-    re_access_ids = re.compile(r'\/files\/[0-9]+\?access_id=([a-zA-Z0-9]+)')
-    return VLE.models.FileContext.objects.filter(access_id__in=re.findall(re_access_ids, rich_text), is_temp=True)
+        return VLE.models.FileContext.objects.none()
+    return VLE.models.FileContext.objects.filter(access_id__in=get_access_ids_from_rich_text(rich_text))
+
+
+def get_temp_files_from_rich_text(rich_text):
```
Larspolo, line 137: This rename is still needed in `views/entry.py`

src/django/VLE/utils/import_utils.py
```diff
@@ -0,0 +1,247 @@
...
+          grade should be set.
+        - jir (:model:`VLE.JournalImportRequest`): JIR instance triggering the import action.
+
+    Returns:
+        The copied entry.
+    '''
+    if jir is None and grade_author is None:
+        raise VLEProgrammingError('A grade author needs to be specified either via a JIR or directly.')
+
+    grade_author = grade_author if grade_author else jir.processor
```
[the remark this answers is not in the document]
Larspolo: For jir yes, but maybe not for other uses as well (which we might want later in life).

You did forget a check is jir is None or not. I would personally change this to:

```
if jir and not grade_author:
    grade_author = jir.processor
```

src/django/VLE/utils/import_utils.py
```diff
@@ -0,0 +1,247 @@
...
+
+    for field in Field.objects.filter(template=source_template_id):
+        field.pk = None
+        field.template = template
+        field.save()
+
+    return template
+
+
+def copy_entry(entry, node=None, grade=None, vle_coupling=None):
```
Larspolo, line 212: Whats the diff between `copy_*` and `import_*`?

Larspolo (commented):
17/90 :upside_down_face:

src/django/VLE/models.py
```diff
@@ -1330,6 +1358,12 @@ def grade(self):
     def unpublished(self):
         return self.node_set.filter(entry__grade__published=False).count()
 
+    @computed(models.IntegerField(null=True), depends=[
+        ['import_request_targets', ['target']],
```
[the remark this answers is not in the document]
mjvkeulen: This is tested here:
```        
jir2.state = JournalImportRequest.DECLINED
jir2.save()
journal.refresh_from_db()
assert journal.import_requests == 0, 'Import requests should update on the state change of a JIR'
```

What went wrong there?

src/django/VLE/serializers.py
```diff
@@ -361,6 +363,12 @@ def get_stats(self, assignment):
                 'needs_marking_own_groups': own_group.aggregate(Sum('needs_marking'))['needs_marking__sum'] or 0,
                 'unpublished_own_groups': own_group.aggregate(Sum('unpublished'))['unpublished__sum'] or 0,
             })
+        if self.context['user'].has_permission('can_manage_journal_import_requests', assignment):
+            own_group = journal_set.filter(authors__user__in=own_group_users)
+            stats.update({
+                'import_requests': journal_set.aggregate(Sum('import_requests'))['import_requests__sum'] or 0,
```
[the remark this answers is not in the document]
mjvkeulen: This aggregates the computed property: `import_requests`, which already filters based on `PENDING` state.

src/django/VLE/serializers.py
```diff
@@ -446,13 +454,18 @@ def get_templates(self, assignment):
...
     class Meta:
         model = Assignment
         fields = (
             'id', 'name', 'is_group_assignment', 'is_published', 'points_possible', 'unlock_date', 'due_date',
-            'lock_date', 'deadline', 'journal', 'stats', 'course', 'active_lti_course')
+            'lock_date', 'deadline', 'journal', 'stats', 'course', 'journal_import_requests', 'active_lti_course')
         read_only_fields = ('id', )
 
+    def get_journal_import_requests(self, assignment):
+        return JournalImportRequest.objects.filter(target__assignment=assignment).count()
```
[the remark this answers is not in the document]
mjvkeulen: Good catch will update to include the pending filter.

src/django/VLE/models.py
```diff
@@ -1729,11 +1775,20 @@ def to_string(self, user=None):
         return "Template"
 
 
+@receiver(models.signals.pre_delete, sender=Template)
+def delete_pending_jirs_on_source_deletion(sender, instance, **kwargs):
+    if Content.objects.filter(field__template=instance).exists():
+        raise VLEProgrammingError('Content still exists which depends on a template being deleted')
```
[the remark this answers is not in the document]
mjvkeulen: It's quite clear to me, what do you propose?

src/django/VLE/serializers.py
mjvkeulen: Agreed

src/django/VLE/serializers.py
mjvkeulen: I believe displaying frontend who approved the import request is desired. Could be stripped into yet another user serializer or custom dict, but I actually don't like those options as it spreads user serialization logic even more.

src/django/VLE/utils/import_utils.py
mjvkeulen: These functions have no side effects besides creating a copy of the passed instance. E.g. the `content_import` also duplicates all nested FCs.

mjvkeulen:
```
For jir yes, but maybe not for other uses as well (which we might want later in life).

You did forget a check is jir is None or not. I would personally change this to:

if jir and not grade_author:
    grade_author = jir.processor
```

There is quite a big check right above this line?

mjvkeulen:
> For jir yes, but maybe not for other uses as well (which we might want later in life).
> 
> You did forget a check is jir is None or not. I would personally change this to:

>if jir and not grade_author:
>    grade_author = jir.processor

There is quite a big check right above this line?

src/django/VLE/models.py
mjvkeulen: This was replaced by `auto_now_add=True` on the `last_edited` field. Which according to the [docs](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.DateField.auto_now_add): "_Automatically set the field to now when the object is first created_". How does this line differ?

src/django/VLE/models.py
mjvkeulen: Typical, at least we have a central place now.

src/django/VLE/models.py
```diff
@@ -1885,3 +1941,120 @@ def save(self, *args, **kwargs):
...
+            target = Journal.objects.get(pk=target_id)
+            if target.assignment.lock_date and target.assignment.lock_date < timezone.now():
+                raise ValidationError('You are not allowed to create an import request for a locked assignment.')
+
+        if source_id:
+            source = Journal.objects.get(pk=source_id)
+            if not Entry.objects.filter(node__journal=source).exists():
+                raise ValidationError('You cannot create an import request whose source is empty.')
+
+        # QUESTION: Could even move the permission checks here, e.g. need to be author of both source and target journal
```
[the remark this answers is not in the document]
mjvkeulen: Exactly, this would provide a (much needed) central place. 

- It would remove duplicate validation logic if the model is used in multiple view/http methods/utility functions etc.
- Enforce the same validation rules for our testing data.
- Some more advantages are outlined in Z#1139.

A problem with the save method is that the model could be saved for the first time, so some fields are not as easily accessed, hence the if statements. But this approach works without an `is_new` check, so the same logic both before and after the first save is called.

I am still looking for an optimal solution, and simply place this comment here to highlight that this approach is different from what we do usually, but it comes with the earlier outlined advantages.

src/django/VLE/serializers.py
mjvkeulen: I would love that, and even propose to move to 3.8 so we get access to the walrus operator (or higher).

src/django/VLE/validators.py
```diff
@@ -52,12 +54,22 @@ def validate_entry_content(content, field):
     """Validates the given data based on its field type, any validation error will be thrown."""
     if field.required and not (content or content == ''):
         raise VLEMissingRequiredField(field)
+    # QUESTION: Is '' really valid data for all fields?
```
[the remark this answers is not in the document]
mjvkeulen: So if the field is not required, and content is an empty string, we do not raise a validation error for *all* field types. 
So is an empty string truly ok for all field types?

src/django/VLE/views/journal.py
```diff
@@ -358,6 +358,7 @@ def leave(self, request, pk):
         journal.remove_author(author)
         if journal.authors.count() == 0:
             journal.reset()
+        JournalImportRequest.objects.filter(target=journal, state=JournalImportRequest.PENDING).delete()
```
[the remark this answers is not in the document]
mjvkeulen: Interesting topic, I extended the logic to the journal reset.

If one of your teammates made an import request (does not have to be to a personal journal, could for example be a duo project), would you really expect that import request to disappear when that person leaves the journal?

I would argue that the only non-ambiguous case is the journal reset. What do we think?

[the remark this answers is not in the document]
mjvkeulen: I also didn't think about the group journal `kick` functionality. Shall I move your proposal there?

src/vue/src/store/modules/preferences.js
```diff
@@ -168,6 +179,9 @@ export default {
             sortBy: 'name',
             filterOwnGroups: true,
         },
+        // TODO JIR move to DB preferences
+        journalImportRequestButtonSetting: 'AIG',
+        dismissedJIRs: [],
```
Larspolo, line 184: This is not really a preference. Would keep this in the component itself

src/vue/src/components/journal/JournalImportModal.vue
```diff
@@ -0,0 +1,150 @@
...
+                        Import journal
+                    </b-button>
+                </div>
+
+                <div v-else>
+                    <h4 class="theme-h4">
+                        No importable journals available.
+                    </h4>
+                    <hr class="m-0 mb-1"/>
+                    You can only import journal you own journals, and you cannot import the same journal into itself.
```
Larspolo, line 66: @engelhamer mentioned a better way to say this

src/django/VLE/serializers.py
Larspolo: May be desired, but its currently not used, suggest removal for now?

src/django/VLE/views/entry.py
```diff
@@ -59,11 +59,11 @@ def create(self, request):
 
         entry_utils.check_fields(template, content_dict)
 
-        # Node specific entry
+        # Deadline entry
         if node_id:
             node = Node.objects.get(pk=node_id, journal=journal)
             entry = entry_utils.add_entry_to_node(node, template, request.user)
-        # Template specific entry
+        # Unlimited entry
```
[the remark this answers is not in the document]
mjvkeulen: Could you make an issue regarding the naming scheme of deadline and unlimited entries?

src/django/VLE/views/journal_import_request.py
```diff
@@ -0,0 +1,94 @@
...
+
+        serializer = JournalImportRequestSerializer(
+            journal_target.import_request_targets.filter(state=JournalImportRequest.PENDING),
+            context={'user': request.user},
+            many=True
+        )
+
+        return response.success({'journal_import_requests': serializer.data})
+
+    def create(self, request):
```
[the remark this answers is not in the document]
mjvkeulen: I don't know of any missing permission checks, and tests exist? You are confused because the checks you are looking for can be found in the save of the model?

src/django/test/test_journal_import_request.py
```diff
@@ -0,0 +1,455 @@
...
+        assert len(resp) == 2, 'All JIRs for the given target are serialized'
+        assert resp[0]['id'] == jir.pk and resp[1]['id'] == jir2.pk, 'The correct JIRs are serialized'
+        assert resp[0]['source']['journal']['id'] == jir.source.pk \
+            and resp[0]['target']['journal']['id'] == jir.target.pk, \
+            'The correct source and target journal are serialized'
+        assert resp[0]['source']['journal']['import_requests'] == 0, \
+            'JIR import_requests (count) are only serialized for the target journal'
+        assert resp[0]['target']['journal']['import_requests'] == 2
+
+    def test_create_jir(self):
```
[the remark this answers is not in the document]
mjvkeulen: I don't know of any missing permission checks, and tests exist? You are confused because the checks you are looking for can be found in the save of the model?

src/vue/src/store/modules/preferences.js
mjvkeulen: This is as much a preference as `comment_button_setting` and `grade_button_setting`, I don't quite follow?

src/django/VLE/views/journal_import_request.py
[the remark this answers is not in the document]
Larspolo: Figured you needed some `can_view` persmission checks. Now if you cant view an assignment it will respond with a 404 iso the (for me expected) 403.

Also 
```
        # You cannot create a JIR for an unpublished assignment
        target_assignment.is_published = False
        target_assignment.save()
        api.create(self, 'journal_import_request', params=data, user=student, status=400)
        target_assignment.is_published = True
        target_assignment.save()
```
Currently fails.

src/django/test/test_journal_import_request.py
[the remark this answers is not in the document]
Larspolo:
```
        # You cannot create a JIR for an unpublished assignment
        target_assignment.is_published = False
        target_assignment.save()
        api.create(self, 'journal_import_request', params=data, user=student, status=400)
        target_assignment.is_published = True
        target_assignment.save()
```
`can_view` permission would be nice

src/django/test/test_journal_import_request.py
```diff
@@ -0,0 +1,455 @@
...
+        assert jir.target.authors.get(user=student.pk), 'Deep syntax works for source/target logic'
+
+        jir2 = factory.JournalImportRequest(target=jir.target)
+        assert jir.target.pk == jir2.target.pk, 'Directly assigning journal works'
+
+        course = factory.Course()
+        jir = factory.JournalImportRequest(source__assignment__courses=[course])
+        assert jir.source.assignment.courses.first().pk == course.pk
+
+        # # TODO JIR: Triggers unique constraint of AP (assignment, user) but why..
```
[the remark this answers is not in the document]
mjvkeulen: TODO removed, and unique constraint fixed

src/django/VLE/models.py
engelhamer: I propose we leave as is.

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: Lars advocates **not** implementing it here right?

src/django/VLE/views/journal.py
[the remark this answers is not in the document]
engelhamer: Well, in the current setup any outstanding request is always deleted once _some_ member leaves the journal right? This is not very logical, instead it would be better to remove one if a _specific_ member (the jir author) left or got kicked indeed.

src/django/test/test_entry.py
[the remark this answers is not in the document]
engelhamer: Well in some instances I see `a_c` and the like. That's just not very clear (and better of a bit longer but easier to understand).

src/vue/src/components/entry/Entry.vue
engelhamer: Not required in my opinion: the content is what's most important?

engelhamer: Could be added later on if needed.

src/django/VLE/validators.py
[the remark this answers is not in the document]
engelhamer: Are you talking about line 56 or 58?

src/django/test/factory/entry.py
```diff
@@ -1,63 +1,166 @@
...
+    else:
+        if self.node.journal.assignment.format.template_set.exists():
+            self.template = random.choice(self.node.journal.assignment.format.template_set.all())
+        else:
+            self.template = test.factory.TextTemplate(format=self.node.journal.assignment.format)
+    self.save()
+
+
+class UnlimitedEntryFactory(factory.django.DjangoModelFactory):
+    '''
```
engelhamer, line 35: Again single quote docstrings.

src/django/test/factory/user.py
```diff
@@ -19,6 +19,22 @@ class Meta:
     first_name = factory.Faker('first_name')
     last_name = factory.Faker('last_name')
 
+    @factory.post_generation
+    def preferences(self, create, extracted, **kwargs):
+        '''
```
engelhamer, line 24: Again single quote...

src/django/test/test_journal.py
```diff
@@ -18,63 +18,230 @@ def setUp(self):
...
+        factory.JournalImportRequest(target=g_journal, author=student, state=JournalImportRequest.PENDING)
+        factory.JournalImportRequest(target=g_journal, author=student, state=JournalImportRequest.APPROVED_EXC_GRADES)
+        factory.JournalImportRequest(source=g_journal, author=student, state=JournalImportRequest.PENDING)
+        source_approved_jir = factory.JournalImportRequest(
+            source=g_journal, author=student, state=JournalImportRequest.APPROVED_EXC_GRADES)
+
+        g_journal.authors.remove(g_journal.authors.first())
+        g_journal.reset()
+
+        # TODO: Enable once fixed?
```
engelhamer, line 155: Still disabled?

src/django/test/test_node.py
```diff
@@ -3,14 +3,40 @@
...
+        assert progress_preset_node.target, 'Progress preset node holds a target'
+
+        assert journal.node_set.count() == 2, 'Exactly two nodes have been added to the journal'
+        assert journal.node_set.filter(preset=deadline_preset_node, type=VLE.models.Node.ENTRYDEADLINE).exists(), \
+            'An entry deadline node has been added to the journal'
+        assert journal.node_set.filter(preset=progress_preset_node, type=VLE.models.Node.PROGRESS).exists(), \
+            'A progress node has been added to the journal'
+
     def test_get(self):
+        # TODO: Some meaningful tests on what is returned?
```
engelhamer, line 39: Issue or remove.

src/django/test/test_notification.py
```diff
@@ -80,51 +82,53 @@ def test_get_supervisors_of(self):
 
     @override_settings(CELERY_TASK_ALWAYS_EAGER=True, CELERY_TASK_EAGER_PROPAGATES=True)
     def test_comment_notification(self):
-        entry = factory.Entry()
+        entry = factory.UnlimitedEntry()
+        # journal = entry.node.journal
```
engelhamer, line 86: Disabled.

src/django/test/test_notification.py
```diff
@@ -80,51 +82,53 @@ def test_get_supervisors_of(self):
...
+        new_notifactions = Notification.objects.all().exclude(pk__in=notifications_before)
+
+        # 1 comment notifaction is created
+        new_comment_notifaction = new_notifactions.get(type=Notification.NEW_COMMENT)
+        assert new_notifactions.count() == 1
+
+        assert new_comment_notifaction.user.pk == entry.author.pk
+        assert not new_comment_notifaction.sent
+
+        # TODO JIR
```
engelhamer, line 99: To do left.

src/django/test/utils/generic_utils.py
```diff
@@ -11,21 +21,101 @@ def to_dict(instance, ignore=[]):
...
 
     return data
 
 
-def equal_dicts(d1, d2, ignore=[]):
-    for k, v in d1.items():
-        if d2[k] != v:
-            return False
+def _keys_to_path_ignore_regex(ignore_keys=[]):
+    '''
```
engelhamer, line 35: Again double quote docstrings.

src/vue/src/components/assets/DeadlineDeck.vue
```diff
@@ -132,10 +132,18 @@ export default {
             const filterOwnGroupsCopy = this.filterOwnGroups
             function compareMarkingNeeded (a, b) {
                 if (filterOwnGroupsCopy) {
-                    return (b.stats.needs_marking_own_groups + b.stats.unpublished_own_groups)
-                    - (a.stats.needs_marking_own_groups + a.stats.unpublished_own_groups)
+                    return (
+                        b.stats.needs_marking_own_groups
+                        + b.stats.unpublished_own_groups
+                        + b.stats.import_requests_own_groups
```
engelhamer, line 138: Why is this included in marking needed sort here?

src/vue/src/components/assets/TodoCard.vue
```diff
@@ -53,11 +61,11 @@
 </template>
 
 <script>
-import numberBadge from '@/components/assets/NumberBadge.vue'
+import numberBadgesDisplay from '@/components/assets/NumberBadgesDisplay.vue'
```
engelhamer, line 64: Why was the display suffix added?

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
```diff
@@ -156,7 +156,7 @@ export default {
                 customErrorToast: 'Something is wrong with the uploaded file.',
             })
                 .then(() => {
-                    this.$emit('bonusPointsSuccesfullyUpdated', this.file.name)
+                    this.$emit('bonusPointssuccessfullyUpdated', this.file.name)
```
engelhamer, line 159: Not camelcase anymore.

src/vue/src/components/assignment/StatisticsCard.vue
```diff
@@ -4,32 +4,46 @@
         class="no-hover"
     >
         <div class="multi-form">
-            <b>Needs marking</b><number-badge
-                :leftNum="stats.needsMarking"
+            <b>Needs marking</b><number-badges-display
```
engelhamer, line 7: Move to new line.

src/vue/src/components/journal/JournalImportModal.vue
```diff
@@ -0,0 +1,150 @@
...
+        noEnforceFocus
+    >
+        <b-card class="no-hover">
+            <h2 class="theme-h2 multi-form">
+                Select a journal to import
+            </h2>
+
+            <p>
+                Please select a journal to import. Your request will be sent to your educator for approval. Once
+                approved, the entries of the selected journal will be added to your journal. The reuse of existing
```
engelhamer, line 17: `transfer` of existing grades?

src/vue/src/components/journal/JournalImportModal.vue
```diff
@@ -0,0 +1,150 @@
...
+                        Import journal
+                    </b-button>
+                </div>
+
+                <div v-else>
+                    <h4 class="theme-h4">
+                        No importable journals available.
+                    </h4>
+                    <hr class="m-0 mb-1"/>
+                    You can only import journal your own journals, and you cannot a journal into itself.
```
engelhamer, line 66: `cannot import a journal from within the same assignment` ?

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+
+                    <journal-card
+                        v-if="selectedJir"
+                        v-b-tooltip.hover="'Navigate to journal in full'"
+                        :journal="selectedJir.source.journal"
+                        :assignment="selectedJir.source.assignment"
+                        @click.native="openSelectedJirInNewWindow()"
+                    />
+
+                    <div
```
engelhamer, line 44: Remove preview alltogether for now.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+                        placeholder="Select An Assignment"
+                        class="multi-form"
+                    />
+
+                    <journal-card
+                        v-if="selectedJir"
+                        v-b-tooltip.hover="'Navigate to journal in full'"
+                        :journal="selectedJir.source.journal"
+                        :assignment="selectedJir.source.assignment"
+                        @click.native="openSelectedJirInNewWindow()"
```
engelhamer, line 41: This is very nice and sufficient info for teacher in my opinion.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+                    </b-button>
+
+                    <dropdown-button
+                        v-if="$hasPermission('can_grade')"
+                        :up="true"
+                        :selectedOption="$store.getters['preferences/journalImportRequestButtonSetting']"
+                        :options="{
+                            AIG: {
+                                text: 'Approve including grades',
+                                icon: 'paper-plane',
```
engelhamer, line 73: `check` instead of `paper-plane` would fit better? Also for those below.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+                                class: 'add-button',
+                            },
+                            AWGZ: {
+                                text: 'Approve with grades zeroed',
+                                icon: 'paper-plane',
+                                class: 'change-button',
+                            },
+                            DEC: {
+                                text: 'Decline',
+                                icon: 'ban',
```
engelhamer, line 88: `cross`?

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+                        class="float-right"
+                        :class="{ 'input-disabled': !selectedAssignment }"
+                        @change-option="$store.commit('preferences/SET_JOURNAL_IMPORT_REQUEST_BUTTON_SETTING', $event)"
+                        @click="handleJIR(selectedJir)"
+                    />
+                </div>
+
+                <div v-else>
+                    <h4 class="theme-h4">
+                        No outstanding journal import requests available.
```
engelhamer, line 101: Either remove `outstanding` or `available`. They mean the same.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+            jirs: [],
+            selectedAssignment: null,
+            preview: false,
+            loading: true,
+            jirAction: null,
+        }
+    },
+    computed: {
+        // QUESTION:
+        // How to simply provide jirs as options to the multiselect with a display for jir.source.assignment.name?
```
engelhamer, line 139: Display should also include course and year, e.g.:

`Colloquiumlogboek, Academische Vaardigheden Informatica 1 (2019-2020)`

Otherwise teacher does not know from which edition of a course (or which course at all) the journal is.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+    },
+    methods: {
+        jirToJournalUrl (jir) {
+            return `${window.location.protocol}//${window.location.host}/Home/`
+                + `Course/${jir.source.assignment.course.id}/`
+                + `Assignment/${jir.source.assignment.id}/`
+                + `Journal/${jir.source.journal.id}`
+        },
+        handleJIR (jir) {
+            // TODO JIR: Update timeline without required reload
```
engelhamer, line 171: Reloading nodes is not a problem in my opinion as this rarely occurs. Easier to emit and re-init journal.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -0,0 +1,213 @@
...
+            if (event.trigger !== 'event') {
+                this.$store.commit('preferences/ADD_DISMISSED_JIRS_TO_JOURNAL', Array.from(this.jirs, jir => jir.id))
+            }
+        },
+    },
+}
+</script>
+
+<style lang="sass">
+.resp-iframe-container
```
engelhamer, line 201: Remove when removing iframe.

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -146,7 +146,7 @@
                     </h3>
                     <div
                         v-if="$hasPermission('can_grade')"
-                        class="bonus-section grade-section full-width"
+                        class="bonus-section grade-section mt-1 full-width"
```
engelhamer, line 149: Prefer `mb-2` on all buttons.

src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -209,6 +225,7 @@ export default {
         JournalStartCard,
         JournalEndCard,
         ProgressNode,
+        journalImportRequestApprovalModal,
```
engelhamer, line 228: Capitalize.

src/vue/src/components/journal/JournalStudent.vue
```diff
@@ -142,6 +142,25 @@
                 />
             </b-card>
 
+            <h3 class="theme-h3">
```
engelhamer, line 145: Wrap this whole thing in a `v-if`: there should never be an 'actions' heading when no actions are available.

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -65,6 +65,7 @@ export default {
                 return 'calendar'
             }
         },
+        /* TODO JIR: Meaningfull way to display show that the entry was imported */
```
engelhamer, line 68: Not needed in my opinion (or create separate issue).

src/vue/src/store/modules/preferences.js
engelhamer: Dismissed requests? How are they preferences?

src/vue/src/store/modules/preferences.js
```diff
@@ -168,6 +179,9 @@ export default {
             sortBy: 'name',
             filterOwnGroups: true,
         },
+        // TODO JIR move to DB preferences
```
engelhamer, line 182: Still needs to be done.

src/vue/src/views/Assignment.vue
```diff
@@ -50,6 +50,9 @@
                     <option value="markingNeeded">
                         Sort by marking needed
                     </option>
+                    <option value="importRequests">
```
engelhamer, line 53: Only provide option _if_ there are any at all

src/vue/src/views/Assignment.vue
```diff
@@ -106,7 +109,7 @@
                         :endpoint="'assignments/' + $route.params.aID + '/add_bonus_points'"
                         :aID="$route.params.aID"
                         class="mt-2"
-                        @bonusPointsSuccesfullyUpdated="hideModal('bonusPointsModal'); init()"
+                        @bonusPointssuccessfullyUpdated="hideModal('bonusPointsModal'); init()"
```
engelhamer, line 112: Also inconsistent camelcase here.

src/vue/src/views/Assignment.vue
```diff
@@ -605,15 +608,18 @@ export default {
...
             for (let i = 0; i < filteredJournals.length; i++) {
                 needsMarking += filteredJournals[i].needs_marking
                 unpublished += filteredJournals[i].unpublished
                 points += filteredJournals[i].grade
+                importRequests += filteredJournals[i].import_requests
             }
             this.stats = {
                 needsMarking,
                 unpublished,
+                importRequests,
```
engelhamer, line 622: Only include if larger than 0. Not needed to show the whole time.

engelhamer (changes requested):

src/django/VLE/models.py
Larspolo: Aha missremembered what auto_now_add was

src/vue/src/store/modules/preferences.js
Larspolo: comment_button_setting grade_button_setting are both on how you like the buttons to be set

src/django/VLE/management/commands/preset_db.py
```diff
@@ -25,484 +22,256 @@ class Command(BaseCommand):
...
+                grade = random.choice([factory.Grade(grade=random.randint(0, 3)), None])
+                factory.UnlimitedEntry(node__journal=j, grade=grade, gen_content__from_file=True)
+
+                # Colloquium holds some preset nodes
+                if a.name == "Colloquium":
+                    deadline_nodes = list(j.node_set.filter(type=Node.ENTRYDEADLINE))
+                    n_deadline_entries = random.randint(0, j.node_set.filter(type=Node.ENTRYDEADLINE).count())
+                    for deadline_node in random.sample(deadline_nodes, n_deadline_entries):
+                        factory.PresetEntry(
+                            node__preset=deadline_node.preset, node__journal=j, gen_content__from_file=True)
```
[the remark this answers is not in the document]
mjvkeulen: Turned into an issue, just need consensus on the numbers.

src/django/VLE/models.py
```diff
@@ -1398,10 +1432,16 @@ def reset(self):
             Node.objects.create(type=preset_node.type, journal=self, preset=preset_node)
 
     def save(self, *args, **kwargs):
+        if not self.author_limit == self.UNLIMITED and self.authors.count() > self.author_limit:
+            raise ValidationError('Journal users exceed author limit.')
+        # QUESTION: Can a journal belong to a non group assignment with an author limit > 1?
+        # Otherwise this would form a reasonable validation rule
```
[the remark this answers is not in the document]
mjvkeulen: Added

src/django/VLE/models.py
mjvkeulen: See other comment

src/django/VLE/models.py
mjvkeulen: Added

src/django/VLE/factory.py
mjvkeulen: Z#1164

src/django/VLE/models.py
mjvkeulen: Since this was not live yet I updated it in place.

src/django/VLE/models.py
mjvkeulen: Added test and validation rule

src/django/test/test_entry.py
```diff
@@ -1,41 +1,309 @@
...
+            ValidationError,
+            validate_entry_content,
+            faker.date(pattern='NOT-Y%-ALLOWED-%m-FORMAT-%d'),
+            valid_datetime_content.field
+        )
+
+        valid_file_field_content = entry.content_set.filter(field__type=Field.FILE).first()
+        fc = FileContext.objects.get(content=valid_file_field_content)
+
+        # QUESTION: Now the data which is validated is no longer equal to the data of the actual content.
```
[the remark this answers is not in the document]
mjvkeulen: Likely to be addressed in Z#1122

src/vue/src/components/entry/Entry.vue
mjvkeulen: I strongly feel we should present the information on where the journal was imported from

src/django/VLE/utils/file_handling.py
mjvkeulen: Good catch, double check if correct now

src/django/VLE/views/journal_import_request.py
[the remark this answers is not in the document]
mjvkeulen: Added test and check, anything else?

src/django/test/factory/entry.py
mjvkeulen: We use single quotes when possible, quite the trap haha

src/django/test/test_journal.py
mjvkeulen: This is a newly added test revolving around code introduced by #359. Created dedicated issue Z#1150, "Floating entries on journal reset"

src/django/test/test_node.py
mjvkeulen: Z#1165

src/vue/src/views/Assignment.vue
mjvkeulen: Good subject, we need to move to kebab case for events anyway.

https://eslint.vuejs.org/rules/custom-event-name-casing.html

Currently, the rule is enforced for my editor by default but not for our env, need to check why there is a mismatch to begin with.

Z#1166

src/vue/src/views/Assignment.vue
mjvkeulen: Number badge is not shown if zero, filers should remain visible if zero. Or what do you mean?

Larspolo (approved):

src/vue/src/components/timeline/TimelineNodeCircle.vue
engelhamer: Remove.

src/vue/src/store/modules/preferences.js
engelhamer: Remove.

src/django/VLE/models.py
```diff
@@ -1885,3 +1941,120 @@ def save(self, *args, **kwargs):
...
+                state__in=JournalImportRequest.APPROVED_STATES, source=source)
+            if self.pk:
+                existing_import_qry = existing_import_qry.exclude(pk=self.pk)
+            if existing_import_qry.exists():
+                raise ValidationError('You cannot import the same journal multiple times.')
+
+        super(JournalImportRequest, self).save(*args, **kwargs)
+
+        if is_new:
+            # TODO JIR Create notifications depending on state
```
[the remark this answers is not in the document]
engelhamer: Remove.

src/django/test/factory/__init__.py
```diff
@@ -1,58 +1,100 @@
...
+    '''
+    if not create:
+        return
+
+    if isinstance(extracted, model):
+        setattr(instance, key, extracted)
+    elif not (kwargs or non_cond_kwargs or cond_kwargs) and default:
+        setattr(instance, key, default)
+    else:
+        # TODO Handle size more cleanly
```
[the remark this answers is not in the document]
engelhamer: Remove.

src/django/test/factory/content.py
```diff
@@ -0,0 +1,124 @@
...
+                filename = factory.Faker('file_name', category='image').generate()
+                # Use a preexisintg file on os
+                if 'from_file' in kwargs and not kwargs['from_file'] == '':
+                    from_file = _from_file_to_file(IMG_FILE_PATH)
+                    filename = factory.Faker('file_name', category='image', extension='png').generate()
+
+                # NOTE: Params are not available in post generation
+                # https://github.com/FactoryBoy/factory_boy/issues/544
+                for _ in range(kwargs['n_files'] if 'n_files' in kwargs else 1):
+                    # TODO deep syntax could be made available by using a dedicated class with its own related
```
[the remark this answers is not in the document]
engelhamer: Remove.

src/django/VLE/validators.py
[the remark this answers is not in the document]
engelhamer: Remove question

engelhamer (changes requested):
Still TODO and QUESTION left.

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
engelhamer: Still not removed

mjvkeulen:
JIR approval modal now does not select the correct assignment by default

src/django/VLE/utils/import_utils.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Args:
+        entry (:model:`VLE.entry`): Entry the copied grade should be attached to.
+        author (:model:`VLE.user`): Author of the new grade
+        action (str): Choice of (:model:`VLE.JournalImportRequest`).APPROVED_STATES
+    '''
+    if action not in JournalImportRequest.APPROVED_STATES:
+        raise VLEProgrammingError('Copy grade request based on unrecognized action')
+
+    if action == JournalImportRequest.APPROVED_INC_GRADES:
```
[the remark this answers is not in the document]
mjvkeulen: Updated and tested

src/django/VLE/utils/import_utils.py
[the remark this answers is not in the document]
mjvkeulen: If the grade author is specifically specified I would prefer to use that instead.

src/django/VLE/views/journal_import_request.py
```diff
@@ -0,0 +1,94 @@
...
+        jir.state = jir_action
+        jir.processor = request.user
+
+        source_entries = Entry.objects.none()
+        if not jir_action == jir.DECLINED:
+            source_entries = Entry.objects.filter(node__journal=jir.source)
+            if not source_entries.exists():
+                jir.state = jir.EMPTY_WHEN_PROCESSED
+
+        for source_entry in source_entries:
```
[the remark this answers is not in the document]
mjvkeulen: Just a way to skip the copy loop

src/django/VLE/serializers.py
mjvkeulen: Resolving as it's now used to display in the imported badge

src/django/test/test_import_utils.py
```diff
@@ -0,0 +1,208 @@
...
+    def test_copy_entry_and_node(self):
+        source_entry = factory.UnlimitedEntry()
+        target_journal = factory.Journal(entries__n=0)
+
+        copied_node_instance = import_utils.copy_node(source_entry.node, target_journal)
+        assert copied_node_instance.pk != source_entry.node.pk
+        assert equal_models(
+            source_entry.node,
+            copied_node_instance,
+            ignore_keys=['last_edited', 'update_date', 'id', 'creation_date', 'journal', 'entry']
```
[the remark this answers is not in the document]
mjvkeulen: Removed and test held

src/django/test/test_import_utils.py
```diff
@@ -0,0 +1,208 @@
...
+        ), 'Copied node is equal to the source node apart from the journal and entry'
+        assert copied_node_instance.journal.pk == target_journal.pk, \
+            'Node should be linked to the provided target journal'
+
+        copied_entry_instance = import_utils.copy_entry(source_entry, node=copied_node_instance)
+        assert copied_entry_instance.pk != source_entry
+        assert equal_models(
+            source_entry, copied_entry_instance, ignore_keys=['last_edited', 'update_date', 'id', 'creation_date'])
```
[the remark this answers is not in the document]
mjvkeulen: Fixed for both comment and entry

src/django/test/test_content.py
```diff
@@ -0,0 +1,125 @@
...
+from VLE.utils.error_handling import VLEProgrammingError
+
+
+class ContentTest(TestCase):
+    def test_temp_todo_remove(self):
+        journal = factory.Journal(entries__n=0, assignment__format__templates=[{'type': Field.RICH_TEXT}])
+        entry = factory.UnlimitedEntry(node__journal=journal)
+
+        for c in entry.content_set.all():
+            VLE.validators.validate_entry_content(c.data, c.field)
```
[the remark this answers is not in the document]
mjvkeulen: Was added in this PR, see `test_validate_entry_content`

src/django/VLE/serializers.py
```diff
@@ -614,12 +628,14 @@ def get_template(self, entry):
     def get_content(self, entry):
         content_dict = {}
 
-        for content in entry.content_set.all():
+        for content in entry.content_set.all().order_by('field__location'):
```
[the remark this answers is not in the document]
mjvkeulen: If the order is handled front end leave it there, removed.

src/django/test/factory/content.py
```diff
@@ -0,0 +1,124 @@
...
+            elif self.field.type == Field.URL:
+                self.data = factory.Faker('url').generate()
+            elif self.field.type == Field.DATE:
+                self.data = factory.Faker('time', pattern=Field.ALLOWED_DATE_FORMAT).generate()
+            elif self.field.type == Field.DATETIME:
+                self.data = factory.Faker('time', pattern=Field.ALLOWED_DATETIME_FORMAT).generate()
+            elif self.field.type == Field.SELECTION:
+                self.data = random.choice(json.loads(self.field.options))
+            elif self.field.type == Field.NO_SUBMISSION:
+                raise VLEProgrammingError('Content should never be generated for a no submission field')
```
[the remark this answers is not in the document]
mjvkeulen: @engelhamer ?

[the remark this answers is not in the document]
mjvkeulen: Engel confirmed that there should never be content for a no submission field

src/django/test/test_course.py
```diff
@@ -38,6 +39,24 @@ def test_rest(self):
                       create_status=403,
                       user=factory.Student())
 
+    def test_course_factory(self):
+        c_c = Course.objects.count()
```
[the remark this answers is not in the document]
mjvkeulen: I find the variables quite clear given the structure, I believe if the variable-length would be increased the code would become more unclear due to single line statements spanning multiple lines. `c_c` course count, `a_c` assignment count. These are also used in very obvious asserts with text.

src/vue/src/views/Assignment.vue
mjvkeulen: This conflicts with our policy of not hiding filters when the result would be zero.

src/vue/src/store/modules/preferences.js
mjvkeulen: And the dismissed requests determines whether the modal should auto open on journal show. If any jirs exists which were not processed yet, but the modal was closed by the user. Unless there are new jirs, the modal will not auto open again.

src/vue/src/components/assets/TodoCard.vue
mjvkeulen: As it is a display component not single (theme styled) number badge

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
mjvkeulen: See other comment, camelcase events are dangerous

src/vue/src/components/assignment/StatisticsCard.vue
mjvkeulen: It still (like the old line) an inline display

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
mjvkeulen: I referred to outstanding as pending JIRs, for the user the result is the same so I will remove it

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
mjvkeulen: Fixed by Lars

engelhamer (approved):

PR #992, opened 2020-08-27
Added support for teacher-initiated entries

engelhamer:
Added the possibility for teachers to post an entry to multiple student journals at once.

- Fixed Z#1034


Makefile
```diff
@@ -159,6 +159,14 @@ postgres-reset:
...
 	-c \"DROP DATABASE IF EXISTS $(postgres_db)\" \
 	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw0\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw1\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw2\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw3\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw4\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw5\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw6\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw7\" \
```
mjvkeulen, line 169: Can be removed, script exists which handles variable number of threads

Makefile
```diff
@@ -203,7 +211,14 @@ db-restore:
...
 	-c \"DROP DATABASE IF EXISTS $(postgres_db)\" \
-	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw0\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw1\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw2\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw3\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw4\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw5\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw6\" \
+	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)_gw7\" \
```
mjvkeulen, line 221: Ditto

src/django/VLE/migrations/0057_teacher_entry_setters.py
```diff
@@ -0,0 +1,29 @@
...
+    Role.objects.filter(name='Teacher').update(can_post_teacher_entries=True)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0056_teacher_entry'),
+    ]
+
+    operations = [
```
mjvkeulen, line 26: Are operations ordered, could these migrations be merged?

src/django/VLE/models.py
```diff
@@ -1709,6 +1716,37 @@ def to_string(self, user=None):
...
+    # are individually graded / grades passed back to the LMS from there.
+    # This allows editing and grade passback mechanics for students like usual.
+    grade = None
+    teacher_entry = None
+    vle_coupling = None
+
+    def save(self, *args, **kwargs):
+        self.grade = None
+        self.teacher_entry = None
+        self.vle_coupling = Entry.NO_LINK
```
mjvkeulen, line 1745: Is it desired that teacher entries never (ever) yield grades?

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                                    "content", "journal_ids", "grades", "publish_grade")
+
+        assignment = Assignment.objects.get(pk=assignment_id)
+        template = Template.objects.get(pk=template_id)
+
+        request.user.check_permission('can_post_teacher_entries', assignment)
+        request.user.check_permission('can_grade', assignment)
+        request.user.check_permission('can_publish_grades', assignment)
+
+        if title == '':
```
Larspolo, line 53: `if not title:` cus title can also be None

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Check if all journals that have been selected to add the teacher entry to belong to the current assignment.
+        if Journal.objects.filter(assignment=assignment, pk__in=journal_ids).count() != len(journal_ids):
+            return response.bad_request('Could not find all selected journals.')
+
+        if grades is not None and any([float(grade) < 0 for grade in grades.values()
+                                       if (grade is not None and grade != '')]):
+            return response.bad_request('All given grades must be greater than or equal to zero.')
+
+        grades = {int(journal_id): float(grade) for journal_id, grade in grades.items()
```
Larspolo, line 64: If you do this before the above check, it is a bit easier to check if any are below 0

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                                       journal_id in publish_grade and publish_grade[journal_id])
+                for content in teacher_entry.content_set.all():
+                    import_utils.import_content(content, copied_entry)
+
+                grading.task_journal_status_to_LMS(journal_id)
+            except Exception:
+                if copied_entry:
+                    copied_entry.delete()
+                if node:
+                    node.delete()
```
Larspolo, line 137: Sounds a bit weird to just ignore the journals that go wrong, do we not want to delete all (teacher) entries / nodes and return a bad response in this case?

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        journal_ids, grades, publish_grade = utils.required_params(request.data, "journal_ids", "grades",
+                                                                   "publish_grade")
+        teacher_entry_id, = utils.required_typed_params(kwargs, (int, 'pk'))
+
+        teacher_entry = TeacherEntry.objects.get(pk=teacher_entry_id)
+
+        request.user.check_permission('can_post_teacher_entries', teacher_entry.assignment)
+        request.user.check_permission('can_grade', teacher_entry.assignment)
+        request.user.check_permission('can_publish_grades', teacher_entry.assignment)
```
Larspolo, line 164: ditto here, I was hoping these checks are done when saving the role

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        if Journal.objects.filter(assignment=teacher_entry.assignment, pk__in=journal_ids).count() != len(journal_ids):
+            return response.bad_request('Could not find all selected journals.')
+
+        if grades is not None and any([float(grade) < 0 for grade in grades.values()
+                                       if (grade is not None and grade != '')]):
+            return response.bad_request('All given grades must be greater than or equal to zero.')
+
+        grades = {int(journal_id): float(grade) for journal_id, grade in grades.items()
+                  if (grade is not None and grade != '')}
+        publish_grade = {int(journal_id): published for journal_id, published in publish_grade.items()}
```
Larspolo, line 176: `publish_grades`*?

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                raise e
+
+        for (file, created_content) in files_to_establish:
+            file_handling.establish_file(request.user, file.access_id, content=created_content,
+                                         in_rich_text=created_content.field.type == Field.RICH_TEXT)
+
+        # Delete old user files
+        file_handling.remove_unused_user_files(request.user)
+
+        for journal_id in journal_ids:
```
Larspolo, line 115: Would like to see this loop get moved to a function (can be within this file) to reduce copied code

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+
+                if journal_id in grades and grades[journal_id] is not None:
+                    factory.make_grade(copied_entry, request.user.pk, grades[journal_id],
+                                       journal_id in publish_grade and publish_grade[journal_id])
+                for content in teacher_entry.content_set.all():
+                    import_utils.import_content(content, copied_entry)
+
+                grading.task_journal_status_to_LMS(journal_id)
+            except Exception:
+                # TODO TEACHERENTRY: What could go wrong here? How shall we proceed?
```
Larspolo, line 218: Would suggest removing all created (teacher) entries / nodes and return a bad_request. And post a more detailed exception to sentry

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+from django.test import TestCase
+
+from VLE.models import Entry, Field, Node, TeacherEntry
+from VLE.serializers import EntrySerializer
+from VLE.utils.error_handling import VLEPermissionError
+
+
+class TeacherEntryAPITest(TestCase):
+    def setUp(self):
+        self.admin = factory.Admin()
```
Larspolo, line 14: Not used

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+        api.create(self, 'teacher_entries', params=self.valid_create_params, user=self.teacher)
+
+        # Check if teacher can still create teacher entry when assignment is locked
+        self.assignment.unlock_date = None
+        self.assignment.lock_date = date.today() - timedelta(1)
+        self.assignment.save()
+        api.create(self, 'teacher_entries', params=self.valid_create_params, user=self.teacher)
+
+        self.assignment.lock_date = None
+        self.assignment.save()
```
Larspolo, line 126: This is not necessary, data will auto reset to `setUp` function

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+                create_params['content'] = {
+                    field.id: invalid_content[field.type]
+                }
+                api.create(self, 'teacher_entries', params=create_params, user=self.teacher, status=400)
+
+    def test_required_for_teacher_entry(self):
+        # Check if required field is required also for teacher entries.
+        required_field = factory.Field(template=self.template, required=True)
+        api.create(self, 'teacher_entries', params=self.valid_create_params, user=self.teacher, status=400)
+        required_field.delete()
```
Larspolo, line 176: ditto

src/vue/src/api/assignment.js
```diff
@@ -54,4 +54,14 @@ export default {
         return auth.get(`assignments/${id}/participants_without_journal`, null, connArgs)
             .then(response => response.data.participants)
     },
+
+    getTemplates (id, connArgs = auth.DEFAULT_CONN_ARGS) {
+        return auth.get(`assignments/${id}/templates`, null, connArgs)
+            .then(response => response.data.templates)
+    },
+
+    getTeacherEntries (id, connArgs = auth.DEFAULT_CONN_ARGS) {
```
Larspolo, line 63: should imo get moved to the dedicated `teacher_entries` url

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def partial_update(self, request, *args, **kwargs):
+        """Update an existing teacher entry.
+
+        Arguments:
+        pk -- the teacher entry to update
+        request -- the request that was sent with
+            journal_ids -- the journal ids of all journals which the entry should be part of
+            grades -- dict of grades with journal id as keys
+            publish_grade -- dict of grade publish state with journal id as keys
```
Larspolo, line 151: Would prefer a different structure, sth like:
```
request.data.entries = {
    [journal_id]: {
        'grade': [grade],
        'published_state': [published_state],
    }
    ...
}
```

Same goes for the create function ofc. Seems to be better aligned with

src/vue/src/components/assignment/PostTeacherEntry.vue
```diff
@@ -0,0 +1,288 @@
...
+                this.$toasted.error('Some required fields are empty.')
+            } else if (this.selectedJournals.length === 0) {
+                this.$toasted.error('No journals selected.')
+            } else if (((this.sameGradeForAllEntries && !this.grade)
+                || (!this.sameGradeForAllEntries && this.selectedJournals.some(journal => !this.grades[journal.id])))
+                && !window.confirm('Students will be able to edit the entry if no grade is set. Are you sure you'
+                + ' want to post ungraded entries?')) {
+                this.$toasted.error('Teacher entry not posted: no grade set.')
+            } else {
+                this.requestInFlight = true
```
Larspolo, line 248: Would like to see a confirmation window before some teacher accidentally creates 50 entries in many different journals

src/vue/src/components/entry/Entry.vue
```diff
@@ -80,18 +80,16 @@
...
+                    <template
+                        v-if="(new Date(node.entry.last_edited).getTime() - new Date(node.entry.creation_date)
+                            .getTime()) / 1000 < 3"
+                    >
+                        Submitted:
+                    </template>
+                    <template v-else>
+                        Last edited:
+                    </template>
+                    {{ $root.beautifyDate(node.entry.last_edited) }} by {{ node.entry.last_edited_by }}
```
Larspolo, line 92: `last_edited_by` is (was?) only set in group_journals

src/vue/src/components/timeline/TimelineNodeInfo.vue
```diff
@@ -38,7 +38,11 @@ export default {
         nodeTitle () {
             switch (this.node.type) {
             case 'e':
-                return this.node.entry.template.name
+                if (this.node.entry.title) {
```
Larspolo, line 41: if you accept previous change in serializer, then this can be reduced to `return this.node.entry.title`

src/django/VLE/views/assignment.py
```diff
@@ -555,3 +556,27 @@ def participants_without_journal(self, request, pk):
...
+        assignment = Assignment.objects.get(pk=pk)
+
+        request.user.check_permission('can_post_teacher_entries', assignment)
+
+        return response.success({'templates': TemplateSerializer(assignment.format.template_set.filter(
+            archived=False).order_by('name'), many=True).data})
+
+    @action(methods=['get'], detail=True)
+    def teacher_entries(self, request, pk):
+        """Get all templates (unlimited and preset-only) for an assignment.
```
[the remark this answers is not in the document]
engelhamer: Yes, the docstring is incorrect here. It is, however, a detail function for a specific assignment id though. Would it not be odd to move it to the teacher entry file as a detailed function despite it not being detailed to a specific teacher entry?

src/django/VLE/views/assignment.py
```diff
@@ -555,3 +556,27 @@ def participants_without_journal(self, request, pk):
...
+
+        Returns a list of templates."""
+        assignment = Assignment.objects.get(pk=pk)
+
+        request.user.check_permission('can_post_teacher_entries', assignment)
```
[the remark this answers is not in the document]
engelhamer: It is currently only used for the teacher entries and does contain preset only templates (which are reserved only for those who can edit assignment otherwise).

src/django/VLE/views/assignment.py
```diff
@@ -555,3 +556,27 @@ def participants_without_journal(self, request, pk):
...
         return response.success({'participants': participants_without_journal})
+
+    @action(methods=['get'], detail=True)
+    def templates(self, request, pk):
+        """Get all templates (unlimited and preset-only) for an assignment.
+
+        Returns a list of templates."""
+        assignment = Assignment.objects.get(pk=pk)
+
+        request.user.check_permission('can_post_teacher_entries', assignment)
```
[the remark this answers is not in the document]
engelhamer: It is currently only used for the teacher entries and does contain preset only templates (which are reserved only for those who can edit assignment otherwise).

Can view should not be required since you cannot post teacher entries if you cant view in the first place?

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        title, show_title_in_timeline, assignment_id, template_id, content_dict, journal_ids, grades, publish_grade \
+            = utils.required_params(request.data, "title", "show_title_in_timeline", "assignment_id", "template_id",
+                                    "content", "journal_ids", "grades", "publish_grade")
+
+        assignment = Assignment.objects.get(pk=assignment_id)
+        template = Template.objects.get(pk=template_id)
+
+        request.user.check_permission('can_post_teacher_entries', assignment)
+        request.user.check_permission('can_grade', assignment)
+        request.user.check_permission('can_publish_grades', assignment)
```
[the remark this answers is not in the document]
engelhamer: That constraint is enforced on the role config, however it is sensible to have explicit checks as grades can be added here right?

src/django/VLE/views/teacher_entry.py
engelhamer: But then you would still need to check for `grade is not None`

src/django/VLE/views/teacher_entry.py
engelhamer: This is simpler to handle than nested dicts (especially frontend)

src/vue/src/components/assignment/PostTeacherEntry.vue
engelhamer: So always a confirmation? Not only for 'weird' things like empty grade or something? We don't do confirm for bonus either right?

src/django/VLE/models.py
```diff
@@ -1682,22 +1694,17 @@ def is_graded(self):
...
 
-        try:
-            node = Node.objects.get(pk=node_id) if node_id else self.node
-        except Node.DoesNotExist:
-            raise ValidationError('Saving entry without corresponding node')
-
-        author = self.author if self.author else User.objects.get(pk=author_id) if author_id else None
-        if author:
-            if not node.journal.authors.filter(user=author).exists():
-                raise ValidationError('Saving entry created by user not part of journal.')
```
mjvkeulen, line 1696: Would prefer to keep this validation rule, during the save how can we determine the entry is created as a teacher entry?

src/django/VLE/serializers.py
```diff
@@ -694,6 +701,50 @@ def get_author(self, grade):
...
+            # Only include the actual content (so e.g. text).
+            if content.field.type == Field.FILE:
+                try:
+                    content_dict[content.field.id] = FileSerializer(FileContext.objects.get(pk=content.data)).data
+                except FileContext.DoesNotExist:
+                    capture_message(
+                        f'FILE content {content.pk} refers to unknown file in data: {content.data}', level='error')
+                    return None
+            else:
+                content_dict[content.field.id] = content.data
```
mjvkeulen, line 732: Is this not a duplicate block?

src/django/VLE/views/teacher_entry.py
engelhamer: Good idea

src/vue/src/components/entry/Entry.vue
engelhamer: Now always set, such that students can see some entry has been posted by a teacher as well.

src/django/VLE/models.py
engelhamer: Maybe directly set the `teacher_entry` foreign key upon creation (currently done afterwards)?

src/django/VLE/serializers.py
engelhamer: This is indeed duplicate: it used to be a content serializer but that was requested for removal in my previous PR 😅

src/vue/src/components/assignment/TeacherEntries.vue
```diff
@@ -0,0 +1,279 @@
...
+                                name="trash"
+                                class="trash-icon"
+                                @click.native="selectedJournals.pop(journal)"
+                            />
+                        </b-td>
+                    </b-tr>
+                </b-tbody>
+            </b-table-simple>
+
+            <b-button
```
mjvkeulen, line 135: image

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+        empty_create_params = self.valid_create_params.copy()
+        empty_create_params['template_id'] = all_types.id
+        empty_create_params['content'] = {}
+
+        api.create(self, 'teacher_entries', params=empty_create_params, user=self.teacher)
+
+        # As all fields are optional, data None should also work
+        empty_create_params['content'] = {field.id: None for field in fields.all()}
+
+        api.create(self, 'teacher_entries', params=empty_create_params, user=self.teacher)
```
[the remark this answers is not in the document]
engelhamer: But this is a different API.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+        api.create(self, 'teacher_entries', params=self.valid_create_params, user=self.teacher, status=400)
+        required_field.delete()
+
+    def test_preset_only_template_teacher_entry(self):
+        # Check if preset only templates are available for teacher entries.
+        preset_only_template = factory.PresetOnlyTextTemplate(format=self.format)
+        preset_only_create_params = self.valid_create_params.copy()
+        preset_only_create_params['template_id'] = preset_only_template.id
+        fields = Field.objects.filter(template=preset_only_template).all()
+        preset_only_create_params['content'] = {field.id: 'test data' for field in fields}
```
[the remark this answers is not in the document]
engelhamer: Can be made into an issue for the future.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+            Field.TEXT: 'text',
+            Field.RICH_TEXT: '<p> RICH </p>',
+            Field.VIDEO: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ',
+            Field.URL: 'https://ejournal.app',
+            Field.DATE: '2019-10-10',
+            Field.DATETIME: '2019-10-10T12:12:00',
+            Field.SELECTION: 'a',
+        }
+        invalid_content = {
+            # Field.VIDEO: 'https://www.test.com/watch?v=dQw4w9WgXcQ', TODO: This should fail.
```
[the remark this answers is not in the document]
engelhamer: This is not for now (todo was already there and issue has been made).

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+        self.assignment.save()
+
+    def test_valid_teacher_entry(self):
+        """Attempt to post an entry with valid and invalid content for each field."""
+        # Get a template which contains all field types (each of them optional).
+        template = factory.TemplateAllTypes(format=self.format)
+        fields = Field.objects.filter(template=template)
+
+        # Define valid and invalid contents for each of the fields.
+        valid_content = {
```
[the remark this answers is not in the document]
engelhamer: Make an issue for later.

src/django/VLE/views/assignment.py
```diff
@@ -555,3 +556,27 @@ def participants_without_journal(self, request, pk):
                 })
 
         return response.success({'participants': participants_without_journal})
+
+    @action(methods=['get'], detail=True)
+    def templates(self, request, pk):
```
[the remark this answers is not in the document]
engelhamer: This is a detail API call which takes an assignment ID, not a format ID. Hence it is best placed here in my opinion.

src/django/VLE/utils/import_utils.py
```diff
@@ -218,13 +218,13 @@ def copy_entry(entry, node, grade=None, vle_coupling=None):
     entry = Entry.objects.create(
         node=node,
         template=entry.template,
-        grade=grade,
```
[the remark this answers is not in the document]
engelhamer: > Went through the python files first.
> 
> * A teacher is not able to edit created entries content? If an edit needs to be made this requires a deletion, and there is no way to manage the position of the entry in the timeline. So a correction will shift the entry location?

That is correct. Since it is possible that teachers do not grade teacher entries, some students' content might differ from the original (if the student changed it before grading). Since it would be ambiguous how to tackle teacher changes after creation for these entries, teacher entry contents themselves are immutable.

> * Journals to add missing a select all option

Can add that.

> * After posting a teacher entry a refresh is required before managing said entries.

No manual action is required: clicking the manage teacher entries button should show the newly created entry without the need to f5.

engelhamer:
> Went through the python files first.
> 
> * A teacher is not able to edit created entries content? If an edit needs to be made this requires a deletion, and there is no way to manage the position of the entry in the timeline. So a correction will shift the entry location?

That is correct. Since it is possible that teachers do not grade teacher entries, some students' content might differ from the original (if the student changed it before grading). Since it would be ambiguous how to tackle teacher changes after creation for these entries, teacher entry contents themselves are immutable.

> * Journals to add missing a select all option

Can add that.

> * After posting a teacher entry a refresh is required before managing said entries.

No manual action is required: clicking the manage teacher entries button should show the newly created entry without the need to f5.

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: It is then consistent with the `GradeView` for example

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: `is_published` state can be `False` and TA's with `can_post_teacher_entries` can then still access this call

src/django/VLE/serializers.py
Larspolo: Whoops...

Larspolo: You could also inherit the EntrySerialzier, and only specify the fields you want to have

src/django/VLE/models.py
engelhamer: Also after creation it is not (currently) desirable that a teacher entry ever gets a grade or VLE coupling. `is_new` is thus too narrow a protection.

mjvkeulen:
> No manual action is required: clicking the manage teacher entries button should show the newly created entry without the need to f5.

Doesn't work atm, after posting a teacher entry clicking on manage teacher entries yields no results

src/vue/src/api/assignment.js
engelhamer: See comment on assignment pk detail.

src/django/VLE/models.py
```diff
@@ -1624,11 +1629,13 @@ class Entry(CreateUpdateModel):
     SENT_SUBMISSION = 'Submission is successfully received by VLE'
     NEEDS_GRADE_PASSBACK = 'Grade needs to be sent to VLE'
     LINK_COMPLETE = 'Everything is sent to VLE'
+    NO_LINK = 'Ignore VLE coupling (e.g. for teacher entries)'
```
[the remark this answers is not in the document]
engelhamer: Yes, but the VLE coupling field could not be `None` (so this is more verbose and safe)

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        grades = {int(journal_id): float(grade) for journal_id, grade in grades.items()
+                  if (grade is not None and grade != '')}
+        publish_grade = {int(journal_id): published for journal_id, published in publish_grade.items()}
+
+        for existing_entry in Entry.objects.filter(teacher_entry=teacher_entry):
+            journal_id = existing_entry.node.journal.pk
+            if journal_id not in journal_ids:
+                # Entry and node must be deleted.
+                existing_entry.node.delete()
+                existing_entry.delete()
```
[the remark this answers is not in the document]
engelhamer: Both the node **and** entry are deleted, so this should not leave anything behind. This is tested in https://github.com/eJourn-al/eJournal/blob/e3bbe76a3577c1e7bfcd327dd6a90e9298081623/src/django/test/test_teacher_entry.py#L102

engelhamer:
> > No manual action is required: clicking the manage teacher entries button should show the newly created entry without the need to f5.
> 
> Doesn't work atm, after posting a teacher entry clicking on manage teacher entries yields no results

On Chrome it does. What browser do you use?

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                        existing_entry.grade.published != (journal_id in publish_grade and publish_grade[journal_id]):
+                    # Current grade differs from previous grade or published. Need to update.
+                    factory.make_grade(existing_entry, request.user.pk, grades[journal_id],
+                                       journal_id in publish_grade and publish_grade[journal_id])
+                    grading.task_journal_status_to_LMS(journal_id)
+                # No new entry needs to be created for this journal in the next step.
+                journal_ids.remove(journal_id)
+            elif existing_entry.grade:
+                # Entry used to have a grade, but now does not anymore. This is not possible.
+                return response.bad_request('It is not possible to remove a grade from an entry.')
```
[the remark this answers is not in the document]
engelhamer: I don't understand what you mean.

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+
+        for journal_id in journal_ids:
```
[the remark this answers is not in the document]
engelhamer: Moved.

src/django/VLE/views/teacher_entry.py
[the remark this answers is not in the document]
mjvkeulen: I believe the (downwards) cascade does not work, could you add the proposed test in Z#1150 to your test delete function? Here it is important we are not just testing with an entry template consisting of text but rather all types.

src/django/VLE/views/teacher_entry.py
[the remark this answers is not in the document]
mjvkeulen: The logic in this block and the block below would be well suited for a dedicated function that could be tested in isolation.

src/django/VLE/models.py
```diff
@@ -1709,6 +1716,37 @@ def to_string(self, user=None):
         return "Entry"
 
 
+class TeacherEntry(Entry):
+    """TeacherEntry.
+
+    An entry posted by a teacher to multiple student journals.
```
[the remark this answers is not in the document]
engelhamer: When you e.g. serialize this, you get an entry posted by a teacher to multiple student journals: this literal thing.

src/django/VLE/views/teacher_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        pk, = utils.required_typed_params(kwargs, (int, 'pk'))
+
+        teacher_entry = TeacherEntry.objects.get(pk=pk)
+        assignment = teacher_entry.assignment
+
+        request.user.check_permission('can_post_teacher_entries', assignment, 'You are not allowed to post or delete '
+                                      'teacher entries.')
+
+        # Delete all nodes corresponding to entries made for this teacher entry.
+        Node.objects.filter(entry__teacher_entry=teacher_entry).delete()
```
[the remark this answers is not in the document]
engelhamer: Entry instances are deleted along with the `teacher_entry` foreign key they are linked to (CASCADE). This happens just below.

src/django/test/test_assignment.py
```diff
@@ -972,6 +972,20 @@ def test_participants_without_journal(self):
         assert t1.user.pk not in ids, 'check if teacher is not in response'
         assert t2.pk not in ids, 'check if author of assignment is not in response'
 
+    def test_get_templates(self):
```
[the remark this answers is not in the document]
engelhamer: The API takes an assignment ID.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+            'template_id': self.template.id,
+            'journal_ids': [self.journal1.id],
+            'content': {},
+            'grades': {
+                self.journal1.id: 1,
+            },
+            'publish_grade': {
+                self.journal1.id: True,
+            },
+        }
```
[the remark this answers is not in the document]
engelhamer: Mixed, some do some don't.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+class TeacherEntryAPITest(TestCase):
+    def setUp(self):
+        self.admin = factory.Admin()
+        self.assignment = factory.Assignment()
+        self.journal1 = factory.Journal(assignment=self.assignment)
+        self.student1 = self.journal1.authors.first().user
+        self.journal2 = factory.Journal(assignment=self.assignment)
+        self.teacher = self.assignment.author
+        self.format = self.assignment.format
+        self.template = factory.TextTemplate(format=self.format)
```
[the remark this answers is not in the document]
engelhamer: All types is already tested below.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+            'content': {},
+            'grades': {
+                self.journal1.id: 1,
+            },
+            'publish_grade': {
+                self.journal1.id: True,
+            },
+        }
+        fields = Field.objects.filter(template=self.template).all()
+        self.valid_create_params['content'] = {field.id: 'test data' for field in fields}
```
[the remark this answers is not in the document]
engelhamer: Suggest an issue for all instances of content. Not inherent to this PR or issue.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+                              user=self.teacher)['teacher_entry']
+            resp2 = api.create(self, 'teacher_entries', params=graded_create_params,
+                               user=self.teacher)['teacher_entry']
+
+            assert resp['id'] != resp2['id'], 'Multiple creations should lead to different ids'
+
+            journal1_entry = Entry.objects.filter(teacher_entry__pk=resp['id'], node__journal=self.journal1)
+            assert journal1_entry.exists(), 'Teacher entry should be added to student journal'
+
+            journal1_entry = journal1_entry.first()
```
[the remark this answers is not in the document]
engelhamer: That is already tested in the copy entry stuff.

src/django/test/test_teacher_entry.py
```diff
@@ -0,0 +1,293 @@
...
+            self.assertRaises(
+                VLEPermissionError,
+                self.student1.check_can_edit,
+                journal1_entry,
+            )
+
+            # Student (who does not have permission can_post_teacher_entries) cannot post teacher entries.
+            api.create(self, 'teacher_entries', params=graded_create_params, user=self.student1, status=403)
+
+    def test_create_ungraded_teacher_entry(self):
```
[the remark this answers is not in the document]
engelhamer: That is already inherent to creating an entry without a grade.

Makefile
```diff
@@ -203,7 +203,7 @@ db-restore:
 	@read -r a
 	@sudo su -c "psql \
 	-c \"DROP DATABASE IF EXISTS $(postgres_db)\" \
-	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)\" \
+
```
mjvkeulen, line 206: hline

src/django/VLE/models.py
```diff
@@ -1709,6 +1721,47 @@ def to_string(self, user=None):
...
+    teacher_entry = None
+    vle_coupling = None
+
+    def save(self, *args, **kwargs):
+        is_new = not self.pk
+        self.grade = None
+        self.teacher_entry = None
+        self.vle_coupling = Entry.NO_LINK
+
+        if not self.title:
```
mjvkeulen, line 1753: As this requires no context besides the field value itself, this would be best suited as a field validation rule.

mjvkeulen:
```
title = models.TextField(
    null=False,
    validators = [non_empty_string], 
)
```

src/django/test/test_teacher_entry.py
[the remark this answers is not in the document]
mjvkeulen: Z#1176

src/django/test/test_teacher_entry.py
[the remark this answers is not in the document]
mjvkeulen: This test if one can make a teacher entry (already tested), the optional fields is already tested in `test_validate_entry_content`. What other paths are tested?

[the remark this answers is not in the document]
engelhamer: Now mocked to ensure validation is called with the right params. Further relies on your test.

src/django/VLE/views/teacher_entry.py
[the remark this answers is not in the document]
engelhamer: There is no 'proposed test' described there.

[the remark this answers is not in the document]
mjvkeulen: I've done it myself in #993

src/django/VLE/views/teacher_entry.py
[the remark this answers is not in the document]
engelhamer: They are now tested better in `test_update_teacher_entry`.

src/django/test/factory/template.py
```diff
@@ -33,6 +33,17 @@ def gen_fields(self, create, extracted):
         FieldFactory(type=Field.TEXT, title='optional', template=self, required=False)
 
 
+class PresetOnlyTextTemplateFactory(TemplateFactory):
```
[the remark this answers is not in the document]
engelhamer: Nice!

src/django/test/test_assignment.py
```diff
@@ -972,6 +972,20 @@ def test_participants_without_journal(self):
...
+        teacher = factory.Teacher()
+        course = factory.Course(author=teacher)
```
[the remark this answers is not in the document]
engelhamer: @mjvkeulen the `student` in the rewrite below turns out to be a teacher. This is such an inconvenience that I think it will be fewer lines simply doing it the original way.

image

src/django/test/test_assignment.py
```diff
@@ -973,18 +973,15 @@ def test_participants_without_journal(self):
         assert t2.pk not in ids, 'check if author of assignment is not in response'
 
     def test_get_templates(self):
-        teacher = factory.Teacher()
-        course = factory.Course(author=teacher)
```
mjvkeulen: Did this line break things? I would hope the same results are achieved by both lines of initialization?

mjvkeulen:
> > > No manual action is required: clicking the manage teacher entries button should show the newly created entry without the need to f5.
> > 
> > 
> > Doesn't work atm, after posting a teacher entry clicking on manage teacher entries yields no results
> 
> On Chrome it does. What browser do you use?

84.0.4147.135, also does not work on test atm (teacher-entries 49de0f3)

1. Assignment without teacher entries
2. Create teacher entry
3. Click manage teacher entries
4. Modal will yield:
image

Makefile
```diff
@@ -203,7 +203,6 @@ db-restore:
 	@read -r a
 	@sudo su -c "psql \
 	-c \"DROP DATABASE IF EXISTS $(postgres_db)\" \
-	-c \"DROP DATABASE IF EXISTS test_$(postgres_db)\" \
```
mjvkeulen, line 206: Best if we keep this line so we only have to fall back to the script if some of pytests threads did not clean properly.

src/django/VLE/models.py
engelhamer: You can't `raise` in a lambda and I think that creating a separate function elsewhere kind of defeats the purpose for this change (harder to read / have to look multiple places).

mjvkeulen: This is the recommend Django approach. We already use some of these features by inheritance, which is indeed not always explicit. However, by following the guidelines here we could move towards the recommended practice and create greater awareness of validation at the field level itself.

src/django/test/test_teacher_entry.py
```diff
@@ -180,22 +137,11 @@ def test_preset_only_template_teacher_entry(self):
...
-        empty_create_params['template_id'] = all_types.id
-        empty_create_params['content'] = {}
-
-        api.create(self, 'teacher_entries', params=empty_create_params, user=self.teacher)
-
-        # As all fields are optional, data None should also work
-        empty_create_params['content'] = {field.id: None for field in fields.all()}
-
-        api.create(self, 'teacher_entries', params=empty_create_params, user=self.teacher)
+    @mock.patch('VLE.utils.entry_utils.check_fields')
```
mjvkeulen, line 155: Nice, should use this more often.

src/django/VLE/views/teacher_entry.py
```diff
@@ -169,8 +169,8 @@ def partial_update(self, request, *args, **kwargs):
                     factory.make_grade(existing_entry, request.user.pk, grades[journal_id],
                                        journal_id in publish_grade and publish_grade[journal_id])
                     grading.task_journal_status_to_LMS.delay(journal_id)
-                # No new entry needs to be created for this journal in the next step.
-                journal_ids.remove(journal_id)
+            # No new entry needs to be created for this journal in the next step.
+            journal_ids.remove(journal_id)
```
mjvkeulen, line 173: Should this not be accompanied by a test (assuming this was not caught by a test already?)

engelhamer:
Yes added a test for it just now. Will follow soon.

Op zo 30 aug. 2020 om 16:22 schreef mjvkeulen <notifications@github.com>:

Larspolo (approved):

mjvkeulen (approved):

src/django/test/test_teacher_entry.py
```diff
@@ -68,11 +70,65 @@ def test_create_graded_teacher_entry(self):
...
+
+        file_template_params['content'] = {str(fields[0].id): {'id': 0}}
+
+        # File content ID is invalid.
+        api.create(self, 'teacher_entries', params=file_template_params, user=self.teacher, status=404)
+
+        assert teacher_entry_count == TeacherEntry.objects.count(), 'Failed request should not leave teacher entry'
+
+        with patch('VLE.utils.file_handling.get_files_from_rich_text', side_effect=Exception()):
+            # Errors in rich text content should not leave dangling teacher entries.
```
mjvkeulen, line 120: This test does not touch db before and after state?

src/django/test/test_teacher_entry.py
```diff
@@ -1,11 +1,13 @@
 import test.factory as factory
+from copy import deepcopy
 from datetime import date, timedelta
 from test.utils import api
 from unittest import mock
+from unittest.mock import patch
```
mjvkeulen: duplicate import really

engelhamer: Was removed in latest commit?

src/django/VLE/models.py
```diff
@@ -1456,6 +1456,20 @@ def to_string(self, user=None):
         return self.get_name()
 
 
+def CASCADE_IF_UNLIMITED_ENTRY_NODE_ELSE_SET_NULL(collector, field, sub_objs, using):
+    # NOTE: Either the function is not yet defined or the node type is not defined.
+    # Tag Node.Entry, update hard coded if changed
+    unlimited_entry_nodes = [n for n in sub_objs if n.type == 'e']
```
Larspolo, line 1463: Would prefer using `Node.ENTRY` iso `'e'`. Would also prefer this being inside the Node class (as class function), as this is specific to that class right?

engelhamer: The comment right above explains that (and why) this is not possible if I understand correctly?

PR #994, opened 2020-08-28
[WIP] 1.4 Checklist fixes

mjvkeulen:
## Description
Fixes issues that came to light during the 1.4 checklist


src/django/VLE/serializers.py
```diff
@@ -343,45 +344,44 @@ def get_stats(self, assignment):
...
-            # Get all users that are in any of the request user's groups.
-            own_participation = own_participation.first()
-            own_group_users = all_users.filter(participation__groups__in=own_participation.groups.all())
+        # Upcoming specifically requests stats for all of the assignment's courses
+        if 'course' in self.context and self.context['course'] is settings.EXPLICITLY_WITHOUT_CONTEXT:
+            relevant_stat_users = stats_utils.assignment_generic_stat_users(assignment, self.context['user'])
+        else:
+            # We are not dealing explicitly without context, and we have no specific course, stats should not be needed.
+            if course is None:
+                return None
```
Larspolo, line 355: If course is not set, it gets set automatically on line 345https://github.com/eJourn-al/eJournal/pull/994/commits/25a9ea19370c634b14cf36394c788fcfa0a1c445#diff-1e23cec44a98717a9bd2632f844c82f2R345

src/django/VLE/serializers.py
```diff
@@ -343,45 +344,44 @@ def get_stats(self, assignment):
...
 
         return stats
 
     def get_course(self, assignment):
         return CourseSerializer(self._get_course(assignment)).data
 
     def _get_course(self, assignment):
-        if self.context.get('course', None) is None:
+        course = self.context.get('course', None)
+        if course is None or course == settings.EXPLICITLY_WITHOUT_CONTEXT:
```
Larspolo, line 384: shouldnt `EXPLICITLY_WITHOUT_CONTEXT` mean that no course is retrieved?

src/django/VLE/utils/statistics.py
```diff
@@ -0,0 +1,62 @@
+"""
+Utility functions in relation to generic statistics
+"""
+
+from VLE.models import Course, Participation
+from VLE.utils.error_handling import VLEProgrammingError
+
+
+# QUESTION: @Lars, can any of these queries result in journals for users without can_have_journal=True?
```
Larspolo, line 9: Yes, all dont check for that. However, I think that is correct. In serializer we also do `Journal.objects.filter(assignment=assignment, authors__user__in=...)` to filter out everyone that does not have a journal

src/django/VLE/utils/statistics.py
```diff
@@ -0,0 +1,62 @@
...
+def course_specific_stat_users(assignment, course, user):
```
Larspolo, line 10: Would change these two functions into one like:

```
def get_user_lists_with_scopes(assignment, user, course=None):
    """Get lists of users in [assignment] in different scopes connected to [user].
    Allows for setting [course] to limit the users retrieved to only being from that course.

    Args:
        - assignment: Assignment of which stats are desired
        - user: Educator requesting stats
        - course: Course for which stats are specifically requested

    Returns:
        dict:
            - all: All users which are participants of the assignment. If course is set, it is limited to only users that participate in Course
            - own: Subset of all with which [user] shares a group. If course is set, it is limited to only users that participate in Course
    """
    if course is None:
        shared_courses = assignment.courses.filter(users=user)
    else:
        shared_course = Course.objects.filter(pk=course.pk, users=user)
    p_all = Participation.objects.filter(
        course__in=shared_course,
        user__in=assignment.assignmentparticipation_set.values('user')
    )
    p_own = p_all.filter(groups__in=user.participation_set.values('groups'))

    return {'all': p_all.values('user').distinct(), 'own': p_own.values('user').distinct()}
```

mjvkeulen: Fair, what would be the toggle to switch between `all`, `own course`, `own groups`?

src/django/VLE/serializers.py
mjvkeulen: Stats of an assignment where course context is `EXPLICITLY_WITHOUT_CONTEXT` should yield all the assignment's courses' stats for the context's user. The line you refer to will set course to None, so then requesting `course_specific_stat_users` with a course of None makes no sense. This is why in this scenario stats are simply None?

src/django/VLE/utils/statistics.py
mjvkeulen: Actually I quite like the explicitness of the naming, especially seeing our earlier discussion. With the new test in place, refactoring should be safe, and I don't care too much (so if you insist go ahead). But like I said, seeing our earlier discussion (and the actual error being fixed) warrants the explicit naming of the desired result over falling back to a course of None by default.

src/django/VLE/utils/statistics.py
mjvkeulen: I thought so too, removing question and resolving

src/django/VLE/utils/statistics.py
Larspolo: `all, own course, own groups` will always be returned

engelhamer (approved):

Larspolo (approved):

PR #1000, opened 2020-09-04
Improved the ease of use of the Django shell

mjvkeulen:
## Description
Added a file which `make shell` will always run directly after being called. This saves us the pain of always importing VLE.models and other commonly used modules.


mjvkeulen:
To save the reviewers some trouble:

image

It is important the `PYTHONSTARTUP` env variable does not persist, else it would be called after the startup of other python files.

Larspolo (approved):

src/django/shell_startup.py
```diff
@@ -0,0 +1,6 @@
+import test.factory as factory  # noqa: F401
```
engelhamer, line 1: Don't you think it is going to be confusing in practice that the `factory` import actually refers to the _test_ factory (with consequences as to what is being created).

engelhamer (changes requested):

src/django/shell_startup.py
mjvkeulen: You would have been welcome to change it to the full path or whatever suits your preferences, the current VLE.factory is slowly being phased out. If you want to test something in the shell, it is quite likely you use the test factories more often (especially since the preset db is created using these test factories).

engelhamer (approved):

PR #1002, opened 2020-09-05
Dedicated task for removing unused files

mjvkeulen:
## Description
Moved scattered file cleanup logic into a dedicated task. Currently, temp or no longer referenced files are cleaned in various places, and the cleanup itself is user-specific and user activity-based. Currently, a user can have as many temporary files as the storage limit allows, provided that the user does partake in any activity to clean previous files (e.g. post an entry).

- `Course` link removed from `FileContext` as currently, we have no course-specific files.
- Hopefully, the expansion of tests dedicated to file handling, and the more streamlined logic will further improve the robustness of our file handling.


src/django/test/test_cleanup.py
```diff
@@ -0,0 +1,112 @@
...
+        Field.objects.filter(
+            type=Field.NO_SUBMISSION, description__contains=f_description_fc.access_id).update(description='')
+        PresetNode.objects.filter(
+            format=assignment.format, description__contains=progress_description_fc.access_id).update(description='')
+        f()
+        assert not FileContext.objects.filter(
+            pk__in=[a_description_fc.pk, f_description_fc.pk, progress_description_fc.pk]).exists(), \
+            'Once the respective descriptions no longer refer to the FCs, they should be cleaned'
+
+    def test_remove_unused_files(self):
```
Larspolo, line 107: Now all these functions get executed twice. If you want to only test them in 1 function, you can change the other names to `check_...`, then pytest wont pick up on those seperatly as well

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+
+def remove_temp_files(older_lte=timezone.now() - relativedelta(days=1)):
+    """Remove temp files older than a day"""
+    VLE.models.FileContext.objects.filter(creation_date__lte=older_lte, is_temp=True).delete()
+
+
+def remove_unused_content_files():
+    """Remove overwritten files of RT and FILE context fields"""
+    VLE.models.FileContext.objects.filter(
+        ~Q(content__data=Cast(F('pk'), TextField())),
```
Larspolo, line 20: This is the same as `.exclude(content__data=Cast(F('pk'), TextField())`, right? If so, I prefer that notion

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+        content__field__type=VLE.models.Field.RICH_TEXT
+    ).delete()
+
+
+def remove_unused_comment_files():
+    """Removes unused comment files, both no longer attached and no longer referenced in RT"""
+    VLE.models.FileContext.objects.filter(
+        ~Q(comment__text__contains=F('access_id')),
+        comment__isnull=False,
+        comment_files__isnull=True,
```
Larspolo, line 35: This doesnt seem to delete any files anymore as long as there is at least 1 file in the `comment_files` field

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def remove_unused_profile_pictues():
+    """Removes unused files associated directly with a single user (profile picture)"""
+    VLE.models.FileContext.objects.filter(
+        ~Q(author__profile_picture__contains=F('access_id')),
+        content__isnull=True,
+        comment__isnull=True,
+        assignment__isnull=True,
+        journal__isnull=True,
```
Larspolo, line 56: Im starting to like the idea of having a computed type field in the filecontext. This seems to be error prone.

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+        found = VLE.models.Field.objects.filter(
+            template__format__assignment=fc.assignment, description__contains=fc.access_id).exists()
+        found = found or VLE.models.PresetNode.objects.filter(
+            format__assignment=fc.assignment, description__contains=fc.access_id).exists()
+        if not found:
+            fc.delete()
+
+
+@shared_task
+def remove_unused_files(older_lte=timezone.now() - relativedelta(days=1)):
```
Larspolo, line 80: This parameter gets set when launched, but does not dynamically update. A better method would be to do sth like:
```
 def remove_unused_files(older_then=None):
    if not older_then:
        older_then = timezone.now() - relativedelta(days=1)
```

Larspolo: Also would maybe increase the relativedelta even further. To sth like refresh token length (which is now "only" 2 days, would maybe ven consider increasing that, as people might be working on entries for longer than that)

Larspolo (changes requested):

src/django/test/test_cleanup.py
mjvkeulen: They are supposed to be executed twice, called individually they call their respective dedicated removal functions. Called from `test_remove_unused_files` it is tested whether calling the aggregated cleanup function does not clean any of its individual parts (which contain healthy state).

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: There is much to improve on our FC structure, most of these flags should simply be set from the respective model itself (e.g.  `attached_comment_files` and `rt_comment_files` or sth)

src/django/test/factory/file_context.py
```diff
@@ -4,7 +4,15 @@
...
 
-from VLE.models import Field, Journal
+from VLE.models import Field, Journal, PresetNode
+
+
+def _fc_to_rt_img_element(fc):
+    return f'<img src="{fc.download_url(access_id=fc.access_id)}"/>'
+
+
+def _none_to_str(el):
```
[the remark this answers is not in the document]
mjvkeulen: It is private exactly for that reason tbh, this pattern occurs so often in this file I think this is perfectly reasonable. If you want to tweak the name go for it.

src/django/VLE/factory.py
```diff
@@ -328,3 +328,19 @@ def make_grade(entry, author, grade, published=False):
     entry.save()
 
     return grade
+
+
+# QUESTION: How about using a test factory instead?
+# This would go towards merging our factories, syncing our test data and production code etc.
```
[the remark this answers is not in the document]
mjvkeulen: I will move this to a different forum for further discussion.

src/django/VLE/views/comment.py
```diff
@@ -20,11 +20,10 @@ def handle_comment_files(user, files, comment):
         file = FileContext.objects.get(pk=int(file_id))
         if not comment.files.filter(pk=file.pk).exists():
             comment.files.add(file)
-            file_handling.establish_file(user, file.access_id, comment=comment)
-    # Remove old files
+            file_handling.establish_file(author=user, identifier=file.access_id, comment=comment)
+    # Remove old atached files
```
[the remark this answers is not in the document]
mjvkeulen: I updated it, for comments RT and temp files are removed in the cleanup task.

Attached files can also be removed during cleanup, but since in this call it clear what should be removed I left it as is. This is the benefit of using a direct DB layer relationship, instead of an ID in text or similar.

src/django/VLE/views/journal.py
```diff
@@ -209,18 +208,7 @@ def partial_update(self, request, *args, **kwargs):
...
-                    file_name=content_file.name,
-                    author=request.user,
-                    journal=journal,
-                    is_temp=False,
-                    in_rich_text=True,
-                )
-                journal.stored_image = file.download_url(access_id=True)
-                journal.save()
-                file_handling.remove_unused_user_files(request.user)
+                VLE.factory.make_journal_image(content_file, journal, request.user)
```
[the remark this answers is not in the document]
mjvkeulen: This is directly tied to the question I posed until that discussion is settled a factory call is as good as any.

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: I quite like being able to use Q for a negated query like this. Would be interesting to see what the difference in SQL would be if (if any)

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: Good catch, I tried to combined `~Q(comment__files__pk=F('access_id'))` but it throws an error which I simply don't follow. I will split the query for unused attached files and unused rt files (for comments).

mjvkeulen: Since `comment_files` is the reverse relation (from FC to comment) the query is correct. I updated the test to show this more thoroughly.

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: `Python's default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby).` I did not know this, good catch.

src/django/test/factory/file_context.py
```diff
@@ -133,17 +144,142 @@ class RichTextContentFileContextFactory(FileContentFileContextFactory):
...
+        self.journal.stored_image = self.download_url(access_id=True)
+        self.journal.save()
+
+    @factory.post_generation
+    def validate(self, create, extracted):
+        if not create:
+            return
+
+        if not self.journal.authors.filter(user=self.author).exists():
+            raise ValidationError('Journal FileContext created with an author not part of its journal.')
```
engelhamer, line 185: Teachers can also set a journal image for their students.

src/django/test/test_cleanup.py
```diff
@@ -0,0 +1,112 @@
...
+            fc.save()
+
+        non_temp_fc = factory.FileContext(is_temp=False)
+        temp_fc_new = factory.FileContext(is_temp=True)
+        temp_fc_6_hours = factory.FileContext(is_temp=True)
+        update_fc_creation_date(temp_fc_6_hours, timezone.now() - relativedelta(hours=6))
+        temp_fc_1_day = factory.FileContext(is_temp=True)
+        update_fc_creation_date(temp_fc_1_day, timezone.now() - relativedelta(days=2))
+
+        f()
```
engelhamer, line 26: Bad name.

src/django/test/test_cleanup.py
engelhamer: Add a comment.

engelhamer (changes requested):

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: ?

src/django/VLE/utils/file_handling.py
```diff
@@ -29,8 +29,6 @@ def get_file_path(instance, filename):
         return '{}/journalfiles/{}/{}'.format(instance.author.id, instance.journal.id, filename)
     elif instance.assignment is not None:
         return '{}/assignmentfiles/{}/{}'.format(instance.author.id, instance.assignment.id, filename)
-    elif instance.course is not None:
-        return '{}/coursefiles/{}/{}'.format(instance.author.id, instance.course.id, filename)
```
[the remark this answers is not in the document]
mjvkeulen: See PR description

src/django/test/factory/file_context.py
mjvkeulen: Good catch, will update

src/django/test/test_cleanup.py
mjvkeulen: Bad name? `f` for a generic function is a commonly used variable, similar to `i` for index. 

The test structure is quite clear, for each of the category FileContexts (e.g. comment or assignment level) we have dedicated queries, for each of these queries dedicated tests exist. It is important each query does not clean files of another category, which is why all tests are called again with the (combined) function `remove_unused_files`.

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -0,0 +1,87 @@
...
+    """Deletes floating user files."""
```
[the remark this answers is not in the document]
mjvkeulen: This is part of our daily business logic temp files which are cleaned, RT is edited and files are no longer referenced in the edited rich text.

src/django/VLE/tasks/beats/cleanup.py
engelhamer: Lars proposed a change which contained a typo

src/django/test/test_cleanup.py
mjvkeulen: Which is why `f` is not simply  `cleanup.remove_unused_profile_pictures` or `cleanup.remove_unused_files`

engelhamer: But then why not name this e.g. `removal_function` instead of f? That would make it easier to read

mjvkeulen: I will update it, but I believe the context makes this so obvious, that we are just nitpicking. `CleanupTest` class, which only consists of tests named `test_remove_X`, where X is the function (`f`) by default, and f() is called each time before still exists (or not) checks take place.

src/django/test/test_cleanup.py
mjvkeulen: This is expected behaviour, and quite clean if you ask me

src/django/VLE/tasks/beats/cleanup.py
Larspolo: No difference in SQL indeed

src/django/VLE/tasks/beats/cleanup.py
Larspolo: Created an issue

Larspolo (approved):

src/django/test/factory/file_context.py
engelhamer: Not yet done

engelhamer (changes requested):

engelhamer (approved):

PR #1007, opened 2020-09-08
Improved build scripts

mjvkeulen:


config/env_vars/base.yml
```diff
@@ -21,7 +21,7 @@ virtualenv_python_version: python3.6
 
 
 # Backup settings
-run_backups: true
+run_backups: false
```
Larspolo, line 24: Would prefer backups to still be made, this reduces the lost data when a deploy needs to be reverted. Only the copy to local machine can be off by default

Larspolo (changes requested):

Larspolo (approved):

engelhamer (approved):

PR #1008, opened 2020-09-08
Improved the robustness of (teacher) entry crash recovery

mjvkeulen:
Let us see what level of crash recovery is desired

For the tests it's important we do not forget how temporary files should be handled during crash recovery.

### DIFF
--- a/src/django/VLE/validators.py
+++ b/src/django/VLE/validators.py
@@ -90,8 +90,8 @@ def validate_entry_content(content, field):
     if field.type == VLE.models.Field.FILE:
         try:
             int(content['id'])
-        except (ValueError, KeyError):
-            raise ValidationError("The content['id'] of a field file should contain the pk of the related file")
+        except (ValueError, KeyError, TypeError):
+            raise ValidationError('The content of a field file should follow {field.pk: FileContext.pk}')
 
         # Ensures the FC still exists
         fc = VLE.models.FileContext.objects.get(pk=int(content['id']))
--- a/src/django/VLE/validators.py
+++ b/src/django/VLE/validators.py
@@ -91,7 +91,8 @@ def validate_entry_content(content, field):
         try:
             int(content['id'])
         except (ValueError, KeyError, TypeError):
-            raise ValidationError('The content of a field file should follow {field.pk: FileContext.pk}')
+            raise VLE.utils.error_handling.VLEProgrammingError(
+                'The content of a field file should follow {field.pk: FileContext.pk}')
 
         # Ensures the FC still exists
         fc = VLE.models.FileContext.objects.get(pk=int(content['id']))
--- a/src/django/test/test_teacher_entry.py
+++ b/src/django/test/test_teacher_entry.py
@@ -7,7 +7,7 @@ from unittest import mock
 from django.db import transaction
 from django.test import TestCase
 
-from VLE.models import Entry, Field, Grade, Node, TeacherEntry
+from VLE.models import Comment, Entry, Field, FileContext, Grade, Node, TeacherEntry
 from VLE.serializers import EntrySerializer
 from VLE.utils.error_handling import VLEPermissionError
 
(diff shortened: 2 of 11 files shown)


src/django/test/test_teacher_entry.py
```diff
@@ -329,3 +337,43 @@ def test_delete_teacher_entry(self):
...
+        check_db_state_after_exception(self, 'VLE.views.teacher_entry._copy_new_teacher_entry')
+        check_db_state_after_exception(self, 'VLE.utils.import_utils.copy_node')
+        check_db_state_after_exception(self, 'VLE.utils.import_utils.copy_entry')
+        check_db_state_after_exception(self, 'VLE.utils.import_utils.import_comment')
+        check_db_state_after_exception(self, 'VLE.utils.import_utils.import_content')
+        check_db_state_after_exception(self, 'VLE.factory.make_content')
+        check_db_state_after_exception(self, 'VLE.utils.file_handling.get_files_from_rich_text')
+        check_db_state_after_exception(self, 'VLE.utils.file_handling.establish_file')
+        check_db_state_after_exception(self, 'VLE.factory.make_grade')
+        check_db_state_after_exception(self, 'VLE.utils.grading.task_journal_status_to_LMS')
```
mjvkeulen, line 379: What would we like to see here?

src/django/test/test_teacher_entry.py
```diff
@@ -329,3 +338,54 @@ def test_delete_teacher_entry(self):
...
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_nodes, all_to_dict(Node))])
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_entries, all_to_dict(Entry))])
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_grades, all_to_dict(Grade))])
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_comments, all_to_dict(Comment))])
+            assert list(FileContext.objects.exclude(pk__in=temp_files).values_list('pk', flat=True)) == pre_crash_fcs
+            assert list(FileContext.objects.filter(author=teacher, is_temp=True).values_list('pk', flat=True)) \
+                == temp_files, \
+                'Teacher can reuse earlier uploaded temporary files, despite a crash occurring'
+
+        check_db_state_after_exception(self, 'VLE.views.teacher_entry._copy_new_teacher_entry')
```
mjvkeulen, line 382: So currently we fail here, because a teacher entry remains in the DB despite crash

src/django/test/test_teacher_entry.py
```diff
@@ -329,3 +338,54 @@ def test_delete_teacher_entry(self):
...
+        temp_files = list(FileContext.objects.filter(author=teacher, is_temp=True).values_list('pk', flat=True))
+
+        def check_db_state_after_exception(self, raise_exception_for):
+            with mock.patch(raise_exception_for, side_effect=Exception()):
+                self.assertRaises(
+                    Exception, api.create, self, 'teacher_entries', params=data, user=teacher)
+
+            # Check if DB state is unchanged after a crash
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_nodes, all_to_dict(Node))])
+            assert all([equal_models(pre, post) for pre, post in zip_equal(pre_crash_entries, all_to_dict(Entry))])
```
mjvkeulen, line 374: So currently we fail here (called with failing `_copy_new_teacher_entry`), because a teacher entry remains in the DB despite crash.

src/django/test/test_teacher_entry.py
```diff
@@ -2,7 +2,7 @@
 from copy import deepcopy
 from datetime import date, timedelta
 from test.utils import api
-from test.utils.generic_utils import equal_model_iterators
+from test.utils.generic_utils import _model_instance_to_dict, equal_models, zip_equal
```
Larspolo, line 5: Importing a private function is a bit weird

mjvkeulen: True, but I always feel its kinda ok in test files. Remove the prefix?

src/django/VLE/validators.py
```diff
@@ -90,8 +90,8 @@ def validate_entry_content(content, field):
     if field.type == VLE.models.Field.FILE:
         try:
             int(content['id'])
-        except (ValueError, KeyError):
-            raise ValidationError("The content['id'] of a field file should contain the pk of the related file")
+        except (ValueError, KeyError, TypeError):
+            raise ValidationError('The content of a field file should follow {field.pk: FileContext.pk}')
```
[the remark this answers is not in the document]
mjvkeulen: The user should only be able to mess this up with intent, ok if I change it to programming error?

PR #1009, opened 2020-09-09
Fixed a bug where email digest task could break on a single failed email send

mjvkeulen:
Digest email will now continue sending emails for all users despite a failure occurring during the individual send call of an email.


src/django/VLE/management/commands/preset_db.py
```diff
@@ -242,8 +242,8 @@ def gen_entries(self):
         for a in self.assignments:
             # NOTE: carefull to use a.journal_set or query via AP, both will yield teacher journals
             for j in Journal.objects.filter(assignment=a):
-                grade = random.choice([factory.Grade(grade=random.randint(0, 3)), None])
-                factory.UnlimitedEntry(node__journal=j, grade=grade, gen_content__from_file=True)
+                entry = factory.UnlimitedEntry(node__journal=j, gen_content__from_file=True)
+                random.choice([factory.Grade(grade=random.randint(0, 3), entry=entry), None])
```
Larspolo, line 246: I dont think this random.choice has any effect.
```
>>> random.choice([print('a'), None])
a
>>> random.choice([print('a'), None])
a
>>> random.choice([print('a'), None])
a
>>> random.choice([print('a'), None])
a
>>> random.choice([print('a'), None])
a
```

src/django/test/test_notification.py
```diff
@@ -301,3 +302,71 @@ def test_save_notification(self):
...
+        journal = factory.Journal(entries__n=0, assignment__courses=[course])
+        assignment = journal.assignment
+        student = journal.authors.first().user
+        teacher = journal.assignment.author
+        entry1 = factory.UnlimitedEntry(node__journal=journal, grade__grade=1)
+        entry2 = factory.UnlimitedEntry(node__journal=journal)
+
+        assert Notification.objects.count() == 5, '''
+            Expected notifications:
+                - NEW ASSiGNMENT: student
```
Larspolo, line 317: *ASSIGNMENT

Larspolo (approved):

src/django/VLE/management/commands/preset_db.py
mjvkeulen: Hah good catch, ok to fix this when we move back to non-random presetdb?

mjvkeulen: Could not reproduce
```
>>> print(random.choice(['a', None]))
None
>>> print(random.choice(['a', None]))
None
>>> print(random.choice(['a', None]))
a
>>> print(random.choice(['a', None]))
None
>>> print(random.choice(['a', None]))
a
>>> 
```

src/django/VLE/models.py
```diff
@@ -1111,21 +1111,13 @@ def save(self, *args, **kwargs):
                         journal = Journal.objects.create(assignment=self)
                         journal.add_author(ap)
 
-        # Send notifications once an assignment is published
-        if (is_new or not was_published) and self.is_published:
-            for ap in AssignmentParticipation.objects.filter(assignment=self):
-                if ap.user.has_permission('can_have_journal', self):
```
engelhamer, line 1117: Why was this permission check removed?

mjvkeulen: The impact of this is that TA's are now notified of the creation of a new assignment

engelhamer (approved):

PR #1011, opened 2020-09-11
Minor QoL improvements to the development process

mjvkeulen:
- Updated .editorconfig (yay to correct spacing by default when working in `.yaml` or `Makefile`)
- Combined some scattered configuration settings for multiple tools into a single file `setup.cfg`
- Cleaned some minor misconceptions in the Makefile stemming back to the PSE days


setup.cfg
```diff
@@ -0,0 +1,27 @@
+[pycodestyle]
```
Larspolo, line 1: From https://pypi.org/project/flake8/:
image

pycodestyle is not needed anymore (I dont think it ever was). Good time imo to remove it fully

Larspolo (approved):

engelhamer (approved):

PR #1013, opened 2020-09-14
Fixed possible version mismatch after deployment

mjvkeulen:
Frontend index.html is no longer cached, this coupled with hashed js output files during the Webpack build process, should trigger a reload of the frontend resources in the case of an update.

To solve the scenario where a user has a tab open with eJournal loaded during the deployment of a new version, a version check is made for each API request. If a code version mismatch is detected, the user is prompted to refresh the page at one's own initiative since this will also reset any data (e.g. a partially created entry).

User facing alert:
image (5)

- Fixes Z#1205


Larspolo (approved):
Would like to see a different banner, but if its okay with @engelhamer this can also do for now

engelhamer (approved):

PR #1016, opened 2020-09-15
Added group only notifications and JIR notifications

Larspolo:
## Description
image
Added group only notifications and JIR notifications
Also users now don't get notifications from:
- courses they cannot see
- courses where they are not in the active lti course
- groups they are not in, if that preference is set
- when the preference is set to off (previously it would still create the notification, just not mail it)

## Summary of changes
- Fixes Z#1158
- group only notifications


src/django/VLE/models.py
```diff
@@ -1029,6 +1081,20 @@ class Assignment(CreateUpdateModel):
...
+            courses = self.courses.all()
+        return User.objects.filter(
+            participations__in=courses,
+            pk__in=self.assignmentparticipation_set.values('user')
+        ).distinct()
+
+    def get_own_group_users(self, user, courses=None):
+        if courses is None:
+            courses = self.courses.all()
+        groups = user.participation_set.filter(course__in=courses).values('groups')
```
mjvkeulen, line 1095: user has a `groups` forward manager

src/django/VLE/models.py
```diff
@@ -2106,6 +2172,22 @@ def get_update_response(self):
...
         JournalImportRequest.objects.filter(source=instance, state=JournalImportRequest.PENDING).delete()
 
+    def save(self, *args, **kwargs):
+        is_new = self._state.adding
+
+        super().save(*args, **kwargs)
+
+        if is_new:
+            if self.target and self.source and self.state == self.PENDING:
+                for user in VLE.permissions.get_supervisors_of(self.target):
```
mjvkeulen, line 2182: There is a dedicated permission for managing JIRs

src/django/VLE/models.py
```diff
@@ -2106,6 +2172,22 @@ def get_update_response(self):
...
+                    Notification.objects.create(
+                        user=user,
+                        type=Notification.NEW_JOURNAL_IMPORT_REQUEST,
+                        jir=self,
+                    )
+        if self.state != self.PENDING:
+            Notification.objects.filter(jir=self).delete()
```
mjvkeulen, line 2189: Should notificatons already sent not remains like others?

src/django/VLE/models.py
```diff
@@ -653,7 +681,18 @@ def url(self):
         return gen_url(
             node=self.node, journal=self.journal, assignment=self.assignment, course=self.course, user=self.user)
 
+    def is_in_group(self):
+        return self.assignment.get_own_group_users(self.user).filter(pk__in=self.journal.authors.values('user'))
+
+    @property
+    def email_preference(self):
```
mjvkeulen, line 688: Was this preference purely tied to email? Or will this likely be tied to other stuff later (e.g. notification tray).

src/django/VLE/models.py
```diff
@@ -548,6 +559,15 @@ class Notification(CreateUpdateModel):
                 'button_text': 'View Comment',
             },
         },
+        NEW_JOURNAL_IMPORT_REQUEST: {
+            'name': 'new_journal_import_request_notifications',
+            'content': {
+                'title': 'New journal import request',
+                'content': '{journal} requested to import another journal.',
```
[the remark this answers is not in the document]
engelhamer: That is a bit too much info for me. I think the current setup is fine and consistent with the other notification content.

src/django/VLE/models.py
```diff
@@ -653,7 +681,18 @@ def url(self):
         return gen_url(
             node=self.node, journal=self.journal, assignment=self.assignment, course=self.course, user=self.user)
 
+    def is_in_group(self):
+        return self.assignment.get_own_group_users(self.user).filter(pk__in=self.journal.authors.values('user'))
```
engelhamer, line 685: Currently this would be a bit inconvenient for a course with only a `staff` group and students not in any group: for the assignment page's group filter, we exclude any groups that do not have journals in them.

Something similar could be done here as well: if there are no journals in the user's group, then include notifications for all journals?

src/django/VLE/models.py
engelhamer: The notification tray will likely always show all, and optionally filter or something?

engelhamer: So email is fine in that case.

src/vue/src/components/profile/NotificationCard.vue
```diff
@@ -177,6 +177,32 @@
                     />
                 </h2>
             </div>
+            <hr/>
+            <div class="clearfix">
+                <radio-button
```
engelhamer, line 182: For toggles between two states we usually display only one of the buttons (the active case). See e.g. `unlimited` vs `preset only` for the `TemplateEditor`. I would prefer that here as well for space reasons and simplicity for limited options.

engelhamer (changes requested):

src/django/VLE/models.py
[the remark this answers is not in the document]
Larspolo: More inline, and sending grades by email is not done

src/django/VLE/models.py
Larspolo: What do you mean?

src/django/VLE/models.py
Larspolo: > Was this preference purely tied to email? Or will this likely be tied to other stuff later (e.g. notification tray).

Didnt know which way we would prefer it for the notification tray, so I names it now "email_preference" as it was only hitting email stuff. When adding a notification tray this needs to be discussed

src/django/VLE/models.py
Larspolo: Should the "own group" behaviour in the ToDo task act the same?

engelhamer: Own group is fine this way because it should bot draw as much focus every time. Hence why its not a button.

src/django/VLE/models.py
mjvkeulen: I got confused with the groups from the permissions mixin, resolving.

src/django/VLE/models.py
mjvkeulen: Digest task or ToDo deck?

mjvkeulen: > Currently this would be a bit inconvenient for a course with only a `staff` group and students not in any group: for the assignment page's group filter, we exclude any groups that do not have journals in them.
> 
> Something similar could be done here as well: if there are no journals in the user's group, then include notifications for all journals?

This would be explicitly triggered by the user enabling group only notifications right? It would indeed not work well for the situation you outlined, but I would simply expect users in that scenario not to opt in for group only notifications. I feel introducing an exception here is not required.

src/django/test/test_notification.py
```diff
@@ -370,3 +370,67 @@ def test_error_during_digest_email_send(self):
...
+        assert not Notification.objects.filter(user=old_author).exists(), \
+            'Should not generate any notifications if the author is not in the active lti course'
+        assert Notification.objects.filter(user=new_author).exists(), \
+            'Should generate any notifications if the author is in the active lti course'
+
+    def test_jir_notification(self):
+        g_assignment = factory.Assignment(is_group_assignment=True)
+        g_journal = factory.GroupJournal(assignment=g_assignment, author_limit=5)
+        for _ in range(g_journal.author_limit - g_journal.authors.count()):
+            g_journal.authors.add(factory.AssignmentParticipation(assignment=g_assignment))
```
mjvkeulen, line 412: A `journal` is always initiated with a single author by default (applies to group and nongroup journals alike). 
Keeping that in mind adding 4 additional authors to a group journal could be done as:

g_journal = factory.GroupJournal(assignment=g_assignment, author_limit=5, add_users=[user2, user3, user4, user5])

src/django/VLE/models.py
```diff
@@ -653,7 +681,18 @@ def url(self):
         return gen_url(
             node=self.node, journal=self.journal, assignment=self.assignment, course=self.course, user=self.user)
 
+    def is_in_group(self):
```
mjvkeulen, line 684: This does not return a boolean, possibly add a docstring?

src/django/VLE/models.py
```diff
@@ -664,17 +703,31 @@ def save(self, *args, **kwargs):
...
+                    self.course = self.assignment.get_active_lti_course()
+                else:
+                    self.course = self.assignment.get_active_course(self.user)
+
+        # Should not create notifications for courses that the user cannot see
+        if not self.user.can_view(self.course):
+            return
+        # Should not create a notification as user only wants notification from within their group
+        # NOTE: it still sends notifications if there are no users in any groups
+        if self.type in self.OWN_GROUP_TYPES and self.user.preferences.group_only_notifications \
```
mjvkeulen, line 722: I would move this to a function for clarity, then the comment can be a docstring

src/django/VLE/models.py
```diff
@@ -2106,6 +2186,23 @@ def get_update_response(self):
...
 
+    def save(self, *args, **kwargs):
+        is_new = self._state.adding
+
+        super().save(*args, **kwargs)
+
+        if is_new:
+            if self.target and self.source and self.state == self.PENDING:
+                for user in VLE.permissions.get_supervisors_of(self.target):
+                    if user.has_permission('can_manage_journal_import_requests', self.target.assignment):
```
mjvkeulen, line 2197: Possibly move the nested if to a kwarg argument of get_supervisors_of ?

src/django/VLE/models.py
Larspolo: Was thinking about it as well, but couldnt come up with a good name, can you?

src/django/VLE/models.py
Larspolo: Added doctring, but now it also only returns the boolean

src/django/VLE/models.py
```diff
@@ -1029,6 +1084,33 @@ class Assignment(CreateUpdateModel):
...
 
+    def get_all_users(self, user=None, courses=None, journals_only=True):
+        """Get all users in an assignment
+
+        if user is set, only get the users that are in the courses that user is connected to as well
+        if courses is set, only get the users that are in the given courses
+        if journals_only is true, only get users that also have a journal (default: true)
+        """
+        if courses is None:
+            courses = self.courses.all()
```
mjvkeulen, line 1095: What is it that self.courses refers to?

src/django/VLE/models.py
```diff
@@ -1029,6 +1084,33 @@ class Assignment(CreateUpdateModel):
...
+        if user is not None:
+            courses = courses.filter(users=user)
+        if journals_only:
+            users = Journal.objects.filter(assignment=self).values('authors__user')
+        else:
+            users = self.assignmentparticipation_set.values('user')
+        return User.objects.filter(participations__in=courses, pk__in=users).distinct()
+
+    def get_own_group_users(self, user, courses=None, **kwargs):
```
mjvkeulen, line 1104: get_users_in_own_groups? I got confused, and this would fix group no longer being singular.

src/django/VLE/models.py
```diff
@@ -653,7 +681,20 @@ def url(self):
         return gen_url(
             node=self.node, journal=self.journal, assignment=self.assignment, course=self.course, user=self.user)
 
+    def is_in_group(self):
```
mjvkeulen, line 684: has_authors_in_own_groups?

src/django/VLE/models.py
```diff
@@ -1029,6 +1084,33 @@ class Assignment(CreateUpdateModel):
...
+        return User.objects.filter(participations__in=courses, pk__in=users).distinct()
+
+    def get_own_group_users(self, user, courses=None, **kwargs):
+        if courses is None:
+            courses = self.courses.all()
+        groups = user.participation_set.filter(course__in=courses).values('groups')
+        return self.get_all_users(
+            user=user, courses=courses, **kwargs).filter(participation__groups__in=groups).distinct()
+
+    def has_users_in_own_group(self, *args, **kwargs):
```
mjvkeulen, line 1111: This would then become has_users_in_own_groups

src/django/VLE/models.py
mjvkeulen: I find it tricky as well, possibly split up the logic?

- self.type in self.OWN_GROUP_TYPES and self.user.preferences.group_only_notifications -> `self.group_specific`
- self.assignment.has_users_in_own_groups(self.user) and not self.has_authors_in_own_groups() -> still tricky.

I suggested some other renames to make this click

mjvkeulen (approved):

src/vue/src/components/profile/NotificationCard.vue
```diff
@@ -177,6 +177,35 @@
                     />
                 </h2>
             </div>
+            <hr/>
+            <div class="clearfix">
+                <template v-if="$store.getters['preferences/saved'].group_only_notifications">
```
engelhamer, line 182: `v-if` can be moved to the button itself (no wrapper with single child needed)

src/vue/src/components/assets/RadioButton.vue
```diff
@@ -8,6 +8,7 @@
             @click="$emit('input', option.value)"
         >
             <icon :name="option.icon"/>
+            {{ option.text }}
```
engelhamer, line 11: This is now not needed anymore right?

engelhamer (approved):

PR #1021, opened 2020-09-16
Improved the performance of the assignment serializer

mjvkeulen:
Improved the performance of the Assignment page by focusing on the number of queries hitting the DB.

- New decorator `query_debugger` which shows the execution time and number of queries of the wrapped function.
- The first addition to the Django core checks, `check_missing_read_only_fields_serializers`. This warns developers when a serializer is created without specifying `read_only_fields` (read-only fields significantly affect serialization speeds). 
- Moved the journal serializer field `groups` to a journal computed property.
- `preset_db` command can be passed `n_performance_students` which is the number of students which will be generated for a performance course and assignment 

Some numbers, based on an assignment with a bit more than 400 participants:

**OLD**
Function : get_journals
Number of Queries : 1626
Finished in : 3.61s

**NEW**
Function : get_journals
Number of Queries : 3
Finished in : 0.11s

**OLD**
Function : get_stats
Number of Queries : 9
Finished in : 0.49s

**NEW**
Function : get_stats
Number of Queries : 4
Finished in : 0.14s

The result of this is a significant speedup to the assignment view (same config):

**OLD**
Function: retrieve
Number of Queries: 1652
Finished in: 3.81s

**NEW**
Function: retrieve
Number of Queries: 24
Finished in: 0.63s

The samples are rough and not averaged, yet this indicates a 600% speedup provided the assignment retrieve is still the bottleneck for the `assignment` page.


src/django/test/test_journal.py
```diff
@@ -295,6 +295,41 @@ def test_computed_grade(self):
...
+        assert g_journal.groups == [group2.pk, group.pk], 'Added group appears in the computed property for all authors'
+
+        student_course_participation.groups.remove(group)
+        g_journal.refresh_from_db()
+        assert g_journal.groups == [group2.pk], \
+            'Removing a student from a group also removes the group pk from the journal computed group property'
+
+        student2_ap.delete()
+        g_journal.refresh_from_db()
+        assert g_journal.groups == [], 'Removing a user from a journal updated the computed property'
```
Larspolo, line 332: Would like to see a test for if course gets removed, groups get updated

src/django/VLE/serializers.py
```diff
@@ -544,12 +561,11 @@ def get_usernames(self, journal):
     def get_groups(self, journal):
         if 'course' not in self.context:
             return None
-        return list(VLE.models.Participation.objects.filter(
-            user__in=journal.authors.values('user'),
-            course=self.context['course']).values_list('groups__pk', flat=True).distinct())
+        return journal.groups
```
Larspolo, line 564: This does change the functionality a bit right? Even if a course is given, it will now display all courses. Previously it would only display the groups from that course.

Also, I think this `SerializerMethodField` can be replaced by a normal field right?

Larspolo (approved):

src/django/test/test_journal.py
mjvkeulen: Added, also seems the computed field dependencies were not all needed. 
I have a suspicion dependencies like:
```
['node_set', ['entry']],
['node_set.entry', ['grade']],
```
Can be reduced to just `['node_set.entry', ['grade']]`

Larspolo: The other one also was there for if the user in an AP changes to another user. This should never happen anyway, so I think its fine to remove it

engelhamer (approved):

PR #1022, opened 2020-09-17
Improved the build process

mjvkeulen:
The build process is now more suitable for working with multiple simultaneous hosts.


engelhamer:
How about we remove the host files from this public repository and just leave a template?

engelhamer (approved):

PR #1024, opened 2020-09-19
Improved load times

mjvkeulen:
# Reduced the overall bundle size.

To provide some insight into the significance of the changes, here is an example from the perspective of `/Home` with cache disabled:

**OLD**
image

**NEW**
image

The red `load` time is the most interesting for the user experience, mostly impacted the initial `resources` size. Which for the example of `/Home` is reduced by ~69%.

## Major bundle size culprits (> 1 MB)
- Tinymce
- PDF
- Excel

We are sort of stuck with these, although some more lightweight variants exist, something to keep in mind.

# Some new Makefile targets
- `make output-webpack-config`: will generate the actual webpack configuration generated by the Vue-CLI, `src/vue/webpack_config_output.js`, ignored by git.
- `make output-vue-build-report`: generates an HTML file showing the distribution of the frontend bundle size: `src/vue/dist/report.html`
- `make update-submodules`: updates all submodules to the latest commits of the default branches.

# CSS variables
- CSS variables can be made globally available in all components style tags, I've added `colors.sass` and `breakpoints.sass`. 
- Use sass variables in javascript files. `import themeColors from 'sass/modules/colors.sass'` -> themeColors.darkblue, I've opted to use non kebab case here, else the object access syntax would be different.
The bundle size impact was not too high ~ 10KB.


src/vue/src/helpers/bootstrap.js
```diff
@@ -0,0 +1,109 @@
...
+    // Vue.use(PopoverPlugin)
+    Vue.use(ProgressPlugin)
+    // Vue.use(SidebarPlugin)
+    // Vue.use(SpinnerPlugin)
+    Vue.use(TablePlugin)
+    // Vue.use(TabsPlugin)
+    // Vue.use(TimePlugin)
+    // Vue.use(ToastPlugin)
+    Vue.use(TooltipPlugin)
+}
```
Larspolo, line 109: What was the method of picking these? Do we need to test the whole application to make sure you didnt miss any, or are you 99+% certain that this is correct?

src/vue/src/views/Assignment.vue
```diff
@@ -449,6 +448,9 @@ import participationAPI from '@/api/participation.js'
 import journalAPI from '@/api/journal.js'
 import { mapGetters, mapMutations } from 'vuex'
 
+const AssignmentSpreadsheetExport = () => import(
+    /* webpackChunkName: 'assignment-spreadsheet-export' */ '@/components/assignment/AssignmentSpreadsheetExport.vue')
```
Larspolo, line 452: Why is this not done in the components export like you do for the others?

src/vue/vue.config.js
```diff
@@ -3,8 +3,52 @@ const webpack = require('webpack') // eslint-disable-line import/no-extraneous-d
...
+
 module.exports = {
+    chainWebpack: config => {
+        if (config.plugins.has('prefetch')) {
+            config.plugin('prefetch').tap(options => {
+                options[0].fileBlacklist = options.fileBlacklist || []
+
+                options[0].fileBlacklist.push(/.+?\.map$/)
+                CHUNKS_WHICH_SHOULD_NOT_BE_PREFETCHED.forEach((chunkName) => {
+                    options[0].fileBlacklist.push(new RegExp(`.*${chunkName}.+?\.js$`))
```
Larspolo, line 23: Did you get this from somewhere, or created yourself? The regex seems to be a bit too inclusive. Wouldnt now also files like `notpdf.js` and `introduction.js` be excluded?

src/vue/vue.config.js
```diff
@@ -3,8 +3,52 @@ const webpack = require('webpack') // eslint-disable-line import/no-extraneous-d
...
+    'draggable',
+    'croppa',
+    'intro',
+]
+
 module.exports = {
+    chainWebpack: config => {
+        if (config.plugins.has('prefetch')) {
+            config.plugin('prefetch').tap(options => {
+                options[0].fileBlacklist = options.fileBlacklist || []
```
Larspolo, line 19: Would have expected this to be `options[0].fileBlacklist = options[0].fileBlacklist || []`

src/vue/package.json
```diff
@@ -8,51 +8,48 @@
...
+    "vue": "^2.6.11",
+    "vue-router": "^3.1.6",
+    "vuex": "^3.1.3",
     "@sentry/browser": "^5.14.2",
     "@sentry/integrations": "^5.14.2",
     "axios": "^0.18.1",
     "bootstrap": "^4.4.1",
     "bootstrap-vue": "^2.16.0",
-    "core-js": "^3.6.4",
+    "jquery": "^3.4.1",
```
Larspolo, line 24: Do we know where and why jquery is used? When I removed it, frontend still seemed to work (even after reinstalling node modules)

src/vue/src/views/Assignment.vue
mjvkeulen: Line length, there is no further difference.

src/vue/package.json
mjvkeulen: Bootstrap can throw warnings, but this might be changed with the more recent versions. Not used is shedded regardless, but I am all in favor of further cleaning.

Makefile
```diff
@@ -115,6 +115,15 @@ setup-sentry-cli:
 		${venv_activate} && curl -sL https://sentry.io/get-cli/ | bash; \
 	fi
 
+update-submodules:
+	git submodule update --remote --merge
```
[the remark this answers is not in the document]
mjvkeulen: Makes sense, non CI version.

src/vue/src/helpers/bootstrap.js
mjvkeulen: For any Vue components missing errors will be thrown, e.g. `b-card`. I also opted to use plugins over raw component imports as it's less error-prone (all dependencies come bundled with the plugin). So for these I am quite sure, to verify just navigate the app with the inspector open.

The CSS imports are less obvious and require visual inspection, but I erred on the side of caution and feel pretty confident.

src/vue/vue.config.js
mjvkeulen: They would be excluded from `prefetching` yes, which in itself is not breaking.

Chunknames are based on the package name their vendor gave them, see optimization webpack config. Feel free to modify the regex as you see fit. Just execute the `output-vue-build-report` target for an overview of the generated bundles.

src/vue/src/views/Assignment.vue
Larspolo: Figured

engelhamer (approved):

src/vue/src/sass/vendor/bootstrap.sass
```diff
@@ -0,0 +1,43 @@
+/* Required */
+@import '~bootstrap/scss/functions'
+@import '~bootstrap/scss/variables'
+@import '~bootstrap/scss/mixins'
+
+/* Based on use */
```
engelhamer, line 6: These selective imports have no significant effect on load times from what I have noticed: only ~.03s which could even be measurement errors. Maybe a full import would still be simpler here.

mjvkeulen: My notes say 0.15MB for bootstrap JS and 60KB for CSS

PR #1028, opened 2020-10-01
Added an admin panel for management tasks

engelhamer:
Created an admin panel for superusers. It contains the following tabs:
- Instance settings 
image
- Course overview 
image
- User overview 
image
- Invite users 
image

Also performed various front-end improvements. (such as updated bootstrap). More will follow in a separate PR.

**Fixed issues**:
- Fixed Z#1081
- Fixed Z#1071
- Fixed Z#1072
- Fixed Z#1073


src/vue/src/views/AdminPanel.vue
```diff
@@ -0,0 +1,97 @@
...
+            errorLogs: '',
+            requestInFlight: false,
+        }
+    },
+    methods: {
+        inviteUsers () {
+            this.requestInFlight = true
+            adminAPI.inviteUsers({
+                users: [
+                    {
```
Larspolo, line 79: To be removed

Larspolo (commented):
I still see a lot of frontend changes as well

src/vue/src/components/admin/UserOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+        this.getAllUsers()
+    },
+    methods: {
+        getAllUsers () {
+            adminAPI.getAllUsers()
+                .then((users) => { this.users = users })
+        },
+        removeUser (user) {
+            if (window.confirm(`Are you sure you want to remove ${user.full_name}? All of their work will be deleted.`
+                + 'This includes courses which they are the author of. This cannot be undone!')) {
```
mjvkeulen, line 122: I didn't even know this. Propose we change this to including course without remaining administrative users or similar?

src/vue/src/components/admin/UserOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+                adminAPI.removeUser(user.id)
+                    .then(() => { this.$toasted.success('Successfully removed user.') })
+                    .finally(() => {
+                        this.getAllUsers()
```
mjvkeulen, line 126: Is this for updating the user list after a successful removal?

src/vue/src/components/assets/CodeVersionAlert.vue
```diff
@@ -2,10 +2,10 @@
     <b-alert
         id="code-version-alert"
         :show="show"
-        class="cursor-pointer unselectable"
+        class="theme-shadow cursor-pointer unselectable"
         @click.native="beginRefreshForNewVersion"
     >
-        eJournal received an update 🎉, click to refresh.
+        eJournal received an update 🎉, click <b>here</b> to reload
```
mjvkeulen, line 8: The alert itself is clickable, not the `here` text. Could this not result in users specifically attempting to click the `here` text?

src/vue/src/components/assets/NumberBadge.vue
```diff
@@ -101,20 +91,4 @@ export default {
     position: absolute
     right: 0px
     top: 0px
-
-.number-badge-container
```
mjvkeulen, line 105: Curious to see how these look now

src/vue/src/sass/global.sass
```diff
@@ -19,12 +21,10 @@
 @import '~sass/partials/select.sass'
 @import '~sass/partials/tours.sass'
 
-@import '~sass/vendor/fonts.sass'
-
 body
     color: $theme-dark-blue
     /* Background pattern from Toptal Subtle Patterns */
-    background-image: url(/background.png)
+    // background-image: url(/background.png)
```
mjvkeulen, line 27: tag

src/vue/src/sass/partials/cards.sass
```diff
@@ -5,14 +5,15 @@
...
+    border-width: 2px 2px 2px 2px
     border-color: $theme-dark-grey
     transition: all 0.3s cubic-bezier(.25,.8,.25,1) !important
     border-radius: 5px !important
-    .card-body
+    &:not(.no-left-border) .card-body
         position: relative
         border-radius: 3px !important
-        border-left: 8px solid $theme-blue
+        // QUESTION: How about only doing this for course, assignment and journal cards?
```
mjvkeulen, line 15: Tagging open question, will come back to this.

src/vue/src/sass/partials/columns.sass
```diff
@@ -1,14 +1,14 @@
 .left-content
     padding: 30px 30px 10px 30px !important
     @include xl
-        height: calc(100vh - 70px)
+        height: calc(100vh - 50px)
```
mjvkeulen, line 4: The 50px seems to be recurring often, time to move it to a variable?

src/vue/src/views/AdminPanel.vue
```diff
@@ -0,0 +1,97 @@
...
+
+<script>
+import WideContent from '@/components/columns/WideContent.vue'
+import BreadCrumb from '@/components/assets/BreadCrumb.vue'
+import EditInstance from '@/components/admin/EditInstance.vue'
+import UserOverview from '@/components/admin/UserOverview.vue'
+import InviteUsers from '@/components/admin/InviteUsers.vue'
+
+
+// import courseAPI from '@/api/course.js'
```
mjvkeulen, line 56: tag

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,151 @@
...
+    def invite_users(self, request):
+        """Invite new users to eJournal.
+
+        This will create accounts for the invited users, which they can activate through an invite link.
+
+        Arguments:
+        TODO
+        """
+        if not request.user.is_superuser:
+            return response.forbidden('You are not allowed to invite new users.')
```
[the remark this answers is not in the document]
engelhamer: Indeed an issue would be fine. This level of configurability is an additional feature.

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,151 @@
...
+        if existing_emails:
+            return response.bad_request({'existing_emails': existing_emails})
+        invalid_emails = [user['email'] for user in users]
+        if existing_emails:
+            return response.bad_request({'existing_emails': existing_emails})
+
+        created_user_ids = []
+        try:
+            for user in users:
+                created_user = factory.make_user(username=user['username'], email=user['email'],
```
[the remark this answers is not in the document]
engelhamer: The `make_user` fuction uses the `set_unusable_password` function for inactive (yet to be activated) account like these.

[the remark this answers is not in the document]
engelhamer: As of the newest changes this is updated to bulk create. We do have to consider the error catch now: do you know if the bulk create could fail still even if the `full_clean` is called on the individual users to be created?

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,151 @@
...
+                                                 full_name=user['full_name'], is_teacher=user['is_teacher'],
+                                                 is_active=False)
+                created_user_ids.append(created_user.id)
+                print(created_user_ids)
+        except Exception as exception:
+            capture_message('Something went wrong while inviting users: {}'.format(exception), level='error')
+            for user in created_user_ids:
+                User.objects.filter(pk__in=created_user_ids).delete()
+        else:
+            for user_id in created_user_ids:
```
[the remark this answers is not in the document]
engelhamer: Moved to background job. Do you think we should do additional crash handling there? This used to be fine becauses they were separate threads, but if something goes wrong now we should maybe receive a notification and keep running for the other users.

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -32,7 +32,7 @@
                 class="template-availability"
             >
                 <b-button
-                    class="delete-button"
```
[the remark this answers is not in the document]
engelhamer: The reason for moving away is that the 'semantic' name was invalid in may cases. E.g.: saving a grade that already existed was done using an `add-button` rather than a `change-button`. The opposite was the case for setting bonus points via the bulk import method.

An orange bulk import seems intuitive rather than a green one, whereas a green save button is intuitive rather than an orange one. This is why I prefered to get the color names instead of this 'semantic' name.

Also this is consitstent with the other color classes for e.g. `text-` and `background-`.

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if not request.user.is_superuser:
+            return response.forbidden('You are not allowed to invite new users.')
+
+        users, = utils.required_params(request.data, 'users')
+
+        # Ensure a username and email are specified for all users to be invited.
+        if any([not user['full_name'] for user in users]):
+            return response.bad_request('Please specify a full name for all users.')
+        if any([not user['username'] for user in users]):
```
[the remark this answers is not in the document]
engelhamer: The user can see this in the front end by the empty fields in that column. We do not have front-end field validation highlights, see [Z#1093](https://zube.io/ejourn-al/ejournal/c/1093).

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,175 @@
...
+            return response.bad_request('Please specify a username for all users.')
+        if any([not user['email'] for user in users]):
```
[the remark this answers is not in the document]
engelhamer: Afaik we do not currently use serializers for creation? Would prefer to tackle that in bulk for consistency: it should then always be reliable to use serializers for creation instead of having to check still for which models / serializers you can rely on this.

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if not request.user.is_superuser:
+            return response.forbidden('You are not allowed to remove users.')
+        if str(request.user.pk) == pk:
+            return response.forbidden('You are not allowed to delete your own account.')
+
+        try:
+            User.objects.get(pk=pk).delete()
+        except User.DoesNotExist:
+            return response.not_found('User to delete does not exist.')
```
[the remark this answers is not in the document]
engelhamer: There is in fact something going wrong.

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if not request.user.is_superuser:
+            return response.forbidden('You are not allowed to get all users.')
+
+        users = UserOverviewSerializer(User.objects.all().order_by('full_name'), context={'user': request.user},
+            many=True).data
+        return response.success({'users': users})
+
+    @action(methods=['post'], detail=True)
+    def remove_user(self, request, pk):
```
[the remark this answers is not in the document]
engelhamer: Improved structure to adhere CRUD and REST framework.

src/vue/src/components/admin/UserOverview.vue
engelhamer: Actually it might not be true now that I come to check.

src/vue/src/components/admin/UserOverview.vue
engelhamer: Yes, I would like to always keep this updated. Also when pagination is added, we cannot just perform the update in the front end manually (since we don't know what the next page will be),

src/vue/src/components/assets/CodeVersionAlert.vue
engelhamer: Then that is no problem. Click to refresh may  be interpreted as 'click anywhere', Not necessarily within this popup itself.

src/vue/src/sass/partials/cards.sass
engelhamer: Moved to UI changes PR.

src/django/VLE/views/admin.py
```diff
@@ -0,0 +1,133 @@
...
+
+import VLE.factory as factory
+import VLE.utils.generic_utils as utils
+import VLE.utils.responses as response
+from VLE.models import Instance, User
+from VLE.serializers import InstanceSerializer, UserOverviewSerializer
+from VLE.tasks import send_invite_email
+
+
+class AdminView(viewsets.ViewSet):
```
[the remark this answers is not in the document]
engelhamer: That is indeed more sensible. All functions have been moved.

src/vue/src/sass/partials/columns.sass
engelhamer: That is the (new) header height. Indeed a variable might be alright, but we'll have to come up with some formula's to scale the elements along (since otherwise changing the 50px itself also required additional changes anyways).

src/django/VLE/serializers.py
```diff
@@ -79,6 +79,13 @@ def get_groups(self, user):
             return None
 
 
+class UserOverviewSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = VLE.models.User
+        fields = ('username', 'full_name', 'email', 'is_teacher', 'is_active', 'id')
+        read_only_fields = fields
```
mjvkeulen, line 86: This serializer does not make use of much of its possible functionality. The view could also simply return values_list(fields)

src/vue/src/components/admin/UserOverview.vue
mjvkeulen: Possibly move it into the `then` block, the state should not change if the request fails.

src/vue/src/components/assets/ResetWrapper.vue
```diff
@@ -25,6 +25,7 @@ export default {
 .reset-wrapper
     position: relative
     .reset-button
+        // TODO: this is broken
```
mjvkeulen, line 28: tag

src/vue/src/sass/partials/columns.sass
mjvkeulen: Valid point, however, without the formula's we will still have the clarity of a named variable.

src/vue/src/views/AdminPanel.vue
```diff
@@ -0,0 +1,72 @@
...
+</template>
+
+<script>
+import WideContent from '@/components/columns/WideContent.vue'
+import BreadCrumb from '@/components/assets/BreadCrumb.vue'
+import EditInstance from '@/components/admin/EditInstance.vue'
+import UserOverview from '@/components/admin/UserOverview.vue'
+import InviteUsers from '@/components/admin/InviteUsers.vue'
+
+
```
mjvkeulen, line 55: hspace

src/django/VLE/factory.py
```diff
@@ -30,16 +30,18 @@ def make_user(username, password=None, email=None, lti_id=None, profile_picture=
...
+    if is_test_student or not is_active:
         user.set_unusable_password()
     else:
         validators.validate_password(password)
         user.set_password(password)
 
     user.full_clean()
-    user.save()
+
+    if save:
```
mjvkeulen, line 43: Having a conditional save in a module dedicated to creating model instances is a bit counter-intuitive. 

I propose moving the validation block to a User class method.

engelhamer: Not if it's used for bulk creation, and `False` by default?

src/django/VLE/views/user.py
```diff
@@ -240,11 +246,9 @@ def destroy(self, request, pk):
         On success:
             success -- deleted message
         """
-        if int(pk) == 0:
-            pk = request.user.id
         user = User.objects.get(pk=pk)
 
-        if not (request.user.is_superuser or request.user == user):
+        if not (request.user.is_superuser):
```
Larspolo, line 251: brakets can be removed

src/django/VLE/views/user.py
```diff
@@ -361,3 +365,65 @@ def get_permissions(self):
...
+            users_to_create = []
+            for user in users:
+                user = factory.make_user(username=user['username'].lower(), email=user['email'].lower(),
+                                         full_name=user['full_name'], is_teacher=user['is_teacher'], is_active=False,
+                                         save=False)
+                users_to_create.append(user)
+            created_user_ids = [user.id for user in User.objects.bulk_create(users_to_create)]
+        except Exception as exception:
+            capture_message('Something went wrong while inviting users: {}'.format(exception), level='error')
+            User.objects.filter(pk__in=created_user_ids).delete()
```
Larspolo, line 424: created_user_ids will only be set at the end, and thus this wont delete anything if for loop crashes. Would suggest moving the `users_to_create` outside of the try scope, and use `[u.pk for u in users_to_create]`

src/django/VLE/views/user.py
```diff
@@ -361,3 +365,65 @@ def get_permissions(self):
...
+            return response.bad_request('An error occured while creating new users.')
+        else:
+            send_invite_emails.delay(created_user_ids)
```
Larspolo, line 427: except already returns sth, else is not needed (now it is not consistent with the success return)

src/vue/src/components/admin/EditInstance.vue
```diff
@@ -0,0 +1,83 @@
...
+                this.instance = instance
+            })
+    },
+    methods: {
+        saveChanges (changes) {
+            this.saveRequestInFlight = true
+            instanceAPI.update(changes, {
+                customSuccessToast: 'Successfully updated instance details.',
+            })
+                .finally(() => {
```
Larspolo, line 77: Would also suggest a `then((instance) => { this.instance = instance }` in case the server did sth while updating (example: strip whitespace from name (maybe this should be added s well now I type it, or move to issue?)), it changes back to what the server responded

src/django/VLE/views/admin.py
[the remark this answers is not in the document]
Larspolo: I cannot say anything for sure about this

src/vue/src/components/admin/UserOverview.vue
```diff
@@ -0,0 +1,163 @@
...
+                userAPI.delete(user.id)
+                    .then(() => { this.$toasted.success('Successfully removed user.') })
+                    .finally(() => {
+                        this.getAllUsers()
+                    })
+            }
+        },
+        makeTeacher (user) {
+            userAPI.update(user.id, { is_teacher: true })
+                .then(() => { this.$toasted.success(`Successfully made ${user.full_name} teacher.`) })
```
Larspolo, line 141: I personally like having response messages in the backend. Can be combined with the `responseSuccessToast: true` in frontend

src/django/VLE/views/instance.py
```diff
@@ -26,11 +26,9 @@ def retrieve(self, request, pk=0):
 
     def partial_update(self, request, *args, **kwargs):
         """Update instance details.
-
         Arguments:
         request -- request data
             data -- the new data for the journal
-
```
Larspolo, line 33: This is now inconsistent with the style, also the docstring apperently already was wrong: `data -- the new data for the journal` shoudl be `data -- the new data for the instance`

Larspolo (changes requested):
image
When the options are opened, the bottom gets a couple pixel margin


image
Two error messages shown, would suggest only using the message that the backend sends

image
Would suggest clearing the error output also when new errors occure, now it can be outdated

image
When there is whitespace, it gives a weird error, the whitespace is even imported from the excel file. So if there is a whitespace there, it can be really shitty debugging it

src/vue/src/components/admin/UserOverview.vue
engelhamer: It could be that the request failed because some other admin already deleted the user. In that case we should update.

src/django/VLE/serializers.py
engelhamer: No? There are much fewer fields included here. E.g. lti id, verified email, test student status etc.

src/django/VLE/views/user.py
engelhamer: That's true, but users to create do not have a pk since they are not created yet?

src/vue/src/components/admin/UserOverview.vue
engelhamer: This now uses the `User` update method (according to your request) which can be used to update a whole variety of things. I would not propose adding all of those as individual messages in the back end for readability.

src/django/VLE/views/instance.py
engelhamer: Went wrong while reverting the copy.

engelhamer:
> image
> When the options are opened, the bottom gets a couple pixel margin
> 

Seems to be a Firefox-only bug. Have not been able to fix yet.

> image
> Two error messages shown, would suggest only using the message that the backend sends

If the back end sends a message we show that. If the back end sends an object (error log) it is shown separately in the view itself and a separate error is shown front end only.

> image
> When there is whitespace, it gives a weird error, the whitespace is even imported from the excel file. So if there is a whitespace there, it can be really shitty debugging it

All strings are now trimmed. If a row is present with only falsy values after trimming, it is discarded,

src/vue/src/sass/partials/columns.sass
engelhamer: Added it in the breakpoints sass file. Kind of an awkward position, but global usage of sass vars is kind of finicky, couldn't make it accessible just from the global sass file.

src/django/VLE/views/admin.py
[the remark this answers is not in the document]
mjvkeulen: I believe full_clean also performs unique contraint checks, I think it will be reasonable to assume bulk create should work

src/django/VLE/serializers.py
mjvkeulen: I think you misunderstood, this serializer doesn't really add anything. If you just return
response.success(x: User.objects.filter(xxx).values_list('username', 'full_name', 'email', 'is_teacher', 'is_active', 'id')) we could just omit this serializer.

src/django/VLE/factory.py
mjvkeulen: I was dealing with this myself in other areas where we require speedups. We need to move toward separating validation, cleaning, and updating/creation, otherwise it's difficult to properly encapsulate application-level constraints or state guarantees. Forcing us to repeat logic in multiple places for each scenario which ignores the default model save (update, bulk_creation, etc).

Leave this solution as it is for now. We will need to agree on an approach but there will be other PR's more suited to this discussion.

src/django/VLE/views/user.py
Larspolo: Owww ofcourse,missed that. Good enough then

mjvkeulen (commented):
Admin panel requires some user search functionality

Larspolo (changes requested):
image


image
Its possible to invite users with an empty username

Same goes for whitespace before a username. This will result in a user that will never be able to be logged in as we remove whitespace on login request

src/django/test/test_course.py
```diff
@@ -57,19 +57,25 @@ def test_course_factory(self):
...
         get_resp = api.get(self, 'courses', user=self.teacher1)['courses']
-        assert len(get_resp) == 2, 'The teacher did not get all the courses it is the author of'
+        assert len(get_resp) == 2, 'Teacher should get all courses they are the author of'
         assert {self.course1.pk, self.course2.pk} == set([c['id'] for c in get_resp]), \
-            'The teacher did not get all the courses it is the author of'
+            'Teacher should get all courses they are the author of'
 
         get_resp = api.get(self, 'courses', user=self.teacher2)['courses']
-        assert len(get_resp) == 2, 'The teacher did not get all the courses it is the author of or is participating in'
+        assert len(get_resp) == 2, 'Teacher should get all courses they are the author of or ar participating in'
```
mjvkeulen, line 70: are

src/django/test/test_course.py
```diff
@@ -57,19 +57,25 @@ def test_course_factory(self):
...
         assert is_response(get_resp, self.course2, self.course3), \
-            'The teacher did not get all the courses it is the author of'
+            'Teacher should get all courses they are the author of or ar participating in'
```
mjvkeulen, line 72: ditto

src/django/test/test_email.py
```diff
@@ -142,22 +161,51 @@ def test_request_email_verification(self):
...
+        assert mail.outbox[1].subject == 'Additional support info: topic', \
+            'Feedback topic should be in developer support email title'
+        assert 'actual_feedback_content_here' in mail.outbox[1].body, \
+            'Feedback should be repeated in developer support email'
+        assert 'ftype' in mail.outbox[1].body, \
+            'Feedback type should be in developer support email'
+        assert 'someBrowser' in mail.outbox[1].body, \
+            'User agent should be in developer support email'
+        assert 'current_user_url.com' in mail.outbox[1].body, \
+            'Current URL should be in developer support email'
```
mjvkeulen, line 204: Nice strict email content checks

src/django/test/test_email.py
```diff
@@ -142,22 +161,51 @@ def test_request_email_verification(self):
...
+
+        assert all(mail.outbox[i].from_email == f'eJournal | Support<support@{settings.EMAIL_SENDER_DOMAIN}>' for i in
```
mjvkeulen, line 206: Since we are now also strict with checking for the correct sender name, it's time to move these to a setting's variable. Prob best to do so in a format where the specific email names are linked to their respective sender names.

src/django/test/test_user.py
```diff
@@ -498,3 +500,98 @@ def test_prefetch_course_groups(self):
...
+    def test_invite_users(self):
+        admin = factory.Admin()
+        pre_invite_user_count = User.objects.count()
+        users = [
+            {
+                'full_name': f'User Full Name {i}',
+                'username': f'user{i}username',
+                'email': f'user{i}mail@ejournal.app',
+            } for i in range(10)
+        ]
```
mjvkeulen, line 514: Using the user factory would improve maintainability.

```
users = [factory.Student.build() for _ in range(10)]
params = [{'full_name': u.full_name, 'username': u.username, 'email': u.email} for u in users]
```

src/django/test/test_user.py
```diff
@@ -498,3 +500,98 @@ def test_prefetch_course_groups(self):
                 values, prefetched = prefetched_objects(user, 'participation_set')
                 participation = values[0]  # Test is setup so there no need to check if prefetched with alternative qry
                 assert list(participation.groups.all()) == [group], 'All groups are prefetched for all users'
+
+    @override_settings(EMAIL_BACKEND='anymail.backends.test.EmailBackend', CELERY_TASK_ALWAYS_EAGER=True)
+    def test_invite_users(self):
```
mjvkeulen, line 505: Good test, but quite long. Consider splitting or using nested functions? Possibly validation, creation, validation link etc?

src/vue/src/components/admin/InviteUsers.vue
```diff
@@ -0,0 +1,347 @@
...
+                is_teacher: false,
+            }],
+            errorLogs: null,
+            requestInFlight: false,
+        }
+    },
+    computed: {
+        usersToInviteFiltered () {
+            // Filter fully empty rows.
+            return this.usersToInvite.filter(user => !(!user.full_name && !user.username && !user.email))
```
mjvkeulen, line 254: not not not not to or or or?

src/vue/src/sass/partials/cards.sass
mjvkeulen: Ok, prob best to remove the question here so it will be a new question on that branch.

mjvkeulen (changes requested):
Atm loading home as a teacher gives me a 403 "You are not allowed to get all courses."

Larspolo:
> Atm loading home as a teacher gives me a 403 "You are not allowed to get all courses."

See my comment on how to solve that

src/django/VLE/views/user.py
```diff
@@ -66,7 +67,9 @@ def list(self, request):
         if not request.user.is_superuser:
             return response.forbidden('Only administrators are allowed to request all user data.')
 
-        return response.success({'users': list(User.objects.values(*UserSerializer.Meta.user_fields))})
+        serializer = UserOverviewSerializer(User.objects.all().order_by('full_name'), context={'user': request.user},
```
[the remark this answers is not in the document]
engelhamer: `.values` does not return a dict. So then that would also have to be made here for each user.

src/django/test/test_user.py
engelhamer: Why? That would still keep practically the same list comprehensions of dicts on the second line. Also now whitespace has been added.

src/django/test/test_email.py
engelhamer: > Prob best to do so in a format where the specific email names are linked to their respective sender names

I do not understand what you mean by this?

src/django/VLE/views/user.py
```diff
@@ -366,3 +371,79 @@ def get_permissions(self):
...
+            for user in users:
+                user = factory.make_user(username=user['username'].lower(), email=user['email'].lower(),
+                                         full_name=user['full_name'],
+                                         is_teacher='is_teacher' in user and user['is_teacher'], is_active=False,
+                                         save=False)
+                users_to_create.append(user)
+            created_user_ids = [user.id for user in User.objects.bulk_create(users_to_create)]
+            send_invite_emails.delay(created_user_ids)
+        except Exception as exception:
+            capture_message('Something went wrong while inviting users: {}'.format(exception), level='error')
```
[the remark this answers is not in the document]
engelhamer: > > Atm loading home as a teacher gives me a 403 "You are not allowed to get all courses."
> 
> See my comment on how to solve that

That solution limited functionality. Consider e.g. the case where a superuser also has a course they teach.

engelhamer:
> > Atm loading home as a teacher gives me a 403 "You are not allowed to get all courses."
> 
> See my comment on how to solve that

That solution limited functionality. Consider e.g. the case where a superuser also has a course they teach.

src/django/test/test_user.py
engelhamer: Fine by me

src/django/test/test_email.py
mjvkeulen: Map the corresponding domain prefix to the desired 'from' sender name in the data structure you pick?

src/django/test/test_user.py
mjvkeulen: The logic of how the desired fields should be formatted remains in one central place. The user factory.

mjvkeulen: Why did you opt for adding white space to the base case over introducing whitespace where it is specifically tested? By introducing whitespace as the default the test loses some strength for non-white space cases.

src/django/VLE/views/user.py
```diff
@@ -366,3 +371,89 @@ def get_permissions(self):
...
+                    is_active=False,
+                    save=False
+                )
+                for user in users
+            ]
+            created_user_ids = [user.id for user in User.objects.bulk_create(users_to_create)]
+            send_invite_emails.delay(created_user_ids)
+        except Exception as e:
+            capture_exception(e)
+            User.objects.filter(pk__in=created_user_ids).delete()
```
mjvkeulen, line 456: The try-catch undo structure is redundant. The bulk_create occurs within a single transaction which will either be successful or not (if not state will be reverted to the beginning of the transaction).

src/vue/src/sass/partials/columns.sass
mjvkeulen: Since it is indeed not a breakpoint, how about a new .sass module? E.g. `dimensions.sass`? 

We can then make that module globally available in `vue.config.js` by adding it to the sass loader options configuration:
`@import "~sass/modules/dimensions.sass"`

src/django/test/test_email.py
mjvkeulen: E.g. in `base.py` in the settings module.

```
@dataclass
class EmailToSender:
    label: str

    @property
    def email(self) -> str:
        return f'{self.label}@{EMAIL_SENDER_DOMAIN}'

    @property
    def sender(self) -> str:
        name = self.email.split('@')[0].capitalize()
        return f'eJournal | {name}'


class Emails:
    support = EmailToSender('support')
    info = support = EmailToSender('info')
    contact = support = EmailToSender('contact')
```

Which would allow a bit more clean access, e.g. `Emails.support.sender`. Or just go 

```
EMAIL_SENDER_NAME_MAP = {
    'support': {'email': f'support@{EMAIL_SENDER_DOMAIN}', 'sender': 'eJournal | Support'}
}
```

I just dislike the access pattern, e.g. `EMAIL_SENDER_NAME_MAP['support']['email']`

engelhamer: I kind of think that this is overengineering as there is really only one file where we actually send mails?

engelhamer: But feel free to add whatever you like.

mjvkeulen: I don't care what structure you pick, but the sender names should be moved to a setting

src/django/VLE/views/user.py
mjvkeulen: You have already taken the steps to ensure the data is unique and available, how about just removing the try catch and error log?

How do you perceive an exception will occur?

mjvkeulen:
image
Minor: possibly introduce a minimum width for the column headers?

engelhamer:
> image
> Minor: possibly introduce a minimum width for the column headers?
> 

So with overflow scroll?

src/django/VLE/views/user.py
```diff
@@ -366,3 +371,88 @@ def get_permissions(self):
...
+                    username=user['username'],
+                    email=user['email'],
+                    full_name=user['full_name'],
+                    is_teacher='is_teacher' in user and user['is_teacher'],
+                    is_active=False,
+                    save=False
+                )
+                for user in users
+            ]
+            created_user_ids = [user.id for user in User.objects.bulk_create(users_to_create)]
```
mjvkeulen, line 452: Users created via this method end up without a preferences table, and as such can't use the platform (or even login).

I propose to define our own `bulk_create` method on the extended user manager (`UserManagerExtended`). Where within one checkpoint/transaction we bulk_create both the desired users and their corresponding preferences. 

Initialization of the user instances can be left as is (by the user factory, rewrite can be done another time). 
Validation is already cleanly handled and imo should remain outside of the manager logic.

mjvkeulen (changes requested):

mjvkeulen:
image
Minor: I would expect even division of the tabs for all sizes?

engelhamer:
> image
> Minor: I would expect even division of the tabs for all sizes?

This is default bootstrap pill tab styling. Want to change that?

src/django/test/test_course.py
```diff
@@ -57,19 +57,25 @@ def test_course_factory(self):
...
+        assert len(get_resp) == 2, 'Teacher should get all courses they are the author of or are participating in'
         assert is_response(get_resp, self.course2, self.course3), \
-            'The teacher did not get all the courses it is the author of'
+            'Teacher should get all courses they are the author of or are participating in'
+
+        get_resp = api.get(self, 'courses', params={'get_all': True}, user=self.admin)['courses']
+        assert len(get_resp) == Course.objects.count(), 'Superuser can get all courses on the instance'
+
+        api.get(self, 'courses', params={'get_all': True}, user=self.teacher1, status=403)
+        api.get(self, 'courses', params={'get_all': True}, user=self.student, status=403)
```
Larspolo, line 78: Would prefer this to be tested now with the strings `"true"` and `"false"` to replicate the real deal

Larspolo (approved):
Think we already spoke about this, but I would really like class name changes to get their own PR. Now im reviewing "137"  where probably not even 20% of them are actually useful

src/django/VLE/tasks/email.py
```diff
@@ -100,6 +100,40 @@ def send_email_verification_link(user_pk):
...
+    """Sends an invite link to join the platform to the users email adress."""
+    users = VLE.models.User.objects.filter(pk__in=user_pks)
+    instance_name = VLE.models.Instance.objects.get_or_create(pk=1)[0].name
+    token_generator = PasswordResetTokenGenerator()
+    for user in users:
+        email_data = {}
+        email_data['heading'] = 'Welcome to eJournal!'
+        token = token_generator.make_token(user)
+        email_data['main_content'] = """
+        You have been invited to eJournal by {}. Please click the button below to activate your account.""" \
```
mjvkeulen, line 114: image
I believe the chance this phrasing will clash is relatively high, e.g. for the UvA this would be:
`You have been invited to eJournal by eJournal - UvA`

How about only using the instance name?

src/django/VLE/tasks/email.py
```diff
@@ -100,6 +100,40 @@ def send_email_verification_link(user_pk):
...
+        email_data['extra_content'] = 'Username: {}'.format(user.username)
+        email_data['button_url'] = '{}/SetPassword/{}/{}?new_user=true'.format(settings.BASELINK, user.username, token)
+        email_data['button_text'] = 'Activate account'
+        email_data['profile_url'] = '{}/Profile'.format(settings.BASELINK)
+
+        html_content = render_to_string('call_to_action.html', {'email_data': email_data})
+        text_content = strip_tags(html_content)
+
+        email = EmailMultiAlternatives(
+            subject='Complete your registration for eJournal at {}'.format(instance_name),
```
mjvkeulen, line 126: A similar clash here:
image

src/django/test/test_email.py
mjvkeulen: Let's use the new `EMAILS` variable here as well?

src/django/VLE/tasks/email.py
engelhamer: This is because you did not run preset db again (or at least I changed the default to 'Demo' iso 'eJournal'. For UvA it would need to be changed to 'UvA' iso 'eJournal - UvA' and on the landing page this has been changed accordingly.

src/django/VLE/tasks/email.py
engelhamer: Same as other

src/django/VLE/models.py
```diff
@@ -202,6 +202,18 @@ def auto_delete_file_on_delete(sender, instance, **kwargs):
...
 class UserQuerySet(models.QuerySet):
+    def bulk_create(self, users, *args, **kwargs):
+        with transaction.atomic():
+            users = super().bulk_create(users, *args, **kwargs)
+
+            # Bulk create preferences.
+            preferences = []
+            for user in users:
+                preferences.append(Preferences(user=user))
+            Preferences.objects.bulk_create(preferences)
```
mjvkeulen, line 213: This would be quite suited (and faster, though not really relevant) as a list comprehension.

` Preferences.objects.bulk_create([Preferences(user=user) for user in users])`

mjvkeulen (approved):

PR #1030, opened 2020-10-03
Improved speed and reliability of teacher entries

Larspolo:
## Description
Fixed issue where grade would not also be given

- Fixed Z#1287


src/django/VLE/migrations/0065_teacherentry_updaing.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.10 on 2020-10-08 21:53
```
mjvkeulen, line 1: File name holds a typo

src/django/VLE/views/teacher_entry.py
```diff
@@ -81,6 +81,8 @@ def partial_update(self, request, pk, *args, **kwargs):
         journals, = utils.required_params(request.data, 'journals')
 
         teacher_entry = TeacherEntry.objects.get(pk=pk)
+        if teacher_entry.updating:
+            return response.bad_request('Teacher entry is updating grades in the background, please wait')
```
mjvkeulen, line 85: Forgot the `.`. I am curious how long this takes in practice, any idea?

Larspolo: minute for 200 entries

src/django/VLE/views/teacher_entry.py
```diff
@@ -198,7 +198,7 @@ def _check_teacher_entry_content(self, journals, assignment, is_new=False, teach
 
         Return a list of valid journals if nothing """
         valid_journals = []
-        valid_journal_ids = list(Journal.objects.filter(assignment=assignment).values_list('pk', flat=True))
+        valid_journal_ids = set(Journal.objects.filter(assignment=assignment).values_list('pk', flat=True))
```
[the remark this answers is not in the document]
Larspolo: in set check is faster than in list check I think

src/django/VLE/views/teacher_entry.py
```diff
@@ -133,7 +136,7 @@ def destroy(self, request, *args, **kwargs):
 
     def _create_new_entries(self, teacher_entry, journals_data, author):
         """Copies a teacher entry to journals."""
-        journals = Journal.objects.filter(pk__in=map(lambda j: j['journal_id'], journals_data))
+        journals = Journal.objects.filter(pk__in=map(lambda j: j['journal_id'], journals_data)).order_by('pk')
```
mjvkeulen, line 139: Is ascending pk not the default order?

Larspolo: Dont think so

mjvkeulen: [`If a query doesn’t have an ordering specified, results are returned from the database in an unspecified order.` ](https://docs.djangoproject.com/en/2.2/ref/models/options/#ordering) So fair enough, for our setup it is by asc insert order, but that is no guarantee.

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,21 @@ def publish_all_journal_grades(journal, publisher):
     Comment.objects.filter(entry__node__journal=journal).exclude(entry__grade=None).update(published=True)
 
 
+@shared_task
+def save_teacher_grades(teacher_entry_id, grade_ids):
```
engelhamer, line 28: I would prefer to name this `save_teacher_entry_grades` because practically all grades are given by teachers (which makes the current name confusing).

engelhamer (approved):

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,21 @@ def publish_all_journal_grades(journal, publisher):
     Comment.objects.filter(entry__node__journal=journal).exclude(entry__grade=None).update(published=True)
 
 
+@shared_task
+def save_teacher_entry_grades(teacher_entry_id, grade_ids):
+    """Send new teacher entry grades back to LMS"""
+    TeacherEntry.objects.filter(pk=teacher_entry_id).update(updating=True)
+    results = []
+    try:
+        for grade in Grade.objects.filter(pk__in=grade_ids):
```
mjvkeulen, line 33: This could benefit from `select_related`

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,21 @@ def publish_all_journal_grades(journal, publisher):
...
+            grade.entry.grade = grade
+            grade.entry.save()
+            results.append(task_journal_status_to_LMS(grade.entry.node.journal.pk))
```
mjvkeulen, line 36: Possibly working in an `only` specifically for this call

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,21 @@ def publish_all_journal_grades(journal, publisher):
     Comment.objects.filter(entry__node__journal=journal).exclude(entry__grade=None).update(published=True)
 
 
+@shared_task
+def save_teacher_entry_grades(teacher_entry_id, grade_ids):
+    """Send new teacher entry grades back to LMS"""
+    TeacherEntry.objects.filter(pk=teacher_entry_id).update(updating=True)
```
mjvkeulen, line 30: Using a manual lock (well soft lock) for this specific block is a bit of anti-pattern since this comes shipped with Django itself. What we want to achieve here is locking a specific row of the Entry table. This could be achieved using [`select_for_update`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update). This is then enforced at the DB layer and is guaranteed to be thread-safe.

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,21 @@ def publish_all_journal_grades(journal, publisher):
     Comment.objects.filter(entry__node__journal=journal).exclude(entry__grade=None).update(published=True)
 
 
+@shared_task
+def save_teacher_entry_grades(teacher_entry_id, grade_ids):
```
mjvkeulen, line 28: I believe this leaves us with a race condition, which since this can take a couple of minutes for large assignments, is no longer very unlikely. 

As a teacher or other staff member, I can update a student's Entry grade (including a teacher entry). If this task is running in the background, it will override the set grade since we work with a predefined grade list. 

This is not only tied to moving this logic to a background task since this can occur between threads (let us assume different staff members with reasonable use and timeout config) anyway. But by moving this to a background worker a staff member is much more likely to move on to perform another task (such as grade an Entry scheduled for update), since the posting of the teacher entry itself will now be nearly instantaneous.

I suggest we evaluate this issue more thoroughly, I can imagine our Entry and Grading logic requiring some more checks for `updating` (or locks) to prevent this issue.

src/django/VLE/utils/grading.py
Larspolo: What do you mean?

mjvkeulen: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#only

src/django/VLE/utils/grading.py
Larspolo: Rewrote how grades are set in entries, its now with a computed field that auto takes the last grade by creation_date

src/django/VLE/utils/grading.py
Larspolo: Still used updating for serialization, but also added this

src/django/VLE/utils/import_utils.py
```diff
@@ -8,7 +8,7 @@
 from VLE.utils.error_handling import VLEProgrammingError
 
 
-def _copy_grade_based_on_jir_action(entry, author, action=JournalImportRequest.APPROVED_WITH_GRADES_ZEROED):
+def _copy_grade_based_on_jir_action(entry, grade, author, action=JournalImportRequest.APPROVED_WITH_GRADES_ZEROED):
```
mjvkeulen, line 11: Why the additional parameter? The result still only depends on the current grade of the provided entry.

src/django/VLE/models.py
```diff
@@ -1864,11 +1870,11 @@ class Grade(CreateUpdateModel):
 
     Used to keep a history of grades.
     """
+
     entry = models.ForeignKey(
         'Entry',
-        editable=False,
-        related_name='+',
-        on_delete=models.CASCADE
+        related_name='grade_set',
```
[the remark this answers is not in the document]
Larspolo: Didnt work without this explicitly naming, couldnt find any reason why that was the case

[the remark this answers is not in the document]
Larspolo: >was this needed to force an update over omitting the reverse relation with `+`
Yes

src/django/VLE/utils/import_utils.py
Larspolo: Previously the new grade would be set to the entry from the grade, this will now be set to the first parameter

src/django/VLE/utils/import_utils.py
```diff
@@ -211,23 +213,26 @@ def import_template(template, assignment, archived=None):
...
+        last_edited_by=old_entry.last_edited_by,
         teacher_entry=teacher_entry,
-        vle_coupling=vle_coupling if vle_coupling else entry.vle_coupling,
+        vle_coupling=vle_coupling if vle_coupling else old_entry.vle_coupling,
         jir=jir
     )
+    if grade:
+        grade.entry = entry
+        entry.save()
+        old_entry.save()
```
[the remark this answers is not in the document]
Larspolo: Old code, can be removed indeed

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +25,26 @@ def publish_all_journal_grades(journal, publisher):
...
+    """Send new teacher entry grades back to LMS"""
+    teacher_entry = TeacherEntry.objects.get(pk=teacher_entry_id)
+    teacher_entry.updating = True
+    teacher_entry.save()
+    results = []
+    try:
+        entries = Entry.objects.select_for_update().filter(teacher_entry=teacher_entry)
```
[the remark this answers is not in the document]
Larspolo: > Since we only care about entries where the grade is NOT correctly set, we can exclude entries which are correctly set

This is actually not true, I fixed this in the lastest commit. Non graded entries should still be saved as this will also update the journals marking needed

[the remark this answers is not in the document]
Larspolo: >With this structure we could also just bulk update the correct grades instead

Well, this would then still not update all the connected stats of the journals. Thats why all of this is there in the first place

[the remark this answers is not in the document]
Larspolo: > I feel we need to dive a bit deeper before we can merge this. I do think this is the right track, we just need some more knowledge first. We should possibly extend our test to allow testing for race conditions.

Dissagree, I can remove the `select_for_update` for now, because we dont know what will happen. but a timeout and non synchronized grades IS WAY WORSE then whatever happens after this pull request finally goes through

mjvkeulen (commented):
src/django/test/migrations/test_0064_change_profile_picture2 copy.py rename went wrong I think?

src/django/VLE/utils/grading.py
```diff
@@ -24,6 +24,29 @@ def publish_all_journal_grades(journal, publisher):
...
+                # Change the grade and send it back to the LMS
+                entry.save()
+                results.append(task_journal_status_to_LMS(entry.node.journal.pk))
+            else:
+                # This updates the marking needed state
+                entry.save()
+    finally:
+        teacher_entry.updating = False
+        teacher_entry.save()
+    return results
```
mjvkeulen, line 47: We can still speed this up without too much effort, one transaction block for all bulk creation where we correct the computed properties en masse as well, we can still lock the entry rows for good measure. I think the relevant fields are a Journal's `grade`, `needs_marking`, and `unpublished`. This should provide an excellent user experience (all relevant fields are directly updated), we wil leave the LMS passback as a background task.

I will push a proposal tomorrow morning.

Larspolo: orrr, we could have deployed this today, made further improvements tomorrow, and deployed an even better version in two days

src/django/VLE/views/teacher_entry.py
```diff
@@ -87,24 +97,24 @@ def partial_update(self, request, pk, *args, **kwargs):
...
         entries = Entry.objects.filter(teacher_entry=teacher_entry)
         deleted_entries = entries.exclude(node__journal__pk__in=map(lambda j: j['journal_id'], journals))
         deleted_entry_journal_ids = list(deleted_entries.values_list('node__journal__pk', flat=True))
         existing_journal_ids = list(entries.values_list('node__journal__pk', flat=True))
         deleted_entries.delete()
-        for journal_id in deleted_entry_journal_ids:
-            # Entry and node must be deleted.
-            grading.task_journal_status_to_LMS.delay(journal_id)
 
+        grading.task_bulk_send_journal_status_to_LMS.delay(deleted_entry_journal_ids)
```
Larspolo, line 107: This task is now run 3 times if you update a TE with deleted entries, new entries and updated entries. Do we want to move this to only 1 call?

src/vue/src/components/assignment/PostTeacherEntry.vue
```diff
@@ -258,14 +258,23 @@ export default {
...
                     title: this.title,
                     show_title_in_timeline: this.showTitleInTimeline,
                     assignment_id: this.$route.params.aID,
                     template_id: this.selectedTemplate.id,
                     content: this.teacherEntryContent,
                     journals: this.selectedJournals,
-                }, { customSuccessToast: 'Teacher entry successfully posted.' })
+                }, {
+                    customSuccessToast: 'Teacher entry successfully posted. '
+                        + 'Grades are being updated in the background (this may take a while).',
```
Larspolo, line 276: Do we still keep this here? I would remove it, or change it to sth like: `Grades are being send to Canvas in the background (this may take a while)`

src/vue/src/components/assignment/TeacherEntries.vue
```diff
@@ -149,6 +149,9 @@
                 Save
             </b-button>
         </div>
+        <div v-else-if="selectedTeacherEntry">
+            Grades for this teacher entry are being updated in the background. Please come back later.
```
Larspolo, line 153: This should never be the case anymore right? Can be removed?

src/vue/src/components/assignment/TeacherEntries.vue
```diff
@@ -254,7 +257,10 @@ export default {
                 this.requestInFlight = true
                 teacherEntryAPI.update(this.selectedTeacherEntry.id, {
                     journals: this.selectedJournals,
-                }, { customSuccessToast: 'Teacher entry successfully updated.' })
+                }, {
+                    customSuccessToast: 'Teacher entry successfully updated. '
+                        + 'Grades are being updated in the background (this may take a while).',
```
Larspolo, line 262: Ditto

src/django/VLE/views/teacher_entry.py
mjvkeulen: I don't think we will gain much with the current state of `task_bulk_send_journal_status_to_LMS`, provided the entries do not overlap.

src/django/VLE/utils/import_utils.py
```diff
@@ -87,10 +87,12 @@ def import_entry(entry, journal, jir=None, grade_author=None,
     copied_entry = copy_entry(
         entry,
         node=copied_node,
-        grade=_copy_grade_based_on_jir_action(entry, grade_author, grade_action),
         vle_coupling=_select_vle_coupling_based_on_jir_action(grade_action, entry),
         jir=jir
     )
+    grade = _copy_grade_based_on_jir_action(copied_entry, entry.grade, grade_author, grade_action)
+    if grade:
+        copied_entry.save()
```
mjvkeulen, line 95: Tagging this as well, same reason as earlier.

src/django/test/test_teacher_entry.py
```diff
@@ -284,7 +286,7 @@ def test_valid_content_teacher_entry(self):
         with mock.patch('VLE.utils.entry_utils.check_fields') as check_fields_mock:
             # Ensure teacher entries are validated the same as regular entries.
             api.create(self, 'teacher_entries', params=self.valid_create_params, user=self.teacher)
-            check_fields_mock.assert_called_with(self.template, self.valid_create_params['content'])
```
mjvkeulen, line 287: This was required because the in-memory valid_create_params are converted during the API call, resulting in string keys instead of integer keys.

mjvkeulen (approved):

src/django/VLE/models.py
```diff
@@ -1889,15 +1886,18 @@ class Grade(CreateUpdateModel):
...
-
         if self.published:
             for author in self.entry.node.journal.authors.all():
                 Notification.objects.create(
                     type=Notification.NEW_GRADE,
                     user=author.user,
                     grade=self
                 )
+        # Save entry to set this grade as the new entry grade
+        if is_new:
```
engelhamer, line 1899: This will always be the case as no grades shall ever be edited?

src/django/VLE/utils/import_utils.py
```diff
@@ -211,21 +213,20 @@ def import_template(template, assignment, archived=None):
     return template
 
 
-def copy_entry(entry, node=None, grade=None, vle_coupling=None, teacher_entry=None, jir=None):
+def copy_entry(old_entry, node=None, grade=None, vle_coupling=None, teacher_entry=None, jir=None):
```
engelhamer, line 216: This name change is not needed and the name `old_entry` does not make a lot of sense since it is not contrasted with a 'new' one.

engelhamer (approved):

PR #1032, opened 2020-10-07
Fixed some technical issues

Larspolo:
## Description
Fixed some technical issues to reduce the amount of reports

## Summary of changes
- Updated error message to be more user friendly
- Fixed Sentry#825
- Fixed Sentry#778:  Students can now navigate to assignment url and be redirected to their journal
- Fixed Sentry#256
- Fixed Sentry#179
- Fixed Sentry#828
- Fixed Sentry#885
- Fixed issue where two requests would be send when pressing the enter key
- All cID aID and jID are now converted to integer on load
- Updated config file to: URIs should not be split due to different filenames


src/vue/src/views/LtiLaunch.vue
```diff
@@ -172,15 +172,6 @@ export default {
...
             }
-        }).catch((error) => {
-            this.$router.push({
-                name: 'ErrorPage',
-                params: {
-                    code: error.response.status,
-                    reasonPhrase: error.response.statusText,
-                    description: error.response.data.description,
-                },
-            })
```
mjvkeulen, line 183: When was this fallback an issue?

src/vue/src/views/PasswordRecovery.vue
```diff
@@ -27,7 +27,6 @@
                     type="password"
                     required
                     placeholder="Repeat new password"
-                    @keyup.enter="recoverPassword()"
```
mjvkeulen, line 30: Is this not expected?

Larspolo: It still works, preciously it would be triggered twice

src/vue/src/views/LtiLaunch.vue
Larspolo: Before we had the auto routing on every request

src/vue/src/views/PasswordRecovery.vue
engelhamer: Because the `recoverPassword` function is tied to the form itself.

mjvkeulen (approved):

src/vue/src/views/Assignment.vue
```diff
@@ -1,6 +1,6 @@
-<!-- TODO Is this check really required if we redirect, or even better have correct flow anyway? -->
-<template v-if="$hasPermission('can_view_all_journals')">
-    <content-columns>
+<template>
+    <!-- This check helps prevent user seeing the teacher view of the assignment when -->
```
engelhamer, line 2: When what?

src/vue/src/router/index.js
```diff
@@ -76,37 +76,56 @@ const router = new Router({
...
+            cID: Number.parseInt(route.params.cID, 10),
+        }),
     }, {
         path: '/Home/Course/:cID/Assignment/:aID',
         name: 'Assignment',
         component: Assignment,
-        props: true,
+        props: route => ({
+            cID: Number.parseInt(route.params.cID, 10),
+            aID: Number.parseInt(route.params.aID, 10),
```
engelhamer, line 102: Navigating to e.g. `/Course/1/Assignment/aaa` now causes a 500 error, whereas it previously redirected to home.

engelhamer (commented):
Some weird behaviour for alphabetic characters in an assignment URL.

Larspolo:
> Some weird behaviour for alphabetic characters in an assignment URL.

Problem is bigger than you would expect, created an issue @engelhamer

src/vue/src/router/index.js
Larspolo: Created a separate issue for it is we wanna fix it nicely. I think navigating to home is just hiding the issue, and would prefer notifying the user that the assignment does not exists / sth went wrong

engelhamer (approved):
Issue is fine.

PR #1034, opened 2020-10-11
Added the ability to attach files to preset descriptions

Larspolo:
## Description
Added the ability to add files as teacher in the node description
image
image

## Summary of changes
- Fixes Z#1295
- Fixes Z#1251


src/django/VLE/migrations/0065_presetnode_files.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('VLE', '0064_use_ejournal_default_profile_picture'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='presetnode',
+            name='files',
+            field=models.ManyToManyField(related_name='preset_node_files', to='VLE.FileContext'),
```
mjvkeulen, line 16: I am always torn on how to implement the many to one relation of our files. Is this your preferred method when a file is not embedded in RT? For example, some downsides of this method are:
- DB does not reflect the actual relations (M2M instead of many-to-one). This has the effect that some basic rules need to be enforced manually.
- Additional DB table for the (not truly needed) M2M relationships.
- We mix two methods, consistency drops.

I am also leaning towards this approach, but I am just curious about your input?

src/django/VLE/models.py
```diff
@@ -1699,6 +1699,10 @@ class PresetNode(CreateUpdateModel):
         on_delete=models.SET_NULL,
         null=True,
     )
+    files = models.ManyToManyField(
+        'FileContext',
+        related_name='preset_node_files',
```
mjvkeulen, line 1704: `fc.presetnode_set` vs `fc.preset_node_files`, since the result is a set of `PresetNodes` I would prefer the first.

src/django/VLE/utils/file_handling.py
```diff
@@ -109,7 +114,7 @@ def _move_newly_established_file_context_to_permanent_location(fc):
     os.rename(initial_path, str(new_path))
 
 
-def establish_file(author, identifier, assignment=None, journal=None, content=None, comment=None, in_rich_text=False):
+def establish_file(author, identifier, **kwargs):
```
mjvkeulen, line 117: Would you agree moving to kwargs is more error-prone to leaving out desired context?

src/django/VLE/utils/generic_utils.py
```diff
@@ -236,6 +237,18 @@ def update_presets(assignment, presets, new_ids):
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=new_ids[template['id']])
             else:
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=template['id'])
+
+        preset_node.save()
+        # Add new files
+        for file_date in files:
+            print(file_date)
```
mjvkeulen, line 244: debug

src/django/VLE/utils/generic_utils.py
```diff
@@ -236,6 +237,18 @@ def update_presets(assignment, presets, new_ids):
...
+            file = VLE.models.FileContext.objects.get(pk=file_date['id'])
+            if not preset_node.files.filter(pk=file.pk).exists():
+                preset_node.files.add(file)
+                file_handling.establish_file(author=user, identifier=file.access_id, preset_node=preset_node)
+        # Remove old attached files
+        preset_node.files.exclude(pk__in=[file['id'] for file in files]).delete()
```
mjvkeulen, line 250: We will still need to run similar logic for temp file cleanup, possibly move this to cleanup as well? Working towards a centralized place of logic for file cleanup and all that.

src/django/VLE/utils/file_handling.py
Larspolo: More modular, none of these kwargs are actually needed for this function, so why specify them? Now you only need to specify them in one place, I would say this makes it less error-prone actually, as now you cant forget to add it here as well

src/django/VLE/models.py
Larspolo: I did the same as the `comment_files` implementation

src/django/VLE/migrations/0065_presetnode_files.py
Larspolo: Same implementation as comment files, I do see that this might need to be improved, but how would you say to move that to a new issue and attack both of them at the same time?

src/django/VLE/utils/file_handling.py
mjvkeulen: I understand your argument if it was regarding a helper or private function. However, since this function is called directly (not private), by moving these arguments to kwargs it more likely to leave out context which is in fact, desired. I see it as having access to the expected arguments (e.g. in my IDE), when calling the function (sort of interface/header of what parameters are expected).

src/django/VLE/models.py
mjvkeulen: I still feel my point holds, but you raise a valid point regarding the similarity of the `comment` files naming scheme. 

Comments and PresetNodes, both have two types of files, either embedded in the text (description) or attached. Making this difference more explicit would have my vote. How about:

`comment.attached_files` and `fc.attached_comments`
`preset.attached_files` and `fc.attached_presets` 

Or some variation?

Larspolo: You can change the name if you want, just keep it consistent between the two

src/django/VLE/utils/file_handling.py
Larspolo: Updated to context to make it more clear what the kwargs are

mjvkeulen: Renaming `kwargs` to `context` does not make the actual context parameters more transparent?

Larspolo: Personally I find this more clear. A list of all different parameters you are mostly not gonna use  doesnt make it clear on what needs to be set I think, now its like "context, ah okay, I need to set context". But what I think counts more is that this is now modular, and reduces the amount of places that needs to be updated,if you want to add a new context place

engelhamer:
Hm, there appears to be some inconsistency going on here in terms of related names.. why is that?

src/django/VLE/utils/generic_utils.py
```diff
@@ -236,6 +237,17 @@ def update_presets(assignment, presets, new_ids):
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=new_ids[template['id']])
             else:
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=template['id'])
+
+        preset_node.save()
+        # Add new attached_files
+        for file_date in attached_files:
```
mjvkeulen, line 243: typo

src/django/VLE/utils/generic_utils.py
```diff
@@ -236,6 +237,17 @@ def update_presets(assignment, presets, new_ids):
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=new_ids[template['id']])
             else:
                 preset_node.forced_template = VLE.models.Template.objects.get(pk=template['id'])
+
+        preset_node.save()
```
mjvkeulen, line 241: Was this save required for something specific? Also saved on line 251.

mjvkeulen (approved):

mjvkeulen (changes requested):
I forgot about the outstanding kwargs debate.

mjvkeulen (approved):

src/vue/src/components/assets/file_handling/FilesList.vue
```diff
@@ -0,0 +1,65 @@
...
+        <div
+            v-if="files.length > 0"
+            class="file-list multi-form round-border p-2"
+        >
+            <div
+                v-for="(file, i) in files"
+                :key="i"
+            >
+                <file-display :file="file">
+                    <icon
```
engelhamer, line 12: Clicking this icon not only deletes the file, but also downloads it before deleting (which is undesirable).

src/vue/src/components/assets/file_handling/ImageFileDisplay.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <div class="image-field">
         <div
-            class="image-controls mb-2 unselectable"
+            class="controls unselectable"
```
engelhamer, line 4: Controls may not be specific enough. Be aware this may clash with some libs or something in the future.

engelhamer (changes requested):
Will fix what is required myself now.

engelhamer (approved):

PR #1036, opened 2020-10-12
Permission refresh date compare fix

mjvkeulen:


engelhamer (approved):

Larspolo (approved):
Date is never set to anything other than null or date, so I dont think this will do anything
, but oh well
```
>> new Date() - new Date(null) == new Date() - null
true
```

mjvkeulen:
> Date is never set to anything other than null or date, so I dont think this will do anything
> , but oh well
> 
> ```
> >> new Date() - new Date(null) == new Date() - null
> true
> ```

Store does not load date as a string in some scenarios, resulting in math with a date object and string -> `NaN`. This results in the `canRefreshPermissions` returning false, and the user without valid permissions.

Issue #688, opened 2019-07-03
Entry draft


Larspolo:
**Describe the solution you'd like**
Add the option to save an entry as a draft

Optionally (and ideally):
- Entries auto-save a draft, which upon revisiting the same entry are restored by default.

PR #1038, opened 2020-10-16
Added automatic draft and restore of rich text

mjvkeulen:
Rich text content is now stored in local storage by default as a draft and remains there for 1 week. If a draft can be found for a rich text area, it is automatically restored.

Note that it is important the rich text editor receives a very specific id as it serves as the local storage key for draft storage.


engelhamer:
This is nice but maybe include the other fields of e.g. an Entry as well in this PR? Now it might be weird for users that their rich text content does appear after reloading the page, but not other fields.

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -65,7 +65,7 @@
                 <!-- Newly added fields in template editor have id <0. -->
                 <text-editor
                     v-else-if="field.type == 'rt'"
-                    :id="`rich-text-editor-field-${field.id > 0 ? 'id-' + field.id : 'loc-' + field.location}`"
+                    :id="textEditorID(field)"
                     :key="`rich-text-editor-field-${field.id > 0 ? field.id : field.location}`"
```
Larspolo, line 69: Should key and id be the same? When is which used?

Larspolo (changes requested):
1. go to format editor
2. create new template
3. insert text in first field
4. wait 10 seconds
5. save assignment
6. navigate away
7. repeat 1 and 2
8. -> content is loaded in again, think this should be erased as it is a different new entry

Same goes for when creating a new assignment, or when placing a new comment (most common probably, see screenshot), or when creating a new presetnode in the format editor

image

src/vue/src/components/entry/EntryFields.vue
mjvkeulen: key needs to be unique for that specific dom view, id needs to map the specific RT to storage.

Larspolo: So why not keep it the same?

mjvkeulen: Can do but why does that have your preference? (Either way is fine by me)

Larspolo: For consistency mostly

engelhamer (commented):
In my opinion this is not very usable now, since e.g. comments re-appear illogically after a refresh (like Lars pointed out above). See my previous comment for it to be useful.

mjvkeulen:
Moving to another solution #688

PR #1041, opened 2020-10-22
Added an in flight check to journal import requests

mjvkeulen:


src/django/VLE/models.py
```diff
@@ -2196,6 +2196,14 @@ def get_update_response(self):
 
         return responses[self.state]
 
+    def target_url(self):
+        return '{}/Home/Course/{}/Assignment/{}/Journal/{}'.format(
+            settings.BASELINK, self.target.assignment.courses.first().pk, self.target.assignment.pk, self.target.pk)
+
+    def source_url(self):
+        return '{}/Home/Course/{}/Assignment/{}/Journal/{}'.format(
+            settings.BASELINK, self.source.assignment.courses.first().pk, self.source.assignment.pk, self.source.pk)
```
Larspolo, line 2205: the `gen_url` function is perfect for this use (should also be moved to the utils funciton, but that asside). `gen_url(journal=self.source)` will return this as well

Also, I cant seem to find where this is used, why did you add this?

mjvkeulen: Use: just for debugging.

I couldn't remember the gen_url func, will just call that instead

Larspolo (approved):

src/vue/src/components/journal/JournalImportRequestApprovalModal.vue
```diff
@@ -174,6 +176,8 @@ export default {
                     this.$root.$emit('bv::hide::modal', this.modalID)
                 }
                 this.$emit('jir-processed')
+            }).finally(() => {
+                this.jirPatchInFlight = false
```
Larspolo, line 180: Also for @engelhamer using this is way better than putting it both in the `catch` and `then`. Think we have a couple dozen places where this can also be improved, lets keep this trend going for new changes

Larspolo (approved):

engelhamer (approved):

PR #1042, opened 2020-10-28
Improved LTI speed

Larspolo:
## Description
- add_groups_if_not_exists now doesnt always save anymore, and only once
- select_create_journal now saves once
- only own journals are updated, not every journal in the assignment
- added LTI assignment in preset
- expanded tests


src/django/VLE/lti_launch.py
```diff
@@ -101,16 +101,24 @@ def add_groups_if_not_exists(participation, group_ids):
...
-    participation.save()
+    # Get all existing groups passed by lti that are also in the course
+    groups = Group.objects.filter(lti_id__in=group_ids, course=participation.course)
+    # Get all to-be-created groups
+    non_existing_group_ids = set(group_ids) - set(group.lti_id for group in groups)
+    # Update groups to only contain groups that user is not participant of
+    groups = groups.exclude(pk__in=participation.groups.all())
+    # Add groups
+    for group in groups:
+        participation.groups.add(group)
```
mjvkeulen, line 112: Do I correctly remember that each execution .add executes SQL (and model save)?

src/django/test/test_lti_launch.py
```diff
@@ -16,44 +16,87 @@
...
-    'oauth_timestamp': 'null',
-    'oauth_nonce': 'null'
+    # Participation data
+    'roles': '',
+    'lis_outcome_service_url': 'a grade url',
+    'lis_result_sourcedid': 'a sourcedid',
 }
 
 
+def create_request_body(user=None, course=None, assignment=None):
```
mjvkeulen, line 55: Would be a good opportunity to move this to a params factory?

src/django/VLE/models.py
```diff
@@ -984,6 +984,10 @@ class Participation(CreateUpdateModel):
         default=None,
     )
 
+    def set_groups(self, groups):
+        self.groups.set(groups)
+        update_dependent(self)
```
mjvkeulen, line 989: I don't actually know if this helps, but 
update_dependent(query_set_of_desynced_journals, ['group_dependent_fields']) should trigger a much smaller set of SQL queries?

src/django/VLE/lti_launch.py
Larspolo:
```
>>> p = Participation.objects.get(pk=3)
>>> g = Group.objects.get(pk=1)
>>> p.groups.add(g)
>>> p.save()
asdf
```
Not a model save. SQL probably

Checked the docs, and changed it to the following:
```
    # Add groups
    participation.groups.add(*groups)
```

src/django/VLE/models.py
Larspolo: This already seemed to do the trick of not updating the other computed fields as well. In the screenshot below I added print stmts to all computed fields

image

src/django/VLE/lti_launch.py
Larspolo: Even further improved this. Now it bulk creates, bulk adds, and saves once

src/django/test/test_lti_launch.py
Larspolo: Added issue: https://zube.io/ejourn-al/ejournal/c/1325

src/django/VLE/models.py
mjvkeulen: I misunderstood the relation update_dependent expects towards the updated model instances. It simply expects the (bulk/sql) updated instance as argument (as queryset), instead of the desynced models itself.

src/django/VLE/lti_launch.py
```diff
@@ -101,16 +101,26 @@ def add_groups_if_not_exists(participation, group_ids):
...
+        Group(name='Group {:d}'.format(n_groups + i + 1), course=participation.course, lti_id=group_id)
+        for i, group_id in enumerate(non_existing_group_ids)
+    ]
+    if new_groups:
+        new_groups = Group.objects.bulk_create(new_groups)
+
+    # Add missing newly-created and existing groups and update participation for computed fields
+    if new_groups or non_participating_groups:
+        participation.groups.add(*new_groups, *non_participating_groups)
+        participation.save()
```
mjvkeulen, line 123: Do you know if this save triggers a cascade of update check queries? If so manually updating the journal 'groups' field would be suitable in this scenario.

src/django/VLE/models.py
```diff
@@ -1488,7 +1492,7 @@ def usernames(self):
         return ', '.join(self.authors.values_list('user__username', flat=True))
 
     @computed(ArrayField(models.IntegerField(), default=list), depends=[
-        ['authors.user.participation_set', ['groups']],
+        ['authors.user.participation_set.groups', ['name']],
```
mjvkeulen, line 1495: Would you mind explaining what this change accomplishes? I don't quite follow how 'name' specifically ties into the computed property.

Larspolo: Mostly did this for when we want to upgrade to the later computedfields version (`['authors.user.participation_set', ['groups']]` doesnt work because groups is M2M and cant be in a value field I think). I somehow didnt include this in this PR, wanna upgrade?

src/django/VLE/lti_launch.py
Larspolo: Added test, without save it also works

mjvkeulen: Also in my tests .add(group) will also update the respective journals computed field `groups`, does the .save update something else?

Larspolo: Nothing that is needed, so I removed it

mjvkeulen: Ah sorry, had not refreshed the pages, did not see your follow up comment

src/django/VLE/models.py
mjvkeulen: Right ok, TIL

src/django/VLE/management/commands/preset_db.py
```diff
@@ -173,7 +175,29 @@ def gen_assignments(self):
...
+                    user=user,
+                    assignment=self.lti_assignment,
+                    course=self.lti_assignment.courses.first(),
+                )).url,
+                full_name=user.full_name,
+                username=user.username
+            )
+        launch_links += '</ul>'
+        self.lti_assignment.description = self.lti_assignment.description + '<br/>' + launch_links
+        self.lti_assignment.save()
```
mjvkeulen, line 198: Probably worthy of a function.

Also, this yields some funky urls for me, e.g.:
../../../../../LtiLogin?state=0&lti_params=<shortend>&username=student&username_already_exists=True&full_name=Lars+van+Hijfte

Larspolo: But does it work?

mjvkeulen: They work!

mjvkeulen: However, not so straight forward. Only when I edit assignment -> right click link, open in new window (tinymce panel).

Larspolo: Actually, I think its only the tinymce panel that "beautifies" the link display. When checking the DB they are still absolute links:

image

```
>>> a = Assignment.objects.get(name="Lti Assignment")
>>> a.description
'Test Your LTI stuff with this. Launch URLs (works for 15 min after generation).<br/><ul><li><a href="http://localhost:8080/LtiLogin?state=0&lti_params=...&username=student&username_already_exists=True&full_name=Lars+van+Hijfte">Lars van Hijfte (student)</a></li><li>...
```

Larspolo: >Probably worthy of a function.

Is this a "requested changes"? I would like for this to be expanded in the future with a full issue containing:
- create new user flow
- already connected user flow
- create new course flow
- create new assignment flow (at least 3, to test import, copy and new)
- create new teacher flow
- maybe even remove the limitation of the 15 minutes

If you really want this in another function, go ahead, but I saw this more as a quick addition which can be improved on later

engelhamer (approved):

mjvkeulen (approved):

PR #1043, opened 2020-10-30
Improved the steps for adding deadlines to the timeline

Larspolo:
## Description
Also added a "create new template" button for a more chronological order of adding entrydeadline to timeline

Before:
image

After:
image

- Fixes Z#1371


mjvkeulen (commented):
Some stuff I encountered on Chrome:
image

Icons are now the same? Did you not like the difference?
image

The create new template button should be floating right I believe inline with the rest. This would clash a bit, having a double add button stacked vertically in two different cards. Would it not be possible to present create a new template from the drop down menu?
image

Larspolo:
> The create new template button should be floating right I believe inline with the rest. This would clash a bit, having a double add button stacked vertically in two different cards.

Can be moved to the right, fine with that. Can also be moved to the same line:
image


> Would it not be possible to present create a new template from the drop down menu instead?

Is possible, but then it hides the option too much, I dont think its natural to think to add a template from inside a select. Does the above ss suffice?

mjvkeulen:
> Is possible, but then it hides the option too much, I dont think its natural to think to add a template from inside a select. Does the above ss suffice?

I definitely prefer it over the previous implementation. I still favor the select option, the placeholder could be changed to `Please select or create a template`. 

@engelhamer can be the decider on this.

No further outstanding issues on my end.

Larspolo:
> "Please select or create a template"

I tried this as well, it doesn't look as good as envisioned (especially with the auto capitalization), nor is it common for an option inside a select statement to run other functions besides just selecting the option.

src/vue/src/components/assets/RadioButton.vue
```diff
@@ -3,11 +3,13 @@
         <b-button
             v-for="option in options"
             :key="`radio-button-${option.value}`"
+            v-b-tooltip:hover="option.tooltip"
             :class="[{ 'active': value == option.value }, option.class]"
             class="radio-button-option"
             @click="$emit('input', option.value)"
         >
             <icon :name="option.icon"/>
+            {{ option.text }}
```
engelhamer, line 12: These changes are not needed (not currently used anywhere)

src/vue/src/components/format/FormatAddPresetNode.vue
```diff
@@ -1,87 +1,65 @@
...
-            <h2 class="theme-h2 d-inline multi-form">
-                New preset
-            </h2>
-            <h2 class="theme-h2 field-heading required">
-                Preset Type
-            </h2>
-            <b-row>
-                <b-col md="6">
-                    <b-card
-                        :class="{'unselected': currentPreset.type !== '' && currentPreset.type !== 'd'}"
```
engelhamer, line 16: The old method with opacity made it clearer which was selected. I'll propose a change shortly.

engelhamer (commented):
image

Looks a bit unfortunate on mobile. The create button takes up too much space in my opinion. Will get back to this shortly.

engelhamer:
I think this makes it more clear which option is selected:

image

engelhamer:
Hm, I think the two add buttons (both with + and 'Add') so near one another will introduce confusion during actual use. Especially if a user does not know exactly what the terms mean. In combination with the space issues on mobile and the fact that you cannot change the template in the same location anyways (you still need to navigate to the side menu for that) I reckon a link text would maybe be more appropriate? Kind of like the option to input usernames instead of manual selection for teacher entries.

mjvkeulen:
I can no longer edit the description of a preset when switching between entry and progress

engelhamer (approved):

mjvkeulen (approved):

PR #1045, opened 2020-11-05
Added restrictions on function length and complexity

mjvkeulen:
Over time these can be decreased.


src/vue/src/views/FormatEdit.vue
```diff
@@ -510,6 +511,7 @@ export default {
                 })
                 .catch(() => { this.saveRequestInFlight = false })
         },
+        /* eslint-enable complexity */
```
Larspolo, line 514: Would suggest to not do this, as this later hides these functions. I would opt for one of the following:
- reduce the complexity of these two functions to set a lower minimum
- increase the minimum to the complexity of these functions

After this its fine by me to just add this PR and later reduce the numbers to more sensible stuff

mjvkeulen: > After this its fine by me to just add this PR and later reduce the numbers to more sensible stuff

I am fine with this as well, would you be ok with leaving that one function as an exception? Its cyclomatic complexity is 60, which is quite out of line with the rest of our code so I figured better to just make the exception once so we could keep the number 3x lower without rewriting anything.

Larspolo: Nah, Ill change the function myself then to already get this reduction going. The goal of this PR is to reduce complexity right? Not hide high complexity functions

Larspolo (approved):

setup.cfg
```diff
@@ -1,6 +1,9 @@
 [flake8]
 max-line-length = 120
 exclude = */migrations/*,*/settings/*,*/tasks/__init__.py,*/test/factory/__init__.py
+max-complexity = 22
+max-function-length = 210
```
engelhamer, line 5: Why these numbers? To accept the status quo and prevent worse stuff from happening? That would not really seem like a sound method right?

src/vue/src/views/FormatEdit.vue
```diff
@@ -281,6 +281,94 @@ export default {
 
             currentTemplate: -1,
             newTemplateId: -1,
+            lastTarget: 0,
+            templatesIds: [],
+
+            presetErrors: [
```
engelhamer, line 287: The construction is alright, but it is not really sensible to place these errors here as they are not used anywhere but in the dedicated funtion below. I do not think that moving this to the component data achieves any of the goals (i.e. readability, maintainability etc.) for which this PR is opened.

src/vue/src/views/FormatEdit.vue
```diff
@@ -320,33 +408,19 @@ export default {
                     }
                 })
         },
+        checkDateOrder (first, second) {
+            // Returns false if order is incorrect or one of the dates is null
```
engelhamer, line 412: Returns true if any of the dates is null?

engelhamer: Comment may not be needed at all.

src/vue/src/views/FormatEdit.vue
```diff
@@ -358,130 +432,24 @@ export default {
...
-                        + 'Please check the timeline and try again.')
-                }
-                if (!presetLockAfterLock && this.assignmentDetails.lock_date
-                    && Date.parse(preset.lock_date) > Date.parse(this.assignmentDetails.lock_date)) {
-                    presetLockAfterLock = true
-                    this.$toasted.error('One or more presets have a lock date after the assignment lock date. '
-                        + 'Please check the timeline and try again.')
+            let hasError = false
+            this.presetErrors.forEach((error) => {
+                if (error.happened) {
```
engelhamer, line 447: I'd prefer `occurred` instead (more colloquial to errors).

src/vue/src/views/FormatEdit.vue
engelhamer: Same holds for the `templatesIds` and `lastTarget`.

engelhamer (changes requested):

setup.cfg
mjvkeulen: This is our current upper limit. What do you propose?

engelhamer: Well I do not have a strong preference for applying a specific limit, but just picking the max lines we currently have does not have added value right?

mjvkeulen: It sets an upper limit, something we can reduce over time. 

I could also reduce this limit by creating a couple of exceptions for current high line/complexity cases. So we could use the lower limit as a more practical guideline without introducing a lot of refactoring work. I originally did this for one function (60CC) so we could lower the CC to 22. @Larspolo provided feedback that was not his preferred approach.

engelhamer: Okay! Reducing over time seems valid. Shall we create an issue for that?

src/vue/src/views/FormatEdit.vue
Larspolo: Moving it inside the function means it gets re-initialized every time the function is run. I would prefer this option over inside the function. But I also see your point, do you have another idea on where to put it outside of the function?

engelhamer: Given that this is just some JS client stuff, I don't think that will matter in terms of performance?

Larspolo: Its more that I dont see moving it insight of the function being a better option then leaving it in de component data. Both seem like bad practices, and then I will go for the one where the linter works & is better on performance

engelhamer: The linter is in place to make things more maintainable or readable. This option achieves neither: rather the contrary since you have to scroll elsewhere to read these errors, they are not used anywhere, and they are stored globally..

Larspolo: Only when you are actually in the format editor its loaded. You don't see improvements on maitainability and readability with this approach compared to the old one? For me it was a bit of a mess, whereas now its all in a single list viewable with functions that indicate what is actually being checked

engelhamer: That single list would otherwise be in a much more sensible location, namely in the only function where it is used. The only thing that this does is introduce more scrolling?

engelhamer: Just copy the list to the top of the function itself then.

Larspolo: done

engelhamer (approved):

PR #1046, opened 2020-11-09
Improved speed of publishing assignments

Larspolo:
## Description
Previously it took way to long to publish big assignments, now its about 1.5 seconds for 15 nodes and 300 students

- Fixes Z#1330

Note: this took ~3:30 of development


src/django/VLE/models.py
```diff
@@ -1224,28 +1228,115 @@ def save(self, *args, **kwargs):
...
+            )
+        else:
+            needs_lti_link = '{}'
+
+        journals.update(
+            name=Subquery(
+                AssignmentParticipation.objects.filter(journal__id=OuterRef('pk')).values('user__full_name')[:1]
+            ),
+            image=Subquery(
+                AssignmentParticipation.objects.filter(journal__id=OuterRef('pk')).values('user__profile_picture')[:1]
```
mjvkeulen, line 1289: I believe `__id` can be omitted

src/django/VLE/models.py
```diff
@@ -1224,28 +1228,115 @@ def save(self, *args, **kwargs):
...
+            ),
+            full_names=Subquery(
+                AssignmentParticipation.objects.filter(journal__id=OuterRef('pk')).values('user__full_name')[:1]
+            ),
+            usernames=Subquery(
+                AssignmentParticipation.objects.filter(journal__id=OuterRef('pk')).values('user__username')[:1]
+            ),
+            groups=ArraySubquery(
```
mjvkeulen, line 1297: Mind explaining ArraySubquery? 
I also believe your query is more strict, would this also fetch groups the user is not linked to via the journal? Possibly add this in a test?

src/django/VLE/models.py
```diff
@@ -1224,28 +1228,115 @@ def save(self, *args, **kwargs):
...
+                Group.objects.filter(participation__user__assignmentparticipation__journal__in=OuterRef('pk'))
```
mjvkeulen, line 1298: To replace groups as annotation I used:

```
def annotate_groups(self):
    """Annotates for each journal all journal users their groups as an array of group pks `usernames_annotate`"""
    return self.annotate(groups_annotate=ArrayAgg(
        'authors__user__participation__groups',
        filter=Q(authors__user__participation__groups__isnull=False),
        distinct=True,
    ))
```

src/django/test/test_journal.py
```diff
@@ -739,3 +739,27 @@ def test_lock(self):
             }, user=self.g_teacher)
         assert Journal.objects.get(pk=self.group_journal.pk).locked, \
             'Teacher should sitll be able to lock journal'
+
+    def test_update_journal_computed_fields(self):
+        def correct_content(journal):
+            assert journal.name == journal.authors.first().user.full_name
+            assert journal.image == journal.authors.first().user.profile_picture
```
mjvkeulen, line 746: I believe this would not hold if the user's profile picture was not the default?

mjvkeulen (commented):
Good stuff, thanks for the functions ;) I found it excellent to read/follow.

src/django/test/test_journal.py
engelhamer: Indeed if one of the users has a custom profile picture, that will be used.

src/django/VLE/models.py
```diff
@@ -1274,28 +1274,84 @@ def save(self, *args, **kwargs):
...
+        """
+        aps = [
+            AssignmentParticipation(assignment=self, user=user, journal=journal)
+            for user, journal in zip(users, journals or [None] * users.count())
+        ]
+        aps = AssignmentParticipation.objects.bulk_create(aps)
+
+        if new_assignment_notification:
+            for ap in self.assignmentparticipation_set.exclude(user=self.author):
+                ap.create_new_assignment_notification()
```
mjvkeulen, line 1348: Still makes sense to move this to a single task

Larspolo: Done

src/django/VLE/models.py
```diff
@@ -1224,28 +1228,115 @@ def save(self, *args, **kwargs):
...
+            ),
+            groups=ArraySubquery(
+                Group.objects.filter(participation__user__assignmentparticipation__journal__in=OuterRef('pk'))
+                .values('pk').distinct()
+            ),
+            needs_lti_link=needs_lti_link,
+        )
+        return journals
+
+    def create_nodes(self, journals):
```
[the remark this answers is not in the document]
mjvkeulen: I would be fine with introducing the logic here (since we are doing that now anyway, might as well do it clean), and moving the refactor to an issue?

engelhamer (approved):

src/django/VLE/models.py
```diff
@@ -1866,6 +1923,17 @@ def can_add(self, user):
...
             and self.assignment.format.template_set.filter(archived=False, preset_only=False).exists()
 
+    def generate_missing_nodes(self, create=True):
+        nodes = [Node(
+            type=preset_node.type,
+            entry=None,
+            preset=preset_node,
+            journal=self,
+        ) for preset_node in self.assignment.format.presetnode_set.all()]
+        if create:
```
mjvkeulen, line 1933: Would you be willing to concede to one hspace here as well? Just to separate the if block from the list comprehension?

mjvkeulen (approved):

src/django/VLE/models.py
```diff
@@ -1606,6 +1600,19 @@ class Meta:
...
+            journals = super().bulk_create(journals, *args, **kwargs)
+
+            # Bulk create nodes
+            nodes = []
+            for journal in journals:
+                nodes += journal.generate_missing_nodes(create=False)
+            # Notifications should not be send when the journal is new. A "new assignment" notification is good enough
+            Node.objects.bulk_create(nodes, new_node_notifications=False)
+
+            return journals
```
engelhamer, line 1614: May be a bit confusing that the default `bulk_create` returns ids iso objects, correct?

PR #1047, opened 2020-11-09
Improved load times for most requests via serialization

mjvkeulen:
Node view (timeline serialization) speedup resulted in a more thorough approach to serialization speeds.

Fixes Z#1089


src/django/VLE/serializers.py
```diff
@@ -212,156 +410,93 @@ class AssignmentSerializer(serializers.ModelSerializer):
...
     def get_journals(self, assignment):
-        """Retrieves the journals of an assignment."""
-        if 'journals' in self.context and 'course' in self.context \
-           and self.context['journals'] and self.context['course']:
-            # Group assignment retrieves all journals
+        """
+        The assignment is serialized from the perspective of a specific course. This impacts which users and journals
+        should be used in queries.
+        """
+        journals = self.context.get('journals', False)
```
Larspolo, line 469: This is falsy by default (`None`), imo this can be left out here. as well as other places where `None` will also suffice like:
```
    def get_usernames(self, journal):
        if self.context.get('can_view_all_journals', False):
            return journal.usernames
        return None
```

src/django/VLE/serializers.py
```diff
@@ -554,18 +677,43 @@ class Meta:
...
+
+        permissions = [
+            'can_view_all_journals',
+            'can_manage_journal_import_requests',
+        ]
+
+        if not assignment:
+            if isinstance(self.instance, VLE.models.Journal):
+                assignment = self.instance.assignment
+            # NOTE: This is no longer valid if journals from different assignments are serialized at the same time
```
Larspolo, line 699: Think I would not like this, can we just not set assignment in those cases? Would like to see the assignment set in the context on all places where possible. Places where no single assignment is know, should not then get a random assignment as permission check. We could change this to loop over all assignments of the journals and if for 1 assignment the user doesnt have the permissions -> False for all

src/django/VLE/serializers.py
```diff
@@ -574,44 +722,34 @@ def get_groups(self, journal):
...
 
-    class Meta:
-        model = VLE.models.PresetNode
-        fields = ('id', 'description', 'type', 'unlock_date', 'due_date', 'lock_date', 'target', 'template')
-        read_only_fields = ('id', 'type')
-
     def get_unlock_date(self, node):
         if node.type == VLE.models.Node.ENTRYDEADLINE and node.unlock_date is not None:
-            return node.unlock_date.strftime('%Y-%m-%dT%H:%M')
+            return node.unlock_date.strftime('%Y-%m-%dT%H:%M')  # QUESTION: Why does this format deviate?
```
Larspolo, line 752: QUESTION: why do the others include seconds as well? We only support minutes right?

src/django/VLE/serializers.py
```diff
@@ -629,36 +767,94 @@ def get_target(self, node):
...
+    # NOTE: Comments (comment_set) are only serialized via GDPR, which prefetches the comment_set itself.
+    prefetch_related = [
+        'content_set',
+        'content_set__field',
+        # QUESTION: Is this desired for the common case?
+        # Faster for multiple FILE fields,
+        # slower if not present, or for many embedded images (also prefetched (only the model instance) but not used)
+        'content_set__filecontext_set',
+        # QUESTION: Again common case, 1 query base or N queries for an imported journal
+        # 'jir__source__assignment__courses',
```
Larspolo, line 845: Think both are done as I would do too, JIR is too uncommon

src/django/VLE/serializers.py
```diff
@@ -682,39 +882,57 @@ def _get_content(entry):
...
         if entry.jir:
+            jir_source_course = entry.jir.source.assignment.get_active_course(self.context['user'])
             return {
                 'source': {
-                    'assignment': AssignmentSerializer(entry.jir.source.assignment, context=self.context).data,
+                    'assignment': {
+                        'name': entry.jir.source.assignment.name,
+                        'course': {
+                            # QUESTION: Preference? First no query cost, apart from first prefetch
+                            # second one can_view_course check per entry
```
Larspolo, line 927: Maybe we should just leave it at the assignment name (with PK?), its not even the real course anyway right? Could even generate a 500 error if there is no active course to the user

Larspolo (commented):
General question just to have it all clear:

When do you use the following:
- Just the serializer with the query inside
- Use setup_eager_loading first
- Do sth like `{field: getattr(user, field) for field in UserSerializer.Meta.user_fields}`

Think I know, but then this question (or here it becomes clear that I dont know):
can we make it that the serializers themself reflect this logic -> rewrite the userserializer to just get the methods in a dict for example, then we will use it correctly automatically

src/django/VLE/serializers.py
mjvkeulen: A good place to tag this, the line below u see I omit the default.

```
journals = self.context.get('journals', False)
course = self.context.get('course')
```

From the context, the expected type is not always obvious.

`context['journals']` is just used as a boolean (opposed to a list of journals), which I made clear by the default.
`context['course']` is actually a course instance. 

For me working with the code it's more clear with the default, I think I broke something due to this at some point.

I am generally in favor of omitting function defaults, they just clutter and increase reading complexity (most of the time). 

However such defaults are not always obvious, e.g. a serializer is flagged as `read_only=True` by default when a queryset or many=True is passed as kwarg, unless a custom list serializer is implemented which can handle create or update operations. This is only obvious if the reader has some knowledge about the serializer class.

We have an outstanding issue regarding type hinting, and I am sure I was not fully consistent about providing a default everywhere. So I don't really know where to go with this, it depends I guess? 

For the line you tagged specifically the default would have saved me a bug during coding.

src/django/VLE/serializers.py
mjvkeulen: @engelhamer ? I believe most dates follow the format set on the `Field` class: ALLOWED_DATETIME_FORMAT

src/django/VLE/serializers.py
Larspolo: > which I made clear by the default.

It now indeed is a bit weird and unclear, the default helps, but doesnt solve the core problem I think.

Would prefer an option where it is clear that the name is a boolean: `also_get_journals`. Then no default value is needed, and we dont later on overwrite this variable by the actual journals. Would this solve the problem also?

src/django/VLE/serializers.py
mjvkeulen: (`'%Y-%m-%dT%H:%M:%S'`)

src/django/VLE/serializers.py
mjvkeulen: You mean do not prefetch filecontext_set aswell?

src/django/VLE/timeline.py
```diff
@@ -49,49 +56,65 @@ def get_nodes(journal, author=None):
...
+        'jID': journal.pk,
         'unlock_date': node.preset.unlock_date,
         'due_date': node.preset.due_date,
         'lock_date': node.preset.lock_date,
+        # NOTE: 'template' duplicate serialization, Entry also serializes its template.
+        # Is it needed to serialize the template, if an Entry is present?
         'template': TemplateSerializer(node.preset.forced_template).data,
-        'entry': EntrySerializer(node.entry, context={'user': user}).data if node.entry else None,
+        'entry': EntrySerializer(
+            EntrySerializer.setup_eager_loading(Entry.objects.filter(node=node)).first(),
```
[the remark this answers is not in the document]
mjvkeulen: Good catch, [0] should be get() where possible else first(). I refactored this later and missed this one.

src/django/VLE/models.py
```diff
@@ -2068,6 +2201,9 @@ class Field(CreateUpdateModel):
 
     Defines the fields of an Template
     """
+    class Meta:
+        ordering = ['location']
```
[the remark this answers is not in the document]
mjvkeulen: It does mean ORDER_BY is always applied, not free. We could definitely add this to models for which it makes a lot of sense.

src/django/VLE/serializers.py
Larspolo: No I ment I agree with your current configuration, fetching fielcontext, not fetching JIR courses

src/django/VLE/models.py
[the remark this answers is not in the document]
Larspolo: True, lets not always add it to all, maybe we should add a `sort_default()` function to all models to keep a consistent ordering only when asked for

src/django/VLE/models.py
```diff
@@ -1383,6 +1426,61 @@ def conflicting_lti_link(self):
...
+        """
+        Return the earliest date that an entry has been submitted and is not yet graded or has an unpublished grade
+        """
+        return VLE.models.Journal.objects.filter(assignment=self).require_teacher_action().values(
+            'node__entry__last_edited'
+        ).aggregate(
+            Min('node__entry__last_edited')
+        )['node__entry__last_edited__min']
+
+    def get_student_deadline(self, journal):
```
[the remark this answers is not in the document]
mjvkeulen: They operate on a single instance (which makes it a good candidate for a model function), and this allows one to select the function for unit testing a bit more straight forward as well.

src/django/VLE/models.py
```diff
@@ -1310,32 +1338,47 @@ def get_active_lti_course(self):
...
     def get_active_course(self, user):
-        """"Query for retrieving the course which is most relevant to the assignment."""
+        """"
+        Query for retrieving the course which is most relevant to the assignment.
+
+        Compatible with prefetched courses.
+        Will trigger N permission queries for N courses.
+        """
         # If there are no courses connected, return none
+        courses = self.courses.all()
```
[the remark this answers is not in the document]
mjvkeulen: > When do you use the following:
 
> * Just the serializer with the query inside

If there is no query manipulation required (select_related / prefetching). We have many _leaf_ level serializers, serializers that do not nest any other serializers (e.g. `InstanceSerializer`). So for these, it makes sense. Some of these do not even have any method fields / permission checks, and are basically just model to dict functions. For these simple serializers, if speed is of concern, just using query.values(...) to json (by response) is much faster and is also a valid option. Here is an excellent article on the topic: https://hakibenita.com/django-rest-framework-slow

> * Use setup_eager_loading first

Most likely calling a serializer that contains multiple nested serializers, e.g. `EntrySerializer`. The serializer represents more than a single DB table (entry also serializes its content (fields / files), author, template etc), all of these nested representations will cost multiple additional queries (for each of the primary table rows), so best we get in front of these by calling eager loading.

> * Do sth like `{field: getattr(user, field) for field in UserSerializer.Meta.user_fields}`

So this would leaf nodes which are fetched in sufficient bulk that it makes sense, e.g. list all users. Or it could be that we want the table to be serialized with fewer/different fields than existing serializers (creating a serializer class is sometimes silly compared to just .values('pk', 'name').

> Think I know, but then this question (or here it becomes clear that I dont know):
> can we make it that the serializers themself reflect this logic -> rewrite the userserializer to just get the methods in a dict for example, then we will use it correctly automatically
I am not too sure what you mean here, we could **add** a serializer which is just a function converting a user query set to the desired fields, in this fashion, we would lock down what fields are serialized for the faster calls. But this is also less flexible (we want an additional field in a different view method, new function? or just add it to .values() there).

mjvkeulen:
> When do you use the following:
 
> * Just the serializer with the query inside

If there is no query manipulation required (select_related / prefetching). We have many _leaf_ level serializers, serializers that do not nest any other serializers (e.g. `InstanceSerializer`). So for these, it makes sense. Some of these do not even have any method fields / permission checks, and are basically just model to dict functions. For these simple serializers, if speed is of concern, just using query.values(...) to json (by response) is much faster and is also a valid option. Here is an excellent article on the topic: https://hakibenita.com/django-rest-framework-slow

> * Use setup_eager_loading first

Most likely calling a serializer that contains multiple nested serializers, e.g. `EntrySerializer`. The serializer represents more than a single DB table (entry also serializes its content (fields / files), author, template etc), all of these nested representations will cost multiple additional queries (for each of the primary table rows), so best we get in front of these by calling eager loading.

> * Do sth like `{field: getattr(user, field) for field in UserSerializer.Meta.user_fields}`

So this would leaf nodes which are fetched in sufficient bulk that it makes sense, e.g. list all users. Or it could be that we want the table to be serialized with fewer/different fields than existing serializers (creating a serializer class is sometimes silly compared to just .values('pk', 'name').

> Think I know, but then this question (or here it becomes clear that I dont know):
> can we make it that the serializers themself reflect this logic -> rewrite the userserializer to just get the methods in a dict for example, then we will use it correctly automatically

I am not too sure what you mean here, we could **add** a serializer which is just a function converting a user query set to the desired fields, in this fashion, we would lock down what fields are serialized for the faster calls. But this is also less flexible (we want an additional field in a different view method, new function? or just add it to .values() there).

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: I can see that being a useful addition to the models respective querysets.

src/django/VLE/timeline.py
```diff
@@ -17,13 +17,21 @@ def get_nodes(journal, author=None):
     add-node if the user can add to the journal, the subsequent
     progress node is in the future and maximally one.
     """
-    can_add = author and author not in journal.authors.all() and \
-        author.has_permission('can_have_journal', journal.assignment) and not len(journal.needs_lti_link) > 0
+    can_add = author \
+        and journal.authors.filter(user=author).exists() \
```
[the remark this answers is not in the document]
mjvkeulen: I can't see why it should be there. If you see a scenario I will gladly write a test for it?

src/django/VLE/timeline.py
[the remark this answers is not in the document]
mjvkeulen: So this was this intentional, a deadline might not yet hold an Entry.

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: Doesn't matter if prefetched, .all() does not modify the query so the prefetch is used.

src/django/VLE/serializers.py
```diff
@@ -6,15 +6,84 @@
...
+        instance (model instance)
+        name (str), key expected to be present in the instance cache
+        default, value to be returned if the key is not preset in the cache, None by default
+
+    Returns a tuple
+        0: values of the cache keyed by name, None by default (or the given default)
+        1: (bool), whether the name was found in the instance cache and it was populated sucesfully
+    """
+    if hasattr(instance, '_prefetched_objects_cache'):
+        if name in instance._prefetched_objects_cache:
```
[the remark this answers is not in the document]
mjvkeulen: You will have to point out the other locations, depending on the number of nests / logic I sometimes prefer it for readability or ease of expansion (like dangling commas for git diffs). I don't see a reason here so I will update it.

src/django/VLE/serializers.py
mjvkeulen: I updated the journals flag to `serialize_journals`

src/django/VLE/serializers.py
mjvkeulen: Moving the question to a NOTE, and removing the other question.

src/django/VLE/serializers.py
mjvkeulen: Since we decided against the prefetch of the `jir__source__assignment__courses`, the question is moot, and we should just stick with the get active course.

src/django/VLE/serializers.py
mjvkeulen: Atm, we do not serialize journals across assignments. If that were to change, I would be in favor of tackling it then. 

I don't think the else False for all is a final solution. So let's leave it for when we know the scenario and what would be required?

src/django/VLE/models.py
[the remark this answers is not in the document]
Larspolo: Created issue

src/django/VLE/timeline.py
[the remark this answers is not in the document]
Larspolo: Owww wow. I found what the problem was. It was with the param "author".

`author not in journal.authors.all()` assumes author is an AP -> will almost always return false

`journal.authors.filter(user=author).exists()` correctly assumes author is a user -> will only return true when author is in journal

This ment that untill now users with "can_have_journal" and that were allowed to view other students journal, would have seen an add-button in the timeline of other students journals

You could still write a test for it so it doesnt occure in the future, but this bug fixed now after 8 months

src/django/VLE/timeline.py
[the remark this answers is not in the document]
Larspolo: Then both `.first()`? Would probably doesnt matter much

src/django/VLE/serializers.py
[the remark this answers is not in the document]
Larspolo: The one I pointed out earlier, not in this PR

src/django/VLE/serializers.py
Larspolo: Added sentry message for now if this happens. Quite dangerous if it will happen without us noticing

```
                if self.instance.count() > 1:
                    capture_message(
                        'Journal serializer is used across assignments. This is may lead to permission issues',
                        level='warning'
                    )
```

src/django/VLE/timeline.py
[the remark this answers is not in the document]
mjvkeulen: What do you mean both? First is not prefered since it is both order by and limit 1

src/django/VLE/serializers.py
engelhamer: This is to mitigate issues with date time pickers: Flatpickr uses this format and if you give it something else it will (annoyingly) change the original value a bit.

src/django/VLE/models.py
```diff
@@ -1383,6 +1431,61 @@ def conflicting_lti_link(self):
...
+        name = None
+
+        if journal is not None:
+            for node in VLE.utils.generic_utils.get_sorted_nodes(journal).prefetch_related('entry__grade'):
+                if node.entry:
+                    grade = node.entry.grade
+                else:
+                    grade = None
+
+                # Sum published grades to check if PROGRESS node is fullfiled
```
engelhamer, line 1463: Bonus points should also be included in this sum, so start with that as the initial value for t_grade. (grade_sum as a more semantic name?)

engelhamer: Also maybe add a test for this.

src/django/VLE/models.py
```diff
@@ -1456,11 +1559,34 @@ class Meta:
...
-    def get_queryset(self):
+class JournalQuerySet(models.QuerySet):
+    def order_by_authors_first(self):
+        """Order by journals with authors first, no authors last."""
+        return self.order_by(F('authors__journal').asc(nulls_last=True)).distinct()
+
+    def for_course(self, course):
+        """Filter the journals from the perspective of a single course."""
+        return self.filter(
+            # QUESTION: This effectively makes it impossible to join a journal with participants from a different course
```
engelhamer, line 1570: This is undesirable and unpredictable for teachers right? I would prefer this to be possible.

src/django/VLE/models.py
```diff
@@ -1647,6 +1778,12 @@ def unpublished_nodes(self):
             Q(entry__grade__isnull=True) | Q(entry__grade__published=False),
             entry__isnull=False).order_by('entry__last_edited')
 
+    @property
+    def author(self):
+        if self.author_limit > 1:
+            raise VLEProgrammingError('Unsafe used of journal author property')
```
engelhamer, line 1784: `use`

src/django/VLE/serializers.py
```diff
@@ -24,71 +92,164 @@ class Meta:
...
-        except VLE.models.Participation.DoesNotExist:
-            return None
-        if role:
-            return role.name
-        else:
-            return None
+        user_fields = (
+            'id',
+            'is_test_student',
+            # QUESTION: full_name is not permission guarded compared to username and profile picture.
```
engelhamer, line 143: I think that anyone who can retrieve this serializer may also see the corresponding profile picture (for which only a `can_view` check is used below, which could indeed be removed in my opinion). The username on the other hand should not be visible for all. Take for example a group journal where students should not be able to see usernames (student IDs) of their fellow students. This would leak info that is not freely available elsewhere (e,g, on the LMS).

engelhamer (commented):
Review not finished yet.

src/django/VLE/serializers.py
mjvkeulen: So why do we have two formats @engelhamer

src/django/VLE/models.py
mjvkeulen: Code is quite old, I have not touched the logic apart from speedup here. Renamed t_grade variable as requested.

src/django/VLE/serializers.py
mjvkeulen: Removed check from profile picture, username is still guarded. Removed question.

src/django/VLE/models.py
engelhamer: But besides the name change bonus points should be included in the check too.

mjvkeulen: Ye it makes sense to change this with a test, will do when fresh

src/django/VLE/models.py
mjvkeulen: This is why that question exists, but why was this logic in place? When DO we want to serialize journals from the perspective of a signal course within an assignment?

engelhamer: Well that was not in the question: it wasn't even a question at all 😅

src/django/VLE/serializers.py
mjvkeulen: Added

src/django/VLE/models.py
mjvkeulen: Questions are simply a way to flag lines to get a discussion going during the review process. Whether that's in the form of a point of interest or question..

We currently serialize journals from the AssignmentSerializer (and children) with this filter applied if a `course` is provided as context. When do we want to filter an assignment's journals based on a specific course?  We currently also ignore that the user might be a participant in more than one of the assignments courses.

- We use `AssignmentSerializer` for the teacher journal overview (including group assignments)
- We use `Journal list` for the student journal overview (group assignment)

I moved the current Journal list filter to the queryset right below this filter for comparison. It is clear these are not even consistent (see `for_course` vs `for_course2`), a good reminder why centralizing such logic (on the respective model's QuerySet) is good practice.

Once the questions/discussion settle we can merge these filters.

src/django/VLE/models.py
mjvkeulen: Done

src/django/VLE/models.py
```diff
@@ -1890,6 +1880,44 @@ class Node(CreateUpdateModel):
...
+
+        if self.is_progress:
+            if not grade and not hasattr(self.journal, 'grade'):
+                raise VLEProgrammingError('Expired deadline check requires a journal grade')
+
+            if grade is None:
+                grade = self.journal.grade
+            return self.preset.target > grade and self.preset.due_date > timezone.now()
+
+        if self.is_entry:
```
Larspolo, line 1916: This property doesnt exist right?

mjvkeulen: It's defined on line 1892?

src/django/VLE/models.py
engelhamer: What do you propose?

src/django/VLE/serializers.py
```diff
@@ -6,15 +6,83 @@
...
+        default, value to be returned if the key is not preset in the cache, None by default
+
+    Returns a tuple
+        0: values of the cache keyed by name, None by default (or the given default)
+        1: (bool), whether the name was found in the instance cache and it was populated sucesfully
+    """
+    if hasattr(instance, '_prefetched_objects_cache') and name in instance._prefetched_objects_cache:
+        return instance._prefetched_objects_cache[name], bool(instance._prefetched_objects_cache[name])
+
+    return default, False
```
engelhamer, line 38: Why always directly return this default value as part of a tuple? It is not used above.

src/django/VLE/serializers.py
```diff
@@ -160,47 +307,86 @@ class Meta:
...
+            'due_date',
+            'lock_date',
+            'is_group_assignment',
+            'remove_grade_upon_leaving_group',
+            'can_set_journal_name',
+            'can_set_journal_image',
+            'can_lock_journal',
+            # Not used / missing: active_lti_id, lti_id_set, assigned_groups, format
+        )
+        read_only_fields = ('id', 'has_teacher_entries')
```
engelhamer, line 378: Why not `course_count`? Or `stats`?

src/django/VLE/serializers.py
```diff
@@ -212,156 +398,93 @@ class AssignmentSerializer(serializers.ModelSerializer):
...
+        permissions = [
+            'can_have_journal',
+            'can_grade',
+            'can_view_all_journals',
+            'can_edit_assignment',
+            'can_post_teacher_entries',
+            'can_manage_journal_import_requests',
+        ]
+
+        # QUESTION: Could the permission context be safely set for a queryset of assignments as well?
```
engelhamer, line 414: What do you expect can go wrong? I would like a bit more info.

src/django/VLE/serializers.py
engelhamer: Ah I don't think there's a valid reason for that. It can be changed but be sure to check the flatpickr components that use this.

src/django/VLE/models.py
```diff
@@ -1456,11 +1549,46 @@ class Meta:
...
+            # The can have journal check is pointless since we use the default manager for these calls
+            Q(authors__user__in=course.participation_set.filter(role__can_have_journal=True).values('user'))
+            | Q(authors__isnull=True),
+        ).distinct().order_by('pk')
+
+    def require_teacher_action(self):
+        """Returns journals which have an entry, and are awaiting grading or of which the grade needs publishing"""
+        return self.filter(
+            # QUESTION: Again significance of grade__grade__isnull=True?
+            # TODO: This (by accident) also catches node__entry__grade__isnull=True, make explicit
```
engelhamer, line 1584: TODO left.

src/django/VLE/models.py
```diff
@@ -1456,11 +1549,46 @@ class Meta:
...
+        return self.filter(
+            # The can have journal check is pointless since we use the default manager for these calls
+            Q(authors__user__in=course.participation_set.filter(role__can_have_journal=True).values('user'))
+            | Q(authors__isnull=True),
+        ).distinct().order_by('pk')
+
+    def require_teacher_action(self):
+        """Returns journals which have an entry, and are awaiting grading or of which the grade needs publishing"""
+        return self.filter(
+            # QUESTION: Again significance of grade__grade__isnull=True?
```
engelhamer, line 1583: I think this can be replaced by `grade__isnull=True`. We do not accept grades with empty input, but this was a potential (hacky) method to re-allow editing of entries a while ago. May be left from that feature.

engelhamer (changes requested):

src/django/VLE/serializers.py
mjvkeulen: I guess this is just leftover.

Serializers should be read-only where possible for speed (will not initialize validation rules for said fields).

- Method fields are read-only by design (stats).
- Id is a ReadOnlyField unless overridden.
 
So these could be removed without affecting anything, which I will do since you asked. I don't see much point in also changing this for the `read_only_fields` definitions. It's just making the default obvious (in regards to `id`).

src/django/VLE/serializers.py
mjvkeulen: If we serialize assignments only from the perspective of a single course, the permission checks could be cached for the first assignment queryset instance. If not, a permission cache should be created for each unique course.

src/django/VLE/models.py
mjvkeulen: So can I remove null=True from our Grade model?

src/django/VLE/models.py
mjvkeulen: Ok, will update the DB to reflect this rule.

src/django/VLE/serializers.py
mjvkeulen: That is by design if the attribute is not present in the instance cache the default is returned as value?

src/django/VLE/models.py
mjvkeulen: Merged the for_course filters, moved the for_course questions to an issue.

src/django/VLE/serializers.py
engelhamer: Ah sorry, misread this part.

src/django/VLE/serializers.py
```diff
@@ -750,15 +750,15 @@ class Meta:
 
     def get_unlock_date(self, node):
         if node.is_deadline and node.unlock_date is not None:
-            return node.unlock_date.strftime('%Y-%m-%dT%H:%M')  # QUESTION: Why does this format deviate?
+            return node.unlock_date.strftime(VLE.models.Field.ALLOWED_DATETIME_FORMAT)
```
engelhamer, line 752: Maybe move the `ALLOWED_DATETIME_FORMAT` to a higher level (not within the `Field` model) as this usage is not related to fields at all.

Larspolo (approved):
Can save enough say that the same values are serialized. Only change I could find what that the UserSerializer not the role serialized when no course is supplied where it would preciously supply `role: None`. Checked frontend and found no issues with that change there

engelhamer (approved):

PR #1048, opened 2020-11-15
Removed computed fields

mjvkeulen:
Moved all computed fields to annotations. Should significantly help in all scenario's where updates were slowing logic down, or where bulk updates were a PITA. 

There seems to be no significant speed loss for the sizes we work with. E.g. assignment view (retrieve) took 25 queries and was executed in 0.65s for 400 journals on my local machine.


Larspolo:
Nice

src/django/VLE/models.py
```diff
@@ -1605,13 +1589,156 @@ def allowed_journals(self):
...
+        )
+
+    def annotate_grade(self):
+        """"Annotates for each journal the rounded published grade sum of all entries as `grade`"""
+        grade_qry = Subquery(
+            Entry.objects.filter(
+                node__journal=OuterRef('pk'),
+                grade__published=True,
+            ).values(
+                'node__journal__pk',  # NOTE: Could be replaced by Sum(distinct=True) in Django 3.0+
```
Larspolo, line 1615: When do we wanna upgrade? I was also in need for a 3.0+ feature recently

src/django/VLE/models.py
```diff
@@ -1605,13 +1589,156 @@ def allowed_journals(self):
...
+        return self.annotate(grade=(Round2(F('bonus_points') + Coalesce(grade_qry, 0))))
+
+    def annotate_unpublished(self):
+        """"Annotates for each journal the count of entries which have an unpublished grade as `unpublished`"""
+        unpublished_entries_qry = Subquery(
+            Entry.objects.filter(
+                grade__published=False,
+                node__journal=OuterRef('pk'),
+            ).values(
+                'node__journal',
```
Larspolo, line 1633: Above you used `'node__journal__pk'`. Think both work right? Would prefer them all to be without the `__pk`

src/django/VLE/models.py
```diff
@@ -1605,13 +1589,156 @@ def allowed_journals(self):
...
+        ))
+
+    def annotate_name(self):
+        """
+        Annotates the journal name of each journal as `name`
+        Uses the stored name if found, else defaults to a concat of all author names.
+
+        NOTE: Makes use of `annotate_full_names` as a default, as such that annotation needs to happen first.
+        """
+        return self.annotate(name=Case(
```
Larspolo, line 1690: We could also call `annotate_full_name` on this. Calling it multiple times will not change the end query, only an extra function call, but think that is worth it for the more expected behaviour
```
>>> str(Journal.objects.all().annotate_name().annotate_name().annotate_name().query) == str(Journal.objects.all().annotate_name().query)
True
```

src/django/VLE/models.py
```diff
@@ -1605,13 +1589,156 @@ def allowed_journals(self):
...
+            When(Q(stored_name__isnull=False), then=F('stored_name')),
+            default=F('full_names'),
+            output_field=CharField(),
+        ))
+
+    def annotate_image(self):
+        """
+        Annotates for each journal the stored image or the first non default image for each journal as `image`
```
Larspolo, line 1698: You are officially a Django ORM wizard, well done!

src/django/VLE/models.py
```diff
@@ -2024,12 +2087,52 @@ def to_string(self, user=None):
...
+
+    def annotate_usernames(self):
+        return self.annotate(usernames=StringAgg('node__journal__authors__user__username', ', ', distinct=True))
+
+    def annotate_name(self):
+        """
+        Annotates for each entry the journal name as `name`
+        Uses the stored name if found, else defaults to a concat of all author names.
+
+        NOTE: Makes use of `full_names` annotation as a default, as such that annotation needs to happen first.
```
Larspolo, line 2119: Same here. I would prefer calling it just to be sure (or with an `if hasattr(self, 'full_names'): self.annotate_full_names` check as well)

src/django/VLE/serializers.py
```diff
@@ -714,6 +750,9 @@ def get_import_requests(self, journal):
             return journal.import_requests
         return None
 
+    def get_needs_lti_link(self, journal):
+        return journal.needs_lti_link
```
Larspolo, line 754: Why is this a serializer field and not directly from the model? Is that because you do not have the `outpu_field` param set in the `annotate_needs_lti_link` `ArrayAgg` query?

src/django/VLE/serializers.py
```diff
@@ -946,10 +992,10 @@ class Meta:
     ]
 
     journal_id = serializers.IntegerField(source='node.journal_id')
-    name = serializers.CharField(source='node.journal.name')
-    usernames = serializers.CharField(source='node.journal.usernames')
     grade = serializers.FloatField(source='grade.grade', default=None)
     published = serializers.BooleanField(source='grade.published', default=False)
+    usernames = serializers.CharField()
```
Larspolo, line 997: Could get the same comment here as well, `# Annotated fields need to be explicitly declared as they can't be derived from the Meta model`

src/django/VLE/serializers.py
```diff
@@ -1025,13 +1071,21 @@ def get_source(self, jir):
         if jir.source is None:
             return 'Source journal no longer exists.'
 
+        source = jir.source
+        if source.missing_annotated_field:
+            source = VLE.models.Journal.objects.get(pk=source.pk)
```
Larspolo, line 1076: Could we put this in the `__init__` of the `JournalSerializer`? Then every time the annotate fields are missing it automatically gets them. This will remove the need of using a separate query. e.g.: `JournalSerializer(Assignment.objects.first().journal_set.all(), many=True).data` currently results in an error

src/django/VLE/views/journal.py
```diff
@@ -78,7 +78,8 @@ def retrieve(self, request, pk):
             success -- with journals and stats about the journals
 
         """
-        journal = Journal.all_objects.get(pk=pk)
+        journal = Journal.all_objects.annotate_fields().get(pk=pk)
+        # QUESTION: Why do we reveal all the information below? These checks can be omitted imo, and use our manager
```
Larspolo, line 82: It is a user friendly wanring. If its the using requesting their own journal, its different from a teacher requesting a students journal right? Indeed would not add it because im lazy and its not that important, but removing it seems like just a step backwards

src/django/VLE/views/journal.py
```diff
@@ -186,7 +187,9 @@ def partial_update(self, request, *args, **kwargs):
             journal.bonus_points = bonus_points
             journal.save()
             grading.task_journal_status_to_LMS.delay(journal.pk)
-            return response.success({'journal': JournalSerializer(journal, context={'user': request.user}).data})
+            return response.success({
+                'journal': JournalSerializer(Journal.objects.get(pk=journal.pk), context={'user': request.user}).data
```
Larspolo, line 191: With the suggestion earlier of moving the missing annotations to the journal serializer init, I think this is no longer necessary

src/django/VLE/models.py
engelhamer: Write it down in https://zube.io/ejourn-al/ejournal/c/1123

src/django/VLE/views/comment.py
```diff
@@ -56,7 +56,7 @@ def list(self, request):
         entry_id, = utils.required_params(request.query_params, "entry_id")
 
         entry = Entry.objects.get(pk=entry_id)
-        journal = entry.node.journal
+        journal = Journal.objects.get(node__entry=entry)
```
engelhamer, line 59: Why is this better/faster/desirable?

src/django/VLE/views/journal.py
engelhamer: There is not really any information revealed here,  but the message is actually incorrect: if you were to retrieve a fake/non existing journal ID you will still be presented with a message stating that `This user currently does not have the correct rights to have a journal in this assignment`. Can be removed I guess, just yielding a 404 seems fine.

src/django/test/test_node.py
```diff
@@ -75,6 +75,7 @@ def test_node_properties(self):
     def test_open_deadline(self):
         assignment = factory.Assignment()
         journal = factory.Journal(assignment=assignment, entries__n=0)
+        journal = VLE.models.Journal.objects.get(pk=journal.pk)
```
engelhamer, line 78: Kind of counter intuitive that immediately retrieving what you created one line above yields a different result.

src/django/VLE/views/comment.py
Larspolo: This is to get the annotated fields

src/django/test/test_node.py
engelhamer: @Larspolo yeah but still: it is counter intuitive 😋

Larspolo: Maybe we could auto annotate inside the factory? No clue how /if its possible

engelhamer: Also in actual use (outside of tests) it may be easy to forget this, would indeed be nice to annotate immediately.

Larspolo: Problem is our limitation on how this is done + auto annotating will slow it down when it's not needed

src/django/VLE/serializers.py
mjvkeulen: So the cleanest solution I considered was overriding pythons magic methods `__getattr__` and `__getattribute__` on the Journal class. Problem was that it would obfuscate that SQL is being executed in the background and that performance is really suffering. 

```
def __getattr__(self, attrname):
    if attrname == "AnnotedField_X":
        try:
            return object.__getattribute__(self, attrname)
         except AttributeError:
            fetch_from_db()
```

So for the JIR serializer, your comment could be moved to `setup_eager_loading` where the prefetch of the target and source journals are also annotated. Use eager loading when possible, and init for other issues like permission caching. 

Not really a big issue atm for JIRs tho imo.

src/django/VLE/views/journal.py
mjvkeulen: Why could we just not move via the normal manager? (instead of all_objects manager)

src/django/test/test_node.py
mjvkeulen: I had a look at just moving this logic to the factory but could not see a (quick) solution. Open for suggestions of course.

src/django/VLE/views/comment.py
mjvkeulen: It also saves a query or two

src/django/VLE/models.py
mjvkeulen: There is a significant difference between using __pk as a param to OuterRef('pk'), and using .values for distinct aggregation. 

- `xxx__pk=OuterRef('pk')`, where `xxx` is foreign key, we can skip the `__pk`.
- `.values( 'node__journal__pk')` followed by an aggregration (e.g. Sum) is needed to aggregrate safely.

src/django/VLE/models.py
mjvkeulen: So nest the call remove the comments about the dependency? Fine by me

src/django/test/test_node.py
mjvkeulen: > Also in actual use (outside of tests) it may be easy to forget this, would indeed be nice to annotate immediately.
It is pretty much auto annotated outside of tests right? Provided the default manager is used. Difficulties arise when we acquire a Journal instance via a different route, e.g. `assignment.journal_set`, but we are used to this since the `allowed_journals` filter is applied in the same fashion.

src/django/VLE/serializers.py
mjvkeulen: The array aggregates did not play nice on some points, for example, see the modifications made to `get_stats`. As well as here, where selecting the correct serializer field to make it play like, for example, the plain string annotations did not work. It might well be possible of course.

src/django/test/test_node.py
engelhamer: Well the other comment I posted also is a bit counter intuitive, having to remember to perform `get` iso accessing properties directly

src/django/VLE/views/journal.py
Larspolo: My question is more, why would we change is? Its now more user friendly I think

engelhamer: Normal is ok. @larspolo Users shouldn't ever get to this error via in-app navigation?

Larspolo: If everything works as currently intended, I dont think so

src/django/test/test_node.py
mjvkeulen: The fields are annotated via DB query.

The other comment simply shows how the abstraction of the ORM can lead to such misunderstandings. What does the ORM do if you access a foreign key of the model instance (e.g. `node.journal`)? You are basically executing SQL, select * from X where id = Y, where the row of table data of X is represented as a model instance of type X. Without explicit annotations, why would we expect there to be magical additional columns?

src/django/VLE/serializers.py
Larspolo: > Since we use the Journal manager to both filter on allowed journals and to annotate some fields, I think being forced to work via the manager itself is a good thing.

:+1: resolving

Larspolo (approved):

engelhamer (approved):

PR #1051, opened 2020-11-18
Githooks

mjvkeulen:
A possible structure that allows us to share Githooks. 

Now if one forgets to run Isort before committing one is more easily reminded and can always amend their last commit.


engelhamer (approved):

Larspolo (approved):

.githooks/pre-commit
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+make isort
```
Larspolo, line 3: Do we also want to add linter checks here?

Larspolo (commented):
Do we also wa

mjvkeulen:
Lets try this via mail.

Linting could be included. Problem is that this does nog add changed files
tot your commit as is.

Op vr 20 nov. 2020 5:15 p.m. schreef Lars <notifications@github.com>:

PR #1052, opened 2020-11-18
Allow the title of a teacher entry to be edited after posting

mjvkeulen:


engelhamer (approved):
Maybe we should do something about whitespace before / after the title. This will now be considered as a 'change'.

Also added an edit button for editing, since I am unsure whether the input field permanently in  view will be clear enough (and fixed margin display bug).

Larspolo (approved):
Im scared of leaving design comments without any good suggestions, so Im also approving. But I feel like it is not fully clear for all users what this edit button does. On almost all buttons we have a text added to the button to identify what the button does. This is not only excluded, but its an edit icon while you are already kinda editing it. I dont think I would be able to guess what it would do before clicking on it.

How does @engelhamer think about this?

engelhamer:
> Im scared of leaving design comments without any good suggestions, so Im also approving. But I feel like it is not fully clear for all users what this edit button does. On almost all buttons we have a text added to the button to identify what the button does. This is not only excluded, but its an edit icon while you are already kinda editing it. I dont think I would be able to guess what it would do before clicking on it.
> 
> How does @engelhamer think about this?

Added text to buttons.

image

PR #1053, opened 2020-11-23
Moved to Github actions and moved commands to submodule

Larspolo:
## Description
This moves it to a private submodule which allows for easier non-app related commands

https://github.com/eJourn-al/eJournal-commands


engelhamer (approved):

engelhamer:
Travis does not like this yet.

mjvkeulen (changes requested):
- Another/parallel job to run our linting
- Github actions run in parallel by default (apparently up to 20 jobs for open source/free tier?). We could speed up our tests by dividing the test files over multiple actions. Here is an example project which automates this process (even the merging of the coverage reports) 
https://github.com/jerry-git/pytest-split-gh-actions-demo/blob/master/.github/workflows/test.yml
Which seems quite doable on a low time budget?

Makefile
```diff
@@ -36,7 +36,7 @@ test-back:
 	${venv_activate} \
 	&& flake8 ./src/django \
 	&& ./src/django/manage.py check --fail-level=WARNING \
-	&& pytest ${TOTEST} src/django/test/
+	&& pytest ${TOTEST} src/django/test/ --store-durations
```
mjvkeulen, line 39: I think this is a left over for the upload? Or should this be continually updated?

Larspolo: Would like it to be continually updated. Every time someone adds a test, it will then be automatically added to test test durations

mjvkeulen: Does this slowdown running tests locally significantly? 

This also means an additional file to commit on each make test run right? (Not sure if we would like git changes to be polluted with this file on each PR / commit)

mjvkeulen: I would be in favor of periodically uploading a new version, just so we keep git changes a bit cleaner. From the split test doc: if any test is missing from the duration.json file, it is assumed to be of average execution time. This seems fine for us with a periodical commit that updates this file?

Larspolo: Good point, moved the auto update to seperate command, andcreated an issue

.github/workflows/build.yml
```diff
@@ -0,0 +1,146 @@
...
+        ports:
+          - 5432:5432
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.ACTIONS_TOKEN }}
+          submodules: recursive
+
+      - name: Set up Python 3.6.9
```
mjvkeulen, line 58: Occurs 3 times, good candidate to be reused with an anchor? Assuming this works as valid yaml?

.github/workflows/build.yml
```diff
@@ -0,0 +1,146 @@
...
+          path: ${{ env.pythonLocation }}
+          key: python-${{ hashFiles('requirements/base.txt') }}-${{ hashFiles('requirements/ci.txt') }}
+
+      - name: Install dependencies
+        run: pip3 install -r requirements/ci.txt
+
+      - uses: bahmutov/npm-install@v1.4.5
+        with:
+          working-directory: src/vue
+          useLockFile: false
```
mjvkeulen, line 142: How does this work? Should our node env also be cached?

.github/workflows/build.yml
```diff
@@ -0,0 +1,146 @@
...
+  SENTRY_SENTRY_ORG: ejournal
+  SENTRY_PROJECT: ejournal
+  MAILGUN_API_KEY: placeholder
+  MAILGUN_SENDER_DOMAIN: placeholder
+  API_URL: 'http://localhost:8000'
+  RELEASE_VERSION: '1.0.0'
+  CODE_VERSION: 937aa7be86b36e1d30444336c073bbf6a4e72b2f
+  CI_CD: true
+
+jobs:
```
mjvkeulen, line 27: I could not find much about splitting jobs across different yaml files, do you know if that's possible?

mjvkeulen (commented):
Great to have CI working again! Seems like we got quite a decent speed going again too.

Just some remarks remaining regarding the build.yml file, which is the shiny new addition of course. 

Ideally the environment used during CI matches our local and production environment closely. 
Would it be possible to tweak the makefile commands for testing front and back (tests + lints) so we use the same commands for both CI and local env?

.github/workflows/build.yml
Larspolo: `useLockFile: false` is saying that we do not have a lock file, and to use the package.json instead

Might be posisble, but as is is fast enough. Although if you want you can try to make it cashed as well, should be useful once we start adding front end testing as well

.github/workflows/build.yml
Larspolo: Im not too familliar with yaml, count be possible, but what I tried didnt work

.github/workflows/build.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Cache python environment
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.pythonLocation }}
+          key: python-${{ hashFiles('requirements/base.txt') }}-${{ hashFiles('requirements/ci.txt') }}
+
+      - name: Install dependencies
+        run: pip3 install -r requirements/ci.txt
+
+      - name: Run tests
```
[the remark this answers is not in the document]
Larspolo: Thats all the tests we have though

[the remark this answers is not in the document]
mjvkeulen: Fair enough, calling lints tests is unfair.

.github/workflows/build.yml
mjvkeulen: I looked and atm github actions simply doesn't support anchors. There are composite actions, so we could split the jobs file but it looks like slightly too much effort for the gain.

.github/workflows/build.yml
mjvkeulen: So using the composite action `bahmutov/npm-install@v1.4.5` already sorts out cache for us.

.github/workflows/build.yml
mjvkeulen: Composite actions

.test_durations
```diff
@@ -0,0 +1,1486 @@
+[
```
engelhamer, line 1: Why does this file have to be commited?

Larspolo: To better schedule the tests into parallel actions. Say 1 test takes 10 minutes, then a single action would be created specifically for that test

mjvkeulen (approved):

PR #1056, opened 2020-11-29
Pre release changes (1.5)

mjvkeulen:
- Fixes Z#1379
- Fixed Z#1381
- Fixes Z#1384
- Fixes Z#1385
- Fixes Z#1390
- Fixes Z#1391
- Fixes Z#1392
- Fixes Z#1394
- Fixes Z#1401
- Fixes Z#1405
- Fixes Z#1406
- Fixes Z#1407
- Fixes Z#1409


src/django/VLE/views/journal.py
```diff
@@ -108,14 +108,12 @@ def create(self, request):
...
-
-            journals.append(Journal.objects.create(
-                assignment=assignment,
-                author_limit=author_limit,
-                stored_name=self._get_name(name, amount, assignment)
-            ))
+        journals = [Journal(
+            assignment=assignment,
+            author_limit=author_limit,
+            stored_name=self._get_name(name, amount, assignment),
```
Larspolo, line 114: This doesnt work now, see https://zube.io/ejourn-al/ejournal/c/1394

Larspolo (changes requested):

src/vue/src/views/Home.vue
```diff
@@ -120,6 +120,15 @@ export default {
...
             this.$refs[ref].hide()
         },
+        courseDateDisplay (course) {
+            let display = ''
+
+            if (course.startdate) { display += course.startdate.substring(0, 4) }
+            if (course.startdate && course.enddate) { display += ' - ' }
+            if (course.enddate) { display += course.enddate.substring(0, 4) }
+
+            return display
```
Larspolo, line 130: Now you dont know whether it is the end or start date of the course. Elsewhere this is fixed by doing `if (course.startdate || course.enddate) { display += ' - ' }`

mjvkeulen: I don't see the fix? We don't have a clear distinction between start and end if only one date is shown elsewhere afaik?

Larspolo: `src/vue/src/utils/generic_utils.js:27`

mjvkeulen: Having a single place for this logic has my preference as well, I don't see how `courseWithDatesDisplay` as is would solve this issue. 

`- date`
`date -` 
both of these are not really practical indications of either start or stop.

I think we need a proper solution?

Larspolo: Something better is preferred, but until then reusing existing functions is also preferred

mjvkeulen: Go for it, the function you refer to will need to be split. Currently it returns both the course name and date as a single string for use in multiselects.

Still curious how you will make a distinction between start and end date of only ons is available.

Larspolo: Before the "-" -> start date
After the "-" -> end date

engelhamer:
image

engelhamer (approved):

Larspolo (approved):

PR #1057, opened 2020-12-02
Added display names for timeline deadlines and progress goals

mjvkeulen:
Fixes Z#1360

image

image


Larspolo (commented):
Would expect (and like) for nodes to be synchronized with the template until the name is changed from the default. Now when you add a node, then change the template name, the node also needs to be renamed. Is current behaviour intended?

mjvkeulen:
This PR purposefully moves away from template name == timeline display name.

Making a preset node's display name explicit to the user will help to alleviate the current problem: template duplication purely to achieve different names in the timeline. It subtly forces the users to see template names as different entities than the display names in the timeline, hopefully resulting in more clever reuse. By more clearly creating two different entities, timeline parts (preset nodes) and templates, users can approach templates more abstractly. E.g. creating a report type template, focusing more on the types of fields desired than the direct activity which is handled by creating the preset nodes.
Along this line of thought, syncing template names and display names again creates this strongly linked entity. 

This PR comes with sensible defaults, e.g. the template name is used as the default preset display name for a deadline preset, or `Progress goal` is used as a default display name for a progress node. A bonus of making preset node display names non-nullable is a more clean schema.

Larspolo:
Then we shouldn't auto fill the display name field with "Progress goal" or the template name for that matter. Now it doesnt follow what you are trying to solve

mjvkeulen:
> Then we shouldn't auto fill the display name field with "Progress goal" or the template name for that matter. Now it doesn't fully follow what you are trying to get to.

This is true, but those defaults are there to more subtly transition. Curious what @engelhamer thinks, I would be fine without the defaults as well.

> Also would move the display name above the template

Also fine by me, provided Engel agrees.

Larspolo:
Wouldnt go subtle, already all names will be converted, and then everybody sees the new feature, and will likely be very glad there finally is one. No need anymore to check the changelogs

src/django/VLE/migrations/0070_presetnode_display_name.py
```diff
@@ -8,7 +8,7 @@ def setDefaultDisplayNames(apps, schema_editor):
 
     for preset in PresetNode.objects.all().select_related('forced_template'):
         if preset.is_deadline:
-            preset.display_name = node.forced_template.name
+            preset.display_name = preset.forced_template.name
```
Larspolo, line 11: For these migration testing would be nice. There are already a couple examples

src/django/VLE/migrations/0070_presetnode_display_name.py
```diff
@@ -8,7 +8,7 @@ def setDefaultDisplayNames(apps, schema_editor):
...
         if preset.is_progress:
             preset.display_name = 'Progress goal'
```
Larspolo, line 13: Can be shortened to sth like 
```
PresetNode.objects.filter(type == 'entrydeadline').update(display_name=F('forced_template__name'))
PresetNode.objects.filter(type == 'progress').update(display_name='Progress goal')
```
Not tested, nor mandatory, but aligns with the change to more Django ORM usage over Python stmts

mjvkeulen: I mean I know I can do this in a single SQL statement using update and case, but writing this was faster. This code is executed only once, and performance is not really an issue.

src/django/VLE/migrations/0070_presetnode_display_name.py
mjvkeulen: I didn't add a test because the logic is simple and just a few lines. I feel confident that this block will work especially after two reviewers went over it. 
On top of this, the migration does not alter/delete existing data. In the worst case, a rollback or correction would be straightforward. 

Sometimes writing tests just for the sake of it is not worth it imo?

engelhamer:
> > Then we shouldn't auto fill the display name field with "Progress goal" or the template name for that matter. Now it doesn't fully follow what you are trying to get to.
> 
> This is true, but those defaults are there to more subtly transition. Curious what @engelhamer thinks, I would be fine without the defaults as well.
> 
> > Also would move the display name above the template
> 
> Also fine by me, provided Engel agrees.

I think it's indeed best to keep the preset name explicit, and not possible to nullify it. This way the preset name will always be the same for everyone in the same preset, whereas it could change if changes to the template are made (new entries use the new template, but the existing ones always have the old in our current setup).

src/django/VLE/timeline.py
```diff
@@ -111,6 +112,7 @@ def get_deadline(journal, node, user):
 def get_progress(journal, node):
     """Convert progress node to dictionary."""
     return {
+        'display_name': node.preset.display_name,
```
engelhamer, line 115: Why do we have `nID` camelcase and `display_name` snake case?

src/vue/src/components/format/FormatPresetNodeCard.vue
```diff
@@ -161,22 +86,106 @@
...
+                <reset-wrapper v-model="currentPreset.lock_date">
+                    <flat-pickr
+                        v-model="currentPreset.lock_date"
+                        class="multi-form full-width"
+                        :config="lockDateConfig"
+                    />
+                </reset-wrapper>
+            </b-col>
+        </b-row>
+        <div v-else>
```
engelhamer, line 171: Template iso div?

src/django/VLE/timeline.py
Larspolo: nID is old, one way this needs to be replaced as well

src/vue/src/components/format/FormatPresetNodeCard.vue
mjvkeulen: This is existing code just repositioned right? (container for due date). I would argue a refactor of unneeded divs -> templates is best not to mix in this PR.

src/django/VLE/factory.py
```diff
@@ -190,7 +190,8 @@ def make_progress_node(format, due_date, target):
     format -- format the node belongs to.
     due_date -- due_date of the node.
     """
-    node = VLE.models.PresetNode(type=VLE.models.Node.PROGRESS, due_date=due_date, target=target, format=format)
+    node = VLE.models.PresetNode(
+        type=VLE.models.Node.PROGRESS, due_date=due_date, target=target, format=format, display_name='Progress goal')
```
Larspolo, line 194: Would be nice if this were a parameter in the function. Or we could just remove the whole function as its only used once

Larspolo (approved):

src/vue/src/components/format/FormatPresetNodeCard.vue
engelhamer: This is just a change of 24 characters total. Noticed the opportunity because it was in the diff.

engelhamer (approved):

PR #1058, opened 2020-12-03
Categories 

mjvkeulen:
First draft used to discuss the DB schema changes. 

Once settled same PR can be used to move to actual implementation


engelhamer:
Obsolete since #1067

PR #1059, opened 2020-12-03
Added support for Canvas administrator role

Larspolo:
## Description
Added support for Canvas administrator role


src/django/test/test_lti_launch.py
```diff
@@ -656,12 +656,22 @@ def test_get_lti_params_from_jwt_multiple_roles(self):
             assert_msg='When a student tries to create a new course, with also a new Instructor role, \
                         it should grand its permissions')
 
+    def test_get_lti_params_from_jwt_admin_role(self):
```
mjvkeulen, line 659: Would like this be searchable with a pattern like `test_.*administrator` since it specifically tests that role.

mjvkeulen (approved):

engelhamer (approved):

PR #1064, opened 2020-12-09
Added pagination to the user overview

mjvkeulen:
Pagination can be configured via URL query parameters:
- `page` (int): the window of the query set to serialize.
- `page_size` (int): size of the window.
- `order_by` (str): order by argument, can be descending (-).
- `filter` (str): filter to apply to the query, for the user table this performs a case insensitive search for the columns  `full_name`, `username`, and `email`.

Changing pages is compatible with the set filter and order.

An upper limit to the desired  `page_size` can be configured via `max_page_size`, this can be done globally (`PAGE_SIZE`), or per view, as is done in this PR (see views/user.py).

image

image


src/vue/src/components/admin/UserOverview.vue
```diff
@@ -1,176 +1,180 @@
...
+                    sortable: true,
+                },
+                {
+                    key: 'is_active',
+                    label: 'Active',
+                    sortable: true,
+                },
+                {
+                    key: 'action',
+                    label: '',
```
engelhamer, line 130: This column should not be sortable: clicking in the red square causes a 500 error.

click

src/vue/src/components/admin/UserOverview.vue
```diff
@@ -1,176 +1,180 @@
...
-                    this.getAllUsers()
+            userAPI.update(
+                user.id,
+                { is_teacher: false },
+                { customSuccessToast: `Successfully removed ${user.full_name} as a teacher.` },
+            )
+                .finally(() => { this.$refs.table.refresh() })
+        },
+        provider (context, callback) {
+            console.log(context)
```
engelhamer, line 160: Debug print left?

engelhamer (changes requested):

engelhamer (approved):

Larspolo (approved):
Would prefer the table only showing the load icon after halve a second, this reduces the flash, but its not much used, and already a way better version then currently, so still approving

PR #1066, opened 2020-12-17
Changed toast error to sentry warning and display invalid url as text

Larspolo:
## Description
Changed toast error to sentry warning and display invalid url as text.

Previously you would get a big open white space with a weird toast error, and we wouldnt even know about it. Now it gets sent to Sentry while frontend it "hides" the problem by displaying the youtube URL as text


src/vue/src/components/entry/EntryFields.vue
```diff
@@ -192,7 +199,11 @@ export default {
             if (match && match[2].length === 11) {
                 return `https://www.youtube.com/embed/${match[2]}?rel=0&amp;showinfo=0`
             } else {
-                this.$toasted.error('A YouTube video field contained an invalid URL.')
+                this.$store.commit('sentry/CAPTURE_SCOPED_MESSAGE', {
+                    level: 'warning',
+                    msg: 'A YouTube video field contained an invalid URL.',
+                    extra: { url },
```
mjvkeulen, line 205: Would a link to the entry as additional extra argument be helpful in debugging?

Larspolo: That information was not available as is, but it could be changed a bit to also be available

engelhamer (approved):

src/vue/src/components/entry/EntryFields.vue
Larspolo: Added field ID, feel free to add more if you would like

mjvkeulen (approved):

PR #1067, opened 2020-12-19
Added categories which can be used to filter a timeline, and improved the assignment editor

mjvkeulen:
# New: Categories

Categories are a way of grouping activities which contribute to a particular concept/skill/learning goal. Category functionality is likely to be expanded in the near future, for now categories can be used for the following:

- Categories can be created in the assignment format editor.
- A category can be given a default set of templates. If a student uses such a template to create a journal entry, the category will be linked to the entry by default. It's possible to configure on a per template basis, whether the student can add or remove other categories to the entry.
- Once an entry is graded, categories can only be changed by educators.
- The entries of a journal can be filtered by category.

## Outstanding Questions
- What else would we like to do with categories, e.g. visualize progress within a specific category?
- Apart from which nodes are shown in the timeline, what else should the active timeline filter impact?
- Should timeline category filters be persistent? Currently not beyond the visit and assignment navigation.
- Should it be possible to filter for specific categories on the assignment overview page?
- Should updating a category's templates affect existing entries using that template?
- How will we tackle setting archived templates to their corresponding chain?

## UI Preview
Prone to changes pending review @engelhamer 

image

image

image

image

image

image

<hr/>

TODO:
- [x] Make categories reactive after assignment format change (e.g. add / delete category)
- [x] Add support for importing categories (atleast alongside assignment copy)
- [x] Persist and clean timeline filter when appropriate (switch assignment, a filtered category no longer exists)
- [x] Change or improve color picker styling in manage categories
- [x] More consisting category color generation (switch from \#RGB to HSL colors?)
- [x] Teacher entries use template chain where applicable
- [x] Visual indication of categories in timeline
- [x] Hide category filter and category add button if no assignment categories exist
- [x] @Lars hack category add to templates frontend


src/django/test/test_category.py
```diff
@@ -0,0 +1,235 @@
...
+            api.patch(self, 'categories/edit_entry', params={**params}, user=teacher)
+
+            lock_date = self.assignment.lock_date
+            Assignment.objects.filter(pk=self.assignment.pk).update(lock_date=timezone.now())
+            api.patch(self, 'categories/edit_entry', params={**params}, user=teacher)
+            Assignment.objects.filter(pk=self.assignment.pk).update(lock_date=lock_date)
+
+        def test_as_admin():
+            admin = factory.Admin()
+            # QUESTION: Should superusers be blocked from grading (and thus fail user.has_permission('can_grade'))?
```
Larspolo, line 204: imo not, may be useful some day. Gradebook will clearly state that the grade is set by admin, we can always remove those grades. Why would you want to limit it?

src/django/test/test_teacher_entry.py
```diff
@@ -269,6 +270,64 @@ def test_update_teacher_entry(self):
...
+        add_categories['category_ids'] = [self.category.pk, self.category2.pk]
+        api.update(self, 'teacher_entries', params=add_categories, user=self.teacher)
+        # Teacher entries are correctly added (including setting the author)
+        assert EntryCategoryLink.objects.filter(category=self.category, entry=te, author=self.teacher).exists()
+        assert EntryCategoryLink.objects.filter(category=self.category2, entry=te, author=self.teacher).exists()
+        assert EntryCategoryLink.objects.filter(category=self.category, entry=entry, author=self.teacher).exists()
+        assert EntryCategoryLink.objects.filter(category=self.category2, entry=entry, author=self.teacher).exists()
+
+        # QUESTION: Do you agree with the behaviour outlined in the blocks below (teacher entry category updates
+        # are kept in sync with all associated entries)?
```
Larspolo, line 309: Yes, good catch.

Only difficult case I can think of when a teacher creates a non-graded teacher entry with only template text for the students to then change themself later on. Maybe an entry should be able to add extra categories to the already existing categories from the connected teacherentry? (but not remove the categories attached to the TE)

src/django/VLE/utils/template.py
```diff
@@ -0,0 +1,91 @@
...
+            new_ids[data['id']] = new_template.pk
+
+    return new_ids
+
+
+def delete_or_archive_templates(templates_data):
+    template_ids = [template['id'] for template in templates_data]
+
+    Template.objects.unused().filter(pk__in=template_ids).delete()
+    Template.objects.filter(pk__in=template_ids).update(archived=True)
```
Larspolo, line 91: Are all these functions not more suitable to be placed inside the model? Some being static methods.

src/django/VLE/utils/template.py
```diff
@@ -0,0 +1,91 @@
...
+        len(field_set) != len(data['field_set'])
+        or any(_field_concrete_fields_updated(field, data) for field, data in zip(field_set, data['field_set']))
+    )
+
+
+def _should_be_archived(template, data):
+    assert data['id'], 'Template can only be updated if it was already created'
+
+    return (
+        # QUESTION: What is the advantage of archiving the template if only the concrete fields change?
```
Larspolo, line 41: A thing I can see it that it creates unexpected behavior. Accepting that templates from already created entries do not get changed, it would be weird if the title & description did change

Larspolo (commented):
This is not a complete review

src/django/VLE/management/commands/preset_db.py
```diff
@@ -316,11 +316,32 @@ def gen_format(self):
         # Group Assignment
         factory.ProgressPresetNode(format=self.group_assignment.format, target=10)
 
+    def gen_categories(self):
+        for a in self.assignments:
+            templates = list(a.format.template_set.all())
+            template_count = len(templates)
+
+            for _ in range(random.randint(4, 10)):
```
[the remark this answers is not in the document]
mjvkeulen: Feel free to hardcode your preferred setup

src/django/VLE/models.py
```diff
@@ -2300,6 +2347,27 @@ def is_editable(self):
     def is_graded(self):
         return not (self.grade is None or self.grade.grade is None)
 
+    def add_category(self, category, author):
+        """Should be used over entry.categories.add() so the author is set and the link can be reused."""
+        link = EntryCategoryLink.objects.filter(entry=self, category=category).first()
```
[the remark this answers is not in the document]
mjvkeulen: The kwargs form the filter to look whether the category link already exists, whereas the author can vary. Its a useful function nonetheless. 

Your remark also made me spot the error: the author should not be edited if the link already exists.

src/django/VLE/models.py
```diff
@@ -2435,11 +2503,69 @@ def to_string(self, user=None):
...
+class TemplateChain(CreateUpdateModel):
+    """Identifies multiple templates which were updated as belonging to the same template"""
+    format = models.ForeignKey(
+        'Format',
+        on_delete=models.CASCADE
+    )
+
+
+class TemplateQuerySet(models.QuerySet):
+    def create(self, format, *args, **kwargs):
```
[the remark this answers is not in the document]
mjvkeulen: Format is non optional for both the chain as the template itself

src/django/VLE/views/category.py
```diff
@@ -0,0 +1,144 @@
...
+        category = Category.objects.get(pk=pk)
+        entry = Entry.objects.select_related('node__journal__assignment', 'template').get(pk=entry_id)
+        template = entry.template
+        assignment = entry.node.journal.assignment
+
+        if request.user.has_permission('can_grade', assignment):
+            pass  # A category can always be changed by someone with the permission to grade.
+        elif request.user.has_permission('can_have_journal', assignment):
+            if template.fixed_categories:
+                return response.forbidden('This entry\'s template\'s categories are fixed.')
```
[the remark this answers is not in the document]
mjvkeulen: This should only happen if you were fiddling with requests manually.

Updated to: This entry's categories are locked. 

Any better?

src/django/test/test_category.py
mjvkeulen: This line is fine by me, only thing admin can't do is create entries then right?

src/django/test/test_teacher_entry.py
mjvkeulen: > Maybe an entry should be able to add extra categories to the already existing categories from the connected teacherentry? (but not remove the categories attached to the TE)

Is this prefered over syncing? Would mean a new DB flag on the link.

src/django/VLE/utils/template.py
mjvkeulen: Which one? I already move the query for unused templates to the template query set. 

The full block? That would mean tightly coupling with the data format (template_ids). 
The last two lines?

src/django/VLE/models.py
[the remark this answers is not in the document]
Larspolo: So then `get_or_create` will do right? Where author will only be udes when creating and not when checking if it already exists

src/django/VLE/views/category.py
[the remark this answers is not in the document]
Larspolo: Way better, that will do for me

src/django/test/test_category.py
Larspolo: They cannot have a journal, so creating entries is kinda excludede then as well, I suppose they could still create TE's through

src/django/test/test_teacher_entry.py
Larspolo: synching with the TE's, but potentially added more on personal entry level I ment. But im fine with this not working, but then dont forget to present that frontend somewhere

src/django/VLE/utils/template.py
Larspolo: I ment the whole file

src/django/VLE/models.py
```diff
@@ -2291,6 +2306,38 @@ class Entry(CreateUpdateModel):
...
 
+    categories = models.ManyToManyField(
+        'Category',
+        related_name='entries',
+        through='EntryCategoryLink',
+        through_fields=('entry', 'category'),
+    )
+
+    @staticmethod
+    def validate_categories(category_ids, assignment, template=None):
```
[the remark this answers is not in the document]
mjvkeulen: It validates category ids which are requested to be associated with an Entry/TeacherEntry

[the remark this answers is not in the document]
mjvkeulen: We currently mix models and util modules. 

I favor using models since generally you validate data before updating/creating a model instance.

src/django/test/test_category.py
mjvkeulen: Keeping it as is, removing question.

src/django/test/test_teacher_entry.py
mjvkeulen: I think I prefer the clarity of keeping the categories synced with the teacher entry itself. 
Added this information to the manage teacher entry modal.

src/django/VLE/utils/template.py
mjvkeulen: I am torn:

- Many of these functions work a specific data format
- Many of these functions are never called directly

What do you think?

src/django/VLE/utils/template.py
mjvkeulen: I guess it doesn't really hurt us either way, resolving.

Larspolo:
I would like to be able to edit the catogories form the template editor as well

src/django/test/test_template.py
```diff
@@ -0,0 +1,201 @@
...
+            'The full chain of a single template is the template itself'
+
+        template2_of_chain1 = Template.objects.create(format=self.format, chain=template1_of_chain1.chain)
+        assert template2_of_chain1.chain == template1_of_chain1.chain
+        assert set(Template.objects.full_chain(template1_of_chain1)) == set([template1_of_chain1, template2_of_chain1])
+        assert set(Template.objects.full_chain(template2_of_chain1)) == set([template1_of_chain1, template2_of_chain1])
+
+        template2_of_chain2 = Template.objects.create(format=self.format, chain=template1_of_chain2.chain)
+        assert set(Template.objects.full_chain([template1_of_chain1, template1_of_chain2])) \
+            == set([template1_of_chain1, template2_of_chain1, template1_of_chain2, template2_of_chain2])
```
Larspolo, line 54: Comments about why you assert stuff is really helpful. I see us doing stuff like `"full chains should be same as all templates of chains"`, but even then I think the assert message should focus on the why `"full chain should now also include the newly created template in chain 2"`. I never really payed much attention to what I type in the assert messages, but now reading this code, I really miss the _why_ aspect of it all.

Same above: `'The full chain of a single template is the template itself'` is alright, but sth like `The full chain should not contain templates from another chain` is more representative on why the assert exists

src/django/test/test_teacher_entry.py
```diff
@@ -269,6 +270,66 @@ def test_update_teacher_entry(self):
...
+        assert te.categories.exists()
+        assert entry.categories.exists()
+        api.update(self, 'teacher_entries', params=remove_categories, user=self.teacher)
+        assert not te.categories.exists(), 'Categories are removed from the teacher entry itself'
+        assert not entry.categories.exists(), 'Categories are removed from the entry generated by the teacher entry'
+
+        add_categories = deepcopy(valid_update_params)
+        add_categories['category_ids'] = [self.category.pk, self.category2.pk]
+        api.update(self, 'teacher_entries', params=add_categories, user=self.teacher)
+        # Teacher entries are correctly added (including setting the author)
```
Larspolo, line 301: These comments are nice, but not places everywhere. probably because of the length of the PR, but it really helps debugging especially these longer "unit" tests

src/django/test/test_journal.py
```diff
@@ -45,7 +45,6 @@ def test_journal_factory(self):
             'Generating a journal whose assignment hold no courses, does indeed generate no courses'
         assert Assignment.objects.filter(journal=journal).exists() and Assignment.objects.count() == a_count + 1, \
             'A single assignment is generated when generating a journal'
-        # QUESTION: Lars is het logisch dat hier all_objects nodig is?
```
Larspolo, line 48: Ja ;)

src/django/VLE/views/teacher_entry.py
```diff
@@ -174,6 +206,13 @@ def _create_new_entries(self, teacher_entry, journals_data, author):
         ]
         entries = Entry.objects.bulk_create(entries)
 
+        entry_category_links = [
+            EntryCategoryLink(entry=entry, category_id=category_id, author=author)
+            for category_id in category_ids
+            for entry in entries
+        ]
+        EntryCategoryLink.objects.bulk_create(entry_category_links)
```
Larspolo, line 214: Sth for inside the entry (bulk) create function?

src/django/VLE/views/teacher_entry.py
```diff
@@ -8,12 +8,28 @@
...
 from VLE.utils.error_handling import VLEBadRequest
 
 
+def _update_categories_of_existing_entries(entries, author, new_category_ids, existing_category_ids):
+    """
+    Keeps the categories of all existing entries synced with those set on the teacher entry.
+
+    Does not modifying the entry category links of entries which already belong to the desired categories.
+    """
+    EntryCategoryLink.objects.filter(entry__in=entries).exclude(category_id__in=new_category_ids).delete()
```
Larspolo, line 23: Should this not be in `existing_category_ids`? also really feel like this should all be dealt in the model functions

src/django/VLE/views/teacher_entry.py
```diff
@@ -105,6 +126,9 @@ def partial_update(self, request, pk, *args, **kwargs):
         if title is None or len(title) == 0:
             return response.bad_request('Title cannot be empty.')
 
+        category_ids = Entry.validate_categories(category_ids, teacher_entry.assignment)
```
Larspolo, line 129: Docstring nor function name explains why you need to set `category_ids` to the output of the function. Would also be more fitting for an assignment function `teacher_entry.assignment.validate_categories(category_ids)`.

And lastly just a question, why is this validation not done inside the save function? As is previously done with other save functions

src/django/test/test_template.py
mjvkeulen: I very much agree breaking tests should be super easy to debug, and good comments or assert messages go a long way. 
A large part is also structuring tests simply, like a story. Make sure they are not too lengthy and decent variables are used. 

Since in the first example you used the correction is not quite accurate ("full chain should now also include the newly created template in chain 2"), when in fact it is about full_chain yielding all templates part of any chain**s** it is provided. It is clear the test is not as obvious as I thought. How about the revision?

src/django/VLE/views/teacher_entry.py
mjvkeulen: I mean I could move it there, but there no side effects or any additional configuration settings happening right?

src/django/VLE/views/teacher_entry.py
mjvkeulen: This is a table-level operation, not row level. So could be moved to the queryset I guess.

This is an update function, `existing_category_ids` are those ids which currently exist, `new_category_ids` are the ids which the teacher would like to set.

src/django/VLE/views/teacher_entry.py
mjvkeulen: > Docstring nor function name explains why you need to set category_ids to the output of the function. 

Returns the validated ids as a set. Might be tight coupling, but reading `ids = validate(ids)` must strongly suggest that `ids` are now validated tho right?

> Would also be more fitting for an assignment function teacher_entry.assignment.validate_categories(category_ids).

You validate category ids before linking them to an `Entry`, not sure why you prefer moving this logic to the assignment model?

> Validation in save functions
- Easily bypassed (use anything sql related, e.g. update, bulk_create etc)
- Hides explicit validation steps (making it easy to forget).
- What do you do when you use a route which does not make use of the model save? You still need to validate, do you duplicate the validation logic?

This is why I already moved quite a bit of validation logic outside of save functions. For example in the models file see, Assignment.validate, JIR.validate, Category.validate.

src/django/VLE/views/teacher_entry.py
mjvkeulen: > also really feel like this should all be dealt in the model functions

What is all? This (`_update_categories_of_existing_entries`) particular function? 

If so, where do you feel it should go, and if we move it to a query_set operation what should it be called?

mjvkeulen:
> I would like to be able to edit the categories form the template editor as well
> 
> image
> Categories should already be visible in this list

I prefer keeping the logic regarding categories in one place. For me, it brings clarity, want to do something with categories? Go to manage categories.

> Why cant you edit categories after adding a spacebar to the assignment description?
I will make this more strict.

src/django/VLE/views/teacher_entry.py
Larspolo: > Returns the validated ids as a set. Might be tight coupling, but reading ids = validate(ids) must strongly suggest that ids are now validated tho right?

Mostly for other times when using this function it would not feel natural of resetting the list to the validated list. I do see it happening that this does not get checked properly some day and we have just `Entry.validate_categories(category_ids, assignment)` someehwere in the codebase

> Hides explicit validation steps (making it easy to forget).

*ensures validation steps, not having to remember its needed

> Easily bypassed (use anything sql related, e.g. update, bulk_create etc)
> What do you do when you use a route which does not make use of the model save? You still need to validate, do you duplicate the validation logic?

Could all call the same function

mjvkeulen: > Mostly for other times when using this function it would not feel natural of resetting the list to the validated list. I do see it happening that this does not get checked properly some day and we have just Entry.validate_categories(category_ids, assignment) someehwere in the codebase

The only difference between using the return value and not is that the input is already cast to a set. Validation works the same.

> Could all call the same function

Can't reuse save for routes which can't call save. You still need a separate validation function, which is then called in save but also elsewhere?

src/django/VLE/views/teacher_entry.py
Larspolo: I ment more, when (bulk) creating entries, automatically link the categories. Reduces stuff we might forget when bulk creating entries elsewhere. But bhoth options work, not necessary

src/django/VLE/views/teacher_entry.py
Larspolo: >The only difference between using the return value and not is that the input is already cast to a set.

Only saying that that the function name, nor the docstring reflects it currently. `get_and_validate_categories`?

> You still need a separate validation function, which is then called in save but also elsewhere?

Yeah, sth like that I was thinking

Larspolo:
image
selecting templates does not leave a lot of space to scroll through

image
Default colours are not fully random, would like to see it be one of our theme colours by default, and user can change it to what they like

image
Light colours should (automatically set / be able to) get a black text colour, now some are not readable

image
This gray on this green is is a bit uneasy to look at

image
Would we like the pills to also be included in the timeline for easy visibility?

> .. want to do something with categories? Go to manage categories.

Want to add a category to the template you are working on? First you gonna finalize what you were working on, save it, add category in manage category, and continue where you were with editing the templates. I do reallyb think teachers would like to go like this: go to timeline, add node, add new template to node, add fields to template, add categories to template, repeat from step 1. Now this is not possible. Do you really not see that that might be a flow? Or is it more the technical difficulties?

src/django/VLE/models.py
```diff
@@ -2435,11 +2503,69 @@ def to_string(self, user=None):
...
+    def unused(self):
+        return self.filter(
+            presetnode__isnull=True,
+            entry__isnull=True,
+        ).distinct()
+
+    def full_chain(self, templates):
+        if isinstance(templates, list) or isinstance(templates, QuerySet) or isinstance(templates, set):
+            return self.filter(chain__template__in=templates)
+        return self.filter(chain__template=templates)
```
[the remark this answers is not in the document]
mjvkeulen: Why?

src/django/VLE/models.py
```diff
@@ -2456,6 +2582,24 @@ class Template(CreateUpdateModel):
         default=False
     )
 
+    fixed_categories = models.BooleanField(
```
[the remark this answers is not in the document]
mjvkeulen: That the categories associated with the template are fixed. A student cannot change the categories of an entry created using said template.

mjvkeulen:
You raised soms valid points which I will adress later. 

For future reviews, could you split each remark and where possible tag the relevant code? This allows for a more straightforward discussion on each topic in turn, with the ability to resolve individual topics.

mjvkeulen:
> Light colors should (automatically set/be able to) get a black text color, now some are not readable

I consider this a nice to have. 
Aim for bright colors by default. If a user chooses a light color themselves, they can just change it once they realize readability is an issue (which should be obvious from the start).

mjvkeulen:
> Default colours are not fully random

Could you more specifically point out which generation method is faulty?

> would like to see it be one of our theme colours by default, 

If you prefer a rotation of static colors, please provide a relatively large list (20?) of colors you approve of.

> and user can change it to what they like

This is already in place.

mjvkeulen:
> selecting templates does not leave a lot of space to scroll through

So multi-select expand not auto-scrolling the current view downwards. Not sure if this is desired? Also not introduced in this PR at least.

mjvkeulen:
> Would we like the pills to also be included in the timeline for easy visibility?

How should we handle overflow?

We could also use the color codes to create a new pill consisting of the combined categories. E.g. (color1 | color2| ....)

mjvkeulen:
> Want to add a category to the template you are working on? First you gonna finalize what you were working on, save it, add category in manage category, and continue where you were with editing the templates. I do reallyb think teachers would like to go like this: go to timeline, add node, add new template to node, add fields to template, add categories to template, repeat from step 1. Now this is not possible. Do you really not see that that might be a flow? Or is it more the technical difficulties?

I am sure it could be a flow for some, but I doubt it will be the common approach. I foresee teachers configuring the workload first (as is done currently). And if they would like to introduce categories, they can do so once the workload is finalized. 

These are the benefits of the current approach:
- Centralizing category management logic. Which I believe brings clarity to the already perceived difficult format editor.
- Keeps category logic inside its own view (category), instead of having to tightly couple all format logic in one view method. To me this is an important point, because it makes the code less complex / more straightforward to test / maintain / etc.

This is not to say your example is not valid, I simply believe its advantages do not outweigh the reasons I listed. 

Of course, we can realize both avenues (and obviously this has my preference), by moving template specific logic to its own view and also incorporating autosave. However, this would require a refactor which, timewise, might not be desired at this time.

mjvkeulen:
> Manage Categories > CategoryEdit keeps data the same even when switching assignments

Could not reproduce?

mjvkeulen:
> "Add Category" button is not visible till the catogory is already added? It behaves a bit weird for me

As far as I can tell this is also behaving as intended?

Add button is visible until you click add new category
Once you have made no further changes for longer than 3 seconds, the category is auto-saved. As a user you transition automatically to edit mode, this happens without disrupting the user, I assume this where your confusion stems from? 

Once saved, the add new category button becomes visible again.

mjvkeulen:
> Its possible to have unnames categories in the manage category list untill you reload, and then they dissapear.

How did you produce those? I assume this was via some abuse? 
The categories do not save (and cannot be saved with) an empty name. The user is also warned of this:

image

If the user does not correct this and reloads the page, the names will simply be unchanged and be restored to a new or cached fetch?

mjvkeulen:
> Its possible to have category names with just whitespace, will show up as empty. Also leads to weird behaviour when selecting only whitespace tempalte in filter

Fair point, category names are not intended to be empty. I believe besides the default space, there exist 15+ Unicode space variations. 
Would you not agree that with the non-empty guard in place, we can leave it up to the educator to provide meaningful category names?

Larspolo:
>  Aim for bright colors by default. If a user chooses a light color themselves, they can just change it once they realize readability is an issue (which should be obvious from the start).

If you want to aim for bright colours, the default should be black text right?

Larspolo:
> Could you more specifically point out which generation method is faulty?

Ment that they are random, would like to be system themed colours by default

> would like to see it be one of our theme colours by default, privode 20ish

Sth for @engelhamer ?

I would suggest darken / lighten everry existing colour by some value

mjvkeulen:
> If you want to aim for bright colours, the default should be black text right?

Sorry, I meant vibrant. So not very bright indeed.

image

When looking at colors using the HSL definitions, brightness should be capped at about 70%?

Larspolo:
> > selecting templates does not leave a lot of space to scroll through
> 
> So multi-select expand not auto-scrolling the current view downwards. Not sure if this is desired? Also not introduced in this PR at least.

Other mutliselects are often not at the bottom, moving it higher up would also solve the solution

Larspolo:
> How should we handle overflow?
> 
> We could also use the color codes to create a new pill consisting of the combined categories. E.g. (color1 | color2| ....)

Maybe with three dots? Also sth nice for @engelhamer

Larspolo:
> Could not reproduce?

Same now, idk anymore how I got this.

Although two other issues:
- "TypeError: JSON.parse(...).format is undefined" when visiting http://localhost:8080/Home/Course/1697/Assignment/1/Format
- "NS_ERROR_UNEXPECTED" with no more rich text box when (accidentally) dubbleclicking the edit button and then pressing it another time

mjvkeulen:
> After filter and clicking node in the timeline, it results in the wrong selected node

Yes so currently this is broken :(. I will look into this, but this will take some time.

mjvkeulen:
> Matplotlib also swifts between 10 colours

I would prefer color codes to remain unique, this is why I went with color generation.

Larspolo:
> I would prefer color codes to remain unique, this is why I went with color generation.

Thats when you shift to using your own colours, but now it will look stylish off the shelf

mjvkeulen:
> Current filter functions as a union, should we create an issue to filter as intersection?

We could make this a toggle, e.g. match `all` vs `any`. I do think any is generally preferred.

Larspolo:
image

Quickly navigating elsewhere with unsaves changes in the category editor results in the following toast

Larspolo:
> How did you produce those? I assume this was via some abuse?

Categories with a spacebar will show up unnames

> If the user does not correct this and reloads the page, the names will simply be unchanged and be restored to a new or cached fetch?

Even after reload its persistent

mjvkeulen:
> Quickly navigating elsewhere with unsaves changes in the category editor results in the following toast

Should be fixed.

Larspolo:
> As far as I can tell this is also behaving as intended?
> 
> Add button is visible until you click add new category
> Once you have made no further changes for longer than 3 seconds, the category is auto-saved. As a user you transition automatically to edit mode, this happens without disrupting the user, I assume this where your confusion stems from?
> 
> Once saved, the add new category button becomes visible again.

May be intended, but it doesn't feel natural to me to use. Maybe @engelhamer can also way in on this. It may also be together with the other issues, and that it has a three second delay

mjvkeulen:
>> How did you produce those? I assume this was via some abuse?

> Categories with a spacebar will show up unnames

>> If the user does not correct this and reloads the page, the names will simply be unchanged and be restored to a new or cached fetch?

> Even after reload its persistent

So this in regards to my other invisible character-related comments. I believe we can leave this at the discretion of the educator.

mjvkeulen:
> May be intended, but it doesn't feel natural to me to use.

Could you make natural explicit?

Larspolo:
> > > How did you produce those? I assume this was via some abuse?
> 
> > Categories with a spacebar will show up unnames
> 
> > > If the user does not correct this and reloads the page, the names will simply be unchanged and be restored to a new or cached fetch?
> 
> > Even after reload its persistent
> 
> So this in regards to my other invisible character-related comments. I believe we can leave this at the discretion of the educator.

Nah, a `.strip()` before saving would solve it, why dont at that point?

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
+<template>
```
Larspolo, line 1: > Could you make natural explicit?

Let's switch. Just putting everything on a new line

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+    <div>
```
Larspolo, line 2: Visual response to adding a category is minimal.

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+        <b-form-group :invalid-feedback="nameInvalidFeedback">
```
Larspolo, line 3: Elsewhere the "add X" button already adds something to the section it is in, pressing ti twice, adds two empty things to the section. These are always visible, here it is not.

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+            <b-form-input
```
Larspolo, line 4: Having to wait three seconds before something happens when adding a category

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+                v-model="data.name"
```
Larspolo, line 5: After adding you are immediately going to the editing mode, without any visual change that this has happened

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+                :state="nameInputState"
```
Larspolo, line 6: Then the obvious, this is now automatic sync while the whole entire website is still manual sync

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+                autofocus
```
Larspolo, line 7: No real save button anywhere, nor sth that indicates that changes have been saved

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+                placeholder="Name"
```
Larspolo, line 8: Also other issues I brought make it less clean: "nameless" categories, white text on light backgrounds, template selection cutting off.

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,136 @@
...
+                class="theme-input"
```
Larspolo, line 9: Other "cannot reproduce" bugs sometimes occurring which makes it just feel extra buggy overall.

I sometimes found the edit icon not working anymore, or with a real long delay

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Press the add button -> new menu appears? I would argue that's quite the visual response.

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Is that desired behavior?

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Before what happens? What is the issue you are describing here?

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: There should be no user disruption going from save -> edit mode, so that is working as intended.

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Mostly manual sync*. But yes, I far prefer this.

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: UI updates as you make changes. 

Prob best UX wise, no need to manually save, WYSIWG

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: - Nameless is not possible. A user can enter invisible characters such a space (they just shouldn't and thats fine by me).
- Light backgrounds should not be generated, help me find the issue if so.
- Template selection cutting off? There is no cut off issue afaik. Do you mean the multi-select not auto forcing a downward scroll event?

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: This is too vague, what bugs? 

Edit icon not working?

src/vue/src/components/category/CategoryEdit.vue
Larspolo: and add button dissapears, typing in the name, add button reappears, clicking it, removes all previous input (visual wise)

src/vue/src/components/category/CategoryEdit.vue
Larspolo: I like consistency, having everything work differently is worse

src/vue/src/components/category/CategoryEdit.vue
Larspolo: Before the name shows up in the category list

src/vue/src/components/category/CategoryEdit.vue
Larspolo: But on the rest of the platform there is

src/vue/src/components/category/CategoryEdit.vue
Larspolo: We already noticed teachers finding it unnatural clicking away the template editor without a save button

src/vue/src/components/category/CategoryEdit.vue
Larspolo: >Nameless is not possible. A user can enter invisible characters such a space (they just shouldn't and thats fine by me). 

It is possible, entering a left over spacebar should be expected user error, and handled accordingly, like we also do elsewhere in the system: uploading bonus points csv, creating new users in admin panel, assignment / course / node names & description

> Light backgrounds should not be generated, help me find the issue if so.
image


> Template selection cutting off? There is no cut off issue afaik. Do you mean the multi-select not auto forcing a downward scroll event?

Yeah that one

image

src/vue/src/components/category/CategoryEdit.vue
Larspolo: image
Select icon from the right.

Sometimes just doesnt work, dunno why

src/vue/src/components/category/CategoryEdit.vue
Larspolo: I would too, if its consistent with the rest, or at least visually indicative of changes being actually saved

Larspolo: Think it might also the tthe way that it also is not fully fluent. Like:
- Presenting a big toast in the screen when a name is the same as another already present name
- Only seeing it appear in the categories above after a couple seconds, creates the "what did I do now" feeling

src/vue/src/components/format/ManageAssignmentCategories.vue
```diff
@@ -0,0 +1,229 @@
+<template>
```
Larspolo, line 1: This model opens multiple times (on Firefox) for each time you clicked on a pill to see the details

src/django/VLE/models.py
```diff
@@ -2768,3 +2921,121 @@ def validate_jir_before_save(sender, instance, **kwargs):
...
+        if category:
+            equal_name = equal_name.exclude(pk=category.pk)
+            equal_color = equal_color.exclude(pk=category.pk)
+
+        if name == '':
+            raise ValidationError('Please provide a non empty name.')
+        if equal_name.exists():
+            raise ValidationError('Please provide a unqiue category name.')
+        if equal_color.exists():
+            raise ValidationError('Please provide a unqiue category color.')
```
Larspolo, line 3003: Imo can category colours be the same

src/django/VLE/utils/template.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def handle_template_update(data, format, new_ids):
+    current_template = Template.objects.get(pk=data['id'])
+
+    if _should_be_archived(current_template, data):
+        current_template.archived = True
+        current_template.save()
+
+        new_template = Template.objects.create_template_and_fields_from_data(data, format, current_template)
```
Larspolo, line 59: This should be indented one to the left right? If so, would suggest expanding the test to test if template is also created if current template should be deleted in stead of atrchived.

src/django/VLE/views/category.py
```diff
@@ -0,0 +1,144 @@
...
+        entry = Entry.objects.select_related('node__journal__assignment', 'template').get(pk=entry_id)
+        template = entry.template
+        assignment = entry.node.journal.assignment
+
+        if request.user.has_permission('can_grade', assignment):
+            pass  # A category can always be changed by someone with the permission to grade.
+        elif request.user.has_permission('can_have_journal', assignment):
+            if template.fixed_categories:
+                return response.forbidden('This entry\'s categories are locked.')
+            if not request.user.can_edit(entry):
```
Larspolo, line 130: `can_edit` is not a function, its `VLE.permissions.can_edit(user, entry)` made me realize how inconsistent the permission system is in some places. Also maybe good to add the `can_edit` for categories, so this can be done in 1 function

src/django/VLE/views/category.py
```diff
@@ -0,0 +1,144 @@
...
+                return response.forbidden('You can no longer edit the entry\'s categories.')
+        else:
+            return response.forbidden('You are not allowed to edit the entry\'s categories.')
+
+        if add:
+            if not Category.objects.filter(assignment=assignment, pk=category.pk).exists():
+                raise ValidationError('Categories can only be linked to entries which are part of the same assignment.')
+
+        if add:
```
Larspolo, line 139: Same check twice in a row, can simply be removed

src/django/VLE/models.py
mjvkeulen: Could you also argue why?

Against: Name and color are used as identifiers for the category, this is why these markers should be unique.

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: This is not something I can reproduce?

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: > Light backgrounds should not be generated, help me find the issue if so.

Are these generated via preset db or frontend on creation? Do you see issues with the algo's used?

mjvkeulen: >> Template selection cutting off? There is no cut off issue afaik. Do you mean the multi-select not auto forcing a downward scroll event?

> Yeah that one

Would this fix it? https://github.com/shentao/vue-multiselect/pull/1350

mjvkeulen: >> Nah, a .strip() before saving would solve it, why dont at that point? Accidental spacebars happen more often then expected

> Nameless is not possible. A user can enter invisible characters such a space (they just shouldn't and thats fine by me).

Alright, we'll strip for spacebars specifically, this will not solve the other characters, however.

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: > Presenting a big toast in the screen when a name is the same as another already present name

Now also validates before posting creation (not just edit)

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: > Are these generated via preset db or frontend on creation? Do you see issues with the algo's used?

https://github.com/eJourn-al/eJournal/pull/1067/commits/4d019bdeed2cc860dd9fa2232ef2e88bab49f642

I capped the HSL lightness at 80%, I think this should fix the issue?

src/vue/src/components/format/ManageAssignmentCategories.vue
mjvkeulen: Good spot, should be fixed here: https://github.com/eJourn-al/eJournal/pull/1067/commits/24e62cfad3f6ffe267647945d36a4fcaa92ecba3

src/django/VLE/utils/template.py
mjvkeulen: A new template should only be created if meaningful changes are made (detected in _should_be_archived). The code here is correct afaik.

What is throwing you off in the code or what is incorrect?

src/django/VLE/views/category.py
mjvkeulen: > can_edit is not a function, its VLE.permissions.can_edit(user, entry) made me realize how inconsistent the permission system is in some places.

As in, all can_edit logic should be moved to the permissions module for consistency? I feel if the logic does not transcend a specific view its fine to leave it in the view itself. 
Also, should I change something in the code here? Or was this just an epiphany?

> Also maybe good to add the can_edit for categories, so this can be done in 1 function

A category instance is not aware to which Entry (specifically) it belongs. I don't think what you have in mind will work.

src/django/VLE/views/category.py
mjvkeulen: Currently, it is a block of validation, followed by a block that changes state (after all validation passes). 
Is it not worth keeping the check to keep this split explicit?

src/django/VLE/models.py
Larspolo: For: Giving more freedom to the user. Let then decide how to use it. Maybe they want all alike categories to have the same colour

Difference with colour is that name should be enforced differently as two with the same name would work in multiselect dropdown menu's, or other text-only visualizations.

src/django/VLE/utils/template.py
Larspolo: Maybe change the name to `_should_be_updated` as it now not only should be archived, but also updated. Furthermore, some shouldn't be archived, but deleted instead, this is mostly what confused me

src/django/VLE/views/category.py
Larspolo: I think the code should be changed. I didnt check if its running or not, but there is no `user.can_edit` function in models. See https://github.com/eJourn-al/eJournal/blob/develop/src/django/VLE/models.py#L382. You sure this works?

src/django/VLE/views/category.py
Larspolo: Sure, both work

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -128,16 +79,141 @@
...
+        findNodeIndex (nodes, node) {
+            /* When editing the timeline (edit = true) we are working with preset nodes directly
+             * Otherwise we are working with a timeline consisting of normal nodes */
+            const idKey = (this.edit) ? 'id' : 'nID'
+
+            return nodes.findIndex(elem => elem[idKey] === node[idKey])
+        },
+        /* TODO: fix node selection, mappedSelect should be updated to reflect its new position.
+         * If the currently selected node no longer occurs, emit select to the assignment start node */
+        filterByCategory (filters) {
```
mjvkeulen, line 183: >> After filter and clicking node in the timeline, it results in the wrong selected node

> Yes so currently this is broken :(. I will look into this, but this will take some time.

So this is fixed now, see the javascript in this component. Most of the changes are part of this commit: https://github.com/eJourn-al/eJournal/pull/1067/commits/169aa9182afe899cfc500a0ffff30eb9263fc5d2

Working on that commit I truly highlighted for me the need for a refactor of the TimeLine and parent components regarding state management.

src/django/VLE/views/category.py
mjvkeulen: The user model contains both a `can_edit` and a `check_can_edit` method?

src/django/VLE/migrations/0071_category.py
```diff
@@ -0,0 +1,90 @@
+# Generated by Django 2.2.10 on 2020-12-16 20:24
```
Larspolo, line 1: de migrations 70 bleef hingen omdat zogenaamd is_deadline niet bestond moeten we even onthouden

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -128,16 +79,141 @@
...
+        hasCategory (categories, filters) {
+            return categories.some(category => filters.find(filteredCategory => category.id === filteredCategory.id))
+        },
+        findNodeIndex (nodes, node) {
+            /* When editing the timeline (edit = true) we are working with preset nodes directly
+             * Otherwise we are working with a timeline consisting of normal nodes */
+            const idKey = (this.edit) ? 'id' : 'nID'
+
+            return nodes.findIndex(elem => elem[idKey] === node[idKey])
+        },
```
Larspolo, line 180: Filter knop werkte niet voordat Robert de pagina herladen was.

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -128,16 +79,141 @@
...
+        /* Next to filtering the nodes based on the selected categories, also keeps the selected node index in sync.
```
Larspolo, line 181: Progress goals, zolang die niet ook categorieën kunnen krijgen, wil je er altijd in willen hebben, ook als de filter aanstaat

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -128,16 +79,141 @@
...
+        },
+        hasCategory (categories, filters) {
+            return categories.some(category => filters.find(filteredCategory => category.id === filteredCategory.id))
+        },
+        findNodeIndex (nodes, node) {
+            /* When editing the timeline (edit = true) we are working with preset nodes directly
+             * Otherwise we are working with a timeline consisting of normal nodes */
+            const idKey = (this.edit) ? 'id' : 'nID'
+
+            return nodes.findIndex(elem => elem[idKey] === node[idKey])
```
Larspolo, line 179: De “add knop” lijkt teveel op een category, hierdoor lijkt het te veel op een categorie, en dacht Robert gelijk dat het de 4e categorie is. Het gaat soepel. Als hij het doorleest kan het wel gedaan worden.

src/vue/src/components/category/EntryCategories.vue
```diff
@@ -0,0 +1,129 @@
...
+        </h2>
+
+        <category-display
+            v-if="'categories' in entry"
+            :id="`${id}-display`"
+            :editable="editable"
+            :categories="entry.categories"
+            @remove-category="removeCategory($event)"
+        >
+            <b-badge
```
mjvkeulen, line 17: This button might not be sufficiently distinct from the categories themselves.

src/vue/src/components/category/CategoryDisplay.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+    <div>
+        <b-badge
```
mjvkeulen, line 3: It might not be sufficiently clear that this category display also functions as the button to trigger more category information.

src/vue/src/components/category/CategoryDisplay.vue
```diff
@@ -0,0 +1,73 @@
...
+                v-if="editable"
+                class="fill-red ml-1"
+                name="times"
+                @click.stop.native="$emit('remove-category', category)"
+            />
+        </b-badge>
+
+        <slot/>
+
+        <b-modal
```
mjvkeulen, line 23: The category description is not visible during the add category to entry flow

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -0,0 +1,137 @@
...
+                type="text"
+                trim
+            />
+        </b-form-group>
+
+        <text-editor
```
mjvkeulen, line 15: Use headers to input fields

src/vue/src/components/format/ManageAssignmentCategories.vue
```diff
@@ -0,0 +1,237 @@
...
+    <div
```
Larspolo, line 2: Optional: a visual indication to satisfy the user that its actually being saved. A toast is probably too much as it is constantly saved

src/vue/src/components/format/ManageAssignmentCategories.vue
```diff
@@ -0,0 +1,237 @@
...
+        v-intro="'Categories can be linked to student entries and be used to filter the timeline.'"
```
Larspolo, line 3: Currently students cannot see the description of a category before adding it to the entry. This could be fixed by expanding the model with also showing the description before adding. However, this reduced the "quick add" ability even further. It would be nice to allow for both quick add, and insightfull add options when adding categories.

Possibility is adding a button in the quick dropdown to expand to a larger model view with then the descriptions also viewable

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Fixed in 96a49184fe32dc5814c2dd8bf740292d486cef5d

src/vue/src/components/timeline/Timeline.vue
mjvkeulen: Fixed in d1190b30146f9ca0d01a51de6d43e789190bbd95

src/vue/src/components/timeline/Timeline.vue
mjvkeulen: Fixed in a336cfbe56eb8e9ca12441946b8dcb85021a95a1

src/django/VLE/utils/template.py
```diff
@@ -0,0 +1,96 @@
...
+    check_fields = [
+        'name',
+        'preset_only',
+        'fixed_categories',
+    ]
+
+    return any(getattr(template, f) != data[f] for f in check_fields)
+
+
+def _field_concrete_fields_updated(field, data):
```
engelhamer, line 14: Pretty confusing names.

src/django/VLE/utils/template.py
engelhamer: Template should only be archived when it is used right? If not then there's no need to keep the version as it will never be used anywhere / cannot be reverted to either way.

src/django/VLE/views/assignment.py
```diff
@@ -447,7 +447,8 @@ def copy(self, request, pk):
         """
         course_id, months_offset = utils.required_typed_params(request.data, (int, 'course_id'), (int, 'months_offset'))
         course = Course.objects.get(pk=course_id)
-        assignment_source = Assignment.objects.get(pk=pk)
+        assignment_source_id = pk
```
engelhamer, line 450: Arbitrary rename?

src/django/VLE/views/category.py
```diff
@@ -0,0 +1,144 @@
...
+from rest_framework.decorators import action
+
+import VLE.utils.generic_utils as utils
+import VLE.utils.responses as response
+from VLE.models import Assignment, Category, Entry, Template
+from VLE.serializers import CategoryConcreteFieldsSerializer, CategorySerializer
+from VLE.utils import file_handling
+
+
+def _validate_templates(templates, assignment):
```
engelhamer, line 13: I would like to see it the other way around: add a category to a template rather than add a template to a category. Similar to how GitHub tags work. This would be quicker and more intuitive to find / with fewer clicks per template.

src/django/VLE/views/teacher_entry.py
```diff
@@ -8,12 +8,34 @@
...
+
+    Does not modifying the entry category links of entries which already belong to the desired categories.
+
+    Args:
+        entries (:model:`VLE.entry`): List of entries associated with a teacher entry which should be updated.
+        author (:model:`VLE.author`): Author of the teacher entry edit.
+        new_category_ids ([int]): List of category ids which should now belong to the entry.
+        existing_category_ids ([int]): List of category ids which currently belong to the entry.
+    """
+    EntryCategoryLink.objects.filter(entry__in=entries).exclude(category_id__in=new_category_ids).delete()
```
engelhamer, line 29: What if a student added a category themselves? I could think of a scenario where e.g. students have to mark 5 elements in their journal with a certain category for e.g. 'Final Assessment'. In this case the categories added by the student should not be removed, but they are by this line.

Indeed it would probably make more sense to remove only the removed diffs between `existing_category_ids` (not currently used below, but is a param?) and the `new_category_ids`.

src/django/VLE/views/teacher_entry.py
```diff
@@ -8,12 +8,34 @@
...
 from VLE.serializers import TeacherEntrySerializer
 from VLE.utils import entry_utils, grading
 from VLE.utils.error_handling import VLEBadRequest
 
 
+def _update_categories_of_existing_entries(entries, author, new_category_ids, existing_category_ids):
+    """
+    Keeps the categories of all existing entries synced with those set on the teacher entry.
+
+    Does not modifying the entry category links of entries which already belong to the desired categories.
```
engelhamer, line 21: Modify? Kind of confusing sentence.

engelhamer (commented):
25/72 viewed. About the link tables: would they not be suitable `ArrayFields`?

src/django/VLE/migrations/0072_template_chain.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0071_category'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='TemplateChain',
```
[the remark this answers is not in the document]
mjvkeulen: What would be the advantage? Did you think about how you would query such a structure?

src/django/VLE/models.py
```diff
@@ -2768,3 +2921,121 @@ def validate_jir_before_save(sender, instance, **kwargs):
...
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+    )
+    assignment = models.ForeignKey(
+        'assignment',
+        related_name='categories',
+        on_delete=models.CASCADE,
+    )
+    templates = models.ManyToManyField(
```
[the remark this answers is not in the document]
mjvkeulen: What comparison are you making?

src/django/VLE/serializers.py
```diff
@@ -120,17 +120,65 @@ def get_download_url(self, file):
...
+            'name',
+            'format',
+            'preset_only',
+            'archived',
+            'fixed_categories',
+        )
+        read_only_fields = fields
+
+
+class CategoryConcreteFieldsSerializer(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
mjvkeulen: Django native term

src/django/VLE/serializers.py
```diff
@@ -379,12 +513,27 @@ class Meta:
 
     prefetch_related = [
         'courses',
+        'categories__templates',
```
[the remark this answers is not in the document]
mjvkeulen: I don't follow? Each category has its own set of templates.

src/django/VLE/utils/template.py
mjvkeulen: Why?

`concrete_fields` is a term you are simply not familiar with yet. Field is our model.

src/django/VLE/views/assignment.py
mjvkeulen: What are you referring to, what makes it arbitrary?

src/django/VLE/views/teacher_entry.py
mjvkeulen: This is teacher entry specific?

mjvkeulen: If your comment still holds, we had a discussion on this earlier in the PR if I recall, do you agree with the arguments there?

src/django/VLE/migrations/0072_template_chain.py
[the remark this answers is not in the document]
engelhamer: See the previous comment for the advantages? And querying would require a loop as far as I can tell now indeed. However that is only needed when updating categories.

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: There is a manytomanyfield in this model for templates, but not for entries. Both can have categories.

src/django/VLE/serializers.py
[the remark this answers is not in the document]
engelhamer: I find its semantic meaning confusing and not apparent. Can you add a comment on this?

src/django/VLE/serializers.py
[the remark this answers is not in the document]
engelhamer: Sorry, I thought this was an odd query for obtaining the format templates (but we already have format above so hence the confusion).

src/django/VLE/utils/template.py
engelhamer: Would you know that this function does straight away when reading this somewhere?

src/django/VLE/views/assignment.py
engelhamer: We use pk in other places, you rename it here (for readability?) making this change arbitrary.

src/django/VLE/views/teacher_entry.py
engelhamer: Yes it is specific to teacher entries, but you are now adding stuff to teacher entries and the file I'm commebting on is specific to teacher entries too. I don't get your response?

src/django/VLE/views/assignment.py
mjvkeulen: I see your point regarding using pk in other locations. Most other locations don't have to work with a source and target model instance, however. 

In my opinion, the rename helps with clarifying to which specific model instance the pk belongs.

src/django/VLE/serializers.py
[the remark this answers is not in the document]
mjvkeulen: This is already in use here and there, e.g. see the function `instance_conrete_fields_dict`. I am not sure where to place the explanation of the term `concrete_fields`.

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: I am still not sure what you mean. Are you referring to that the many to many relationship is only defined on one side of the models? E.g. assignment defines a m2m to course on the model but not vice versa despite the relationship existing? And your question is simply regarding the placement of this relationship definition in the models themselves?

src/django/VLE/utils/template.py
mjvkeulen: I think the readbility is excellent:

image

mjvkeulen: The pattern is simply <model>_<concrete_fields>_updated. The clash here is only that the model is name `Field`, I don't think straying from the consistent pattern would help with readability?

src/django/VLE/views/teacher_entry.py
mjvkeulen: > Yes it is specific to teacher entries

Great, mb.

We had a discussion on this topic earlier in the PR if I recall, do you agree with the arguments there?

src/django/VLE/migrations/0072_template_chain.py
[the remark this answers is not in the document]
mjvkeulen: Your comment implicates a rewrite of the working structure I put forth. Of which we have established that querying is faster.
It is also generally accepted that querying a nonrecursive data structure is more straight forward in SQL.

> I think that will be easier to maintain and read

What makes this more difficult to maintain or read? Could you elaborate? The addition of an intermediate model does in itself not make this argument true?

src/django/VLE/views/category.py
mjvkeulen: I agree being able to configure the m2m relationship between categories and templates from both sides of the relationship would be worthwhile.

E.g. when updating or introducing categories to an assignment one would be best helped by being able to create some categories and quickly setting which templates should be associated.
Whereas when creating a new assignment it is likely someone would like to be able to add a preset to the timeline, create a template for it, and add categories in one flow. (This is a TODO in the description).

I also think that the format edit is already considered complex (or just confusing) for some educators. This is something we need to be careful of when introducing more logic to the format editor.

Would you agree that generally a category will be linked to multiple templates. And a template only to a few categories?

src/django/VLE/models.py
mjvkeulen: Keeping colors unique also opens up routes to use them as unique identifiers when there is limited space. E.g. in the timeline.

src/django/VLE/utils/template.py
mjvkeulen: > Template should only be archived when it is used right? If not then there's no need to keep the version as it will never be used anywhere / cannot be reverted to either way.

I don't see where the code differs, could you elaborate?

src/django/VLE/models.py
Larspolo: But leaving it random is not necessarily unique, noone sees the difference between `ffffe` and `ffffff`

src/django/VLE/views/assignment.py
Larspolo: You can also change the function parameter name then right? Or does that create other issues?

src/django/VLE/utils/template.py
Larspolo: `_field_has_concrete_fields_updated` maybe?

src/django/VLE/views/assignment.py
mjvkeulen: Would be fine by me, but as I understand it, that's not what the remark is about.

src/django/VLE/migrations/0072_template_chain.py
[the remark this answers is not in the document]
engelhamer: We want to keep our database as simple as possible. Take for example the `Format` model: we want to remove it in a future version. Adding a new model for something like this does no align with the other proposed DB changes. The quicker query is the sole benefit of this added model. I agree that it works now and is not worth rewriting, but wanted to know why you decided on adding a new model. No further discussion needed.

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: For the end user it is possible to add a category to:
- Templates
- Entries
The many to many relationship is defined in the `Category` model for:
- Templates
Why the above discrepancy?

src/django/VLE/serializers.py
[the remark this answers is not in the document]
engelhamer: I don't see the term frequently being used (at least not explicitly) in Django docs, so would be nice if you have some source where I can read to understand what you mean. As a comment here is also fine if you don't know where to put it.

src/django/VLE/utils/template.py
engelhamer: It is confusing because the same word is used twice, with different meanings, without even a verb in between uses.

src/django/VLE/views/category.py
engelhamer: > Would you agree that generally a category will be linked to multiple templates. And a template only to a few categories?

Yes, so the user will likely spend more time in the templates themselves and link it quickly. From the 1 user test we've performed this problem immediately became apparent so I'd like to have the option to add a category when editing templates. It can remain hidden as long as no categories exist, to keep the same simplicity for users that do not want to use categories at all.

src/django/test/test_teacher_entry.py
engelhamer: I think blocking the ability to change categories is odd: if a teacher deliberately allows changing of content by not grading right away, students should also be able to change categories. Even more: it could be that teachers want students to 'categorise', say, their best 5 entries for final grading.

src/django/VLE/models.py
engelhamer: When colors can be picked by the user they could use the color as an overarching grouping of categories. They could still make them unique enough (and probably will, if there are enough colours) for timeline stuff.

src/django/VLE/views/assignment.py
engelhamer: This is just a consistency thing and does not need this much discussion going on.

src/django/VLE/serializers.py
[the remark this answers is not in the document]
mjvkeulen: Fields that have a column on the direct table, the importance is that the data does not require a join or further traversal of a relation.

src/django/test/test_teacher_entry.py
mjvkeulen: > I think blocking the ability to change categories is odd: if a teacher deliberately allows changing of content by not grading right away, students should also be able to change categories.

This is the current situation.

The discussion is regarding what should happen when a teacher edits an existing teacher entry. What should happen to any category edits made by the students in the meanwhile?

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: > For the end user it is possible to add a category to:
> - Templates

It is the other way round right?

A user can configure the default templates for a category, this is why I opted to describe this relation in the category model itself.

src/django/VLE/views/category.py
mjvkeulen: >> Would you agree that generally a category will be linked to multiple templates. And a template only to a few categories?

> Yes, so the user will likely spend more time in the templates themselves and link it quickly.

I am not sure I follow? The point I made here is that linking a category to multiple templates is done more quickly via the category itself. Instead of navigating to each template edit menu and for each template open the category selection menu and select the category, it becomes for the one category use the multi-select to configure all of its default templates in one go.

What makes more sense and is faster depends on at what stage the user introduces categories and how they reason about the relationship.

I believe the current solution has its advantages and is something a new user will get used to quickly (and the implementation is already finished). Even when you can manage the category association from the template side as well, you will still have to navigate to manage categories to create a category to begin with.

I am fine with adding the category flow to the template side as well, since this is what you both want, and like I said I can see the advantage depending on the flow of the user. I do note that cleanly integrating this will take more than just an hour or two, and we have a working solution already (IMO of course).

src/django/test/test_teacher_entry.py
engelhamer: I think they should remain, as stated in the other comment at the update function https://github.com/eJourn-al/eJournal/pull/1067#discussion_r553337317 . By using the diff between the old and new categories we could have students' changes persist.

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: I find `the default templates for a category` to be confusing: why would this be different from entries? Intuitively I would think that there is not any semantic difference between categorising entries or templates right: both concern 'adding' the template or entry to a category. Or phrased the other way around.

src/django/VLE/migrations/0071_category.py
engelhamer: Same for me on my local DB.

src/django/test/test_teacher_entry.py
mjvkeulen: This was the scenario I had implemented initially but this gives no way for a teacher to correct a category-related mistake apart from on a journal individual basis. 

Also just using the diff is not enough, a teacher can perform multiple edits in a row, which categories should persist?

src/vue/src/components/format/ManageAssignmentCategories.vue
mjvkeulen: Should be fixed in https://github.com/eJourn-al/eJournal/pull/1067/commits/9faf892ce7107045d200f6ae9502bd9c67c462d5

All category addition is done via multi-select, which can also display a category's description before selecting said category via an information icon.

src/vue/src/components/timeline/TimelineNodeInfo.vue
```diff
@@ -32,7 +37,12 @@
 </template>
 
 <script>
+import CategoryDisplay from '../category/CategoryDisplay.vue'
```
mjvkeulen, line 40: Best to stick to using the alias `@`

engelhamer: Ah didn't notice, this import was automatically added

src/vue/src/components/category/CategorySelect.vue
```diff
@@ -60,6 +61,12 @@
                     </template>
                 </div>
             </template>
+
+            <template
+                slot="maxElements"
+            >
+                No more categories.
```
mjvkeulen, line 68: [NIT]: No categories left

src/django/VLE/models.py
mjvkeulen: Category colors are no longer unique

src/django/VLE/views/teacher_entry.py
mjvkeulen: Resolving this as we have a discussion on the same topic ongoing higher up.

src/django/VLE/models.py
```diff
@@ -2459,6 +2909,76 @@ class Template(CreateUpdateModel):
...
+            if not category_ids:
+                return
+
+            if len(category_ids) != len(category_data):
+                raise ValidationError('Duplicate categories provided.')
+            if assignment.categories.filter(pk__in=category_ids).count() != len(category_ids):
+                raise ValidationError('One or more categories are not part of the assignment.')
+
+        # QUESTION: What else would we like to validate regarding the field set?
+        def validate_field_set():
```
mjvkeulen, line 2955: What else would we like to validate regarding the field set?

src/django/test/test_template.py
```diff
@@ -0,0 +1,579 @@
...
+            patch_data = TemplateSerializer(template).data
+            patch_data['pk'] = template.pk
+            patch_data['fixed_categories'] = True
+
+            resp = api.patch(self, 'templates', params=patch_data, user=self.assignment.author)
+
+            template.refresh_from_db()
+            assert template.archived, 'Changing the fixed categories flag archives the template.'
+            # QUESTION: Should fixed_categories changes be updated for the entire template chain or not?
+            assert not template.fixed_categories, 'Fixed categories is unchanged on the archived template.'
```
mjvkeulen, line 530: Should fixed_categories changes be updated for the entire template chain or not?

engelhamer:
All TODOs done.

Larspolo (approved):

engelhamer (approved):
Amazing

PR #1069, opened 2020-12-21
Fixed an issue where group names could get out of sync with DataNose

Larspolo:
## Description
Fixed issue where DN sync would not update existing group names

- Fixes Z#1446


engelhamer (approved):

mjvkeulen (approved):
Not sure how to test this manually

PR #1072, opened 2020-12-22
Add linter checks before commit

Larspolo:
## Description
Do we want this? Increases time before commit, but reduces the "fixed linter issues" commits


engelhamer (approved):
Fine by me

mjvkeulen (approved):

PR #1074, opened 2020-12-22
Improved speed of posting entries and comments

Larspolo:
## Description
Speedup the entry & comment creation by generating notifications in subtask

also 

Added failfast and improved linter checks

- Fixes Z#1422


src/django/VLE/tasks/notifications.py
```diff
@@ -23,3 +23,37 @@ def generate_new_node_notifications(node_ids):
                     user=author.user,
                     node=node,
                 )
+
+
+@shared_task
+def generate_new_comment_notifications(comment_id):
+    comment = VLE.models.Comment.objects.get(pk=comment_id)
```
mjvkeulen, line 30: Prob want to use select related on the author and journal and stuff to save some queries.

src/django/VLE/tasks/notifications.py
```diff
@@ -23,3 +23,37 @@ def generate_new_node_notifications(node_ids):
...
+            comment=comment,
+        )
+
+    # Generate notifications for all other students in journal
+    for author in comment.entry.node.journal.authors.all().exclude(user=comment.author):
+        VLE.models.Notification.objects.create(
+            type=VLE.models.Notification.NEW_COMMENT,
+            user=author.user,
+            comment=comment,
+        )
```
mjvkeulen, line 46: Bulk create? Or was the create function overloaded?

Larspolo: heavily overloaded

src/django/VLE/tasks/notifications.py
Larspolo: Its in a shared task, so it doesnt matter too muhc, but its a good habbit to have, so Ill update it

Larspolo: Nvm, only queries have select related, and other comment values get used in the notification creation

mjvkeulen: This is a query right?

Larspolo: `.get` returns the comment object, not a *queryset

mjvkeulen: But it's still a query and therefore possible. Just flip the order, Comment.objects.select_related('etc'').get(pk=comment_id)

Larspolo: Didnt know that, added it now

engelhamer (approved):

src/django/test/test_notification.py
```diff
@@ -157,14 +160,30 @@ def test_entry_notification(self):
...
+        node.save()
+        entry.refresh_from_db()
+        assert not hasattr(entry, 'node'), 'test setup should remove node from entry'
+        with QueryContext() as context:
+            generate_new_entry_notifications(entry.pk, node.pk)
+        node.entry = entry
+        node.save()
+        entry.refresh_from_db()
+        assert hasattr(entry, 'node'), 'test setup should add node to entry'
+        with self.assertNumQueries(len(context.captured_queries) - 2):
```
mjvkeulen, line 184: What does the -2 signify?

Larspolo: Put it in variable

src/django/test/test_notification.py
```diff
@@ -157,14 +160,31 @@ def test_entry_notification(self):
...
 
         self.check_send_push_notification(new_notifactions.get(type=Notification.NEW_ENTRY))
 
+        # Because of a limitation in the test factory, we also need to provide the node pk
+        # However, when entry already has a node, it should not be used to reduce the number of queries used
+        node = entry.node
+        node.entry = None
+        node.save()
+        entry.refresh_from_db()
+        assert not hasattr(entry, 'node'), 'test setup should remove node from entry'
```
mjvkeulen, line 177: Isn't this basically testing Django internals? (it increases the strain on reading the test to me)

src/django/test/test_notification.py
```diff
@@ -157,14 +160,31 @@ def test_entry_notification(self):
...
+        with QueryContext() as context:
+            generate_new_entry_notifications(entry.pk, node.pk)
+        node.entry = entry
```
mjvkeulen, line 180: Hspace as this is a new test block? Could even move your comment here -> # However, when entry already has a node, it should not be used to reduce the number of queries used

src/django/test/test_notification.py
```diff
@@ -157,14 +160,31 @@ def test_entry_notification(self):
...
+        node.save()
+        entry.refresh_from_db()
+        assert hasattr(entry, 'node'), 'test setup should add node to entry'
+        nr_queries_saved_select_related_and_node_fetching = 2
```
mjvkeulen, line 184: Select related does not increase the query count.

Larspolo: is "saves" queries

mjvkeulen (approved):

PR #1075, opened 2020-12-23
Fixed various minor issues

mjvkeulen:
- Fixes Z#1383
- Fixes Z#1387
- Fixes Z#1389
- Fixes Z#1403
- Fixes Z#1414
- Fixes Z#1418
- Fixes Z#1095


Larspolo:
> Submodule commands updated from f2f2b8 to f7d314

Should not be included in this PR

Larspolo:
Good PR, all works well. Just some other issues that were inline with this PR:

- JIR modal also does not display the assignment due dates, idk if that is meant

- Would be nice if other stuff also (slowly) made use of the new validation input (like the points possible)

- Teacher specific notification settings are still visible for students (same other way around teachers having student notification settings)

Larspolo (approved):

src/django/VLE/validators.py
```diff
@@ -50,7 +50,13 @@ def validate_password(password):
         raise ValidationError("Password needs to contain a special character.")
 
 
-def validate_entry_content(content, field):
+def validate_youtube_video_url(url):
+    r = r"^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w\-]+\?v=|embed\/|v\/)?)([\w\-]+)(\S+)?$"
```
engelhamer, line 54: This regex yields different results from the front end (js) one. For example `https://www.youtube.com/channel/UCDZkgJZDyUnqwB070OyP72g` is valid from this check, but invalid front end. The latter makes sense, since this is a channel URL rather than a video.

src/django/test/test_utils.py
```diff
@@ -147,3 +149,66 @@ def test_prefetched_objects(self):
         value, prefetched = prefetched_objects(instance, 'groups')
         assert prefetched
         assert value[0] == group
+
+    def test_validate_youtube_video_url(self):
+        valid_youtube_video_urls = [
```
engelhamer, line 154: These tests are not all video URLs and while they do all pass the back end regex, they do not all pass the front end regex. There should not be an inconsistency between these.

src/vue/src/utils/validation.js
```diff
@@ -53,4 +53,17 @@ export default {
 
         return validURL && validProtocol
     },
+
+    /* Taken from https://stackoverflow.com/a/28735961 */
+    validateYouTubeUrl (url) {
+        if (url) {
+            const re = /^(?:https?:\/\/)?(?:m\.|www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((\w|-){11})(?:\S+)?$/ // eslint-disable-line
```
engelhamer, line 60: This regex is different than the one used to extract and display the video at https://github.com/eJourn-al/eJournal/blob/e05e3be0a9845a348f08c72ca1e7c5cb62c837f0/src/vue/src/components/entry/EntryFields.vue#L208 . Should be consistent all throughout.

engelhamer (changes requested):
The Youtube embed is not consistent throughout front and back end implementations. This could cause weird errors in the current form.

src/django/test/test_utils.py
```diff
@@ -147,3 +149,66 @@ def test_prefetched_objects(self):
...
+            'http://youtube.com/watch?v=iwGFalTRHDA',
+            'http://www.youtube.com/watch?v=iwGFalTRHDA&feature=related',
+            'https://www.youtube.com/channel/UCDZkgJZDyUnqwB070OyP72g',
```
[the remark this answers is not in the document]
Larspolo: Guess same for the watch, use our promotion video

src/django/test/test_utils.py
```diff
@@ -154,61 +154,55 @@ def test_validate_youtube_video_url(self):
         valid_youtube_video_urls = [
             'http://youtube.com/watch?v=iwGFalTRHDA',
             'http://www.youtube.com/watch?v=iwGFalTRHDA&feature=related',
-            'https://www.youtube.com/channel/UCDZkgJZDyUnqwB070OyP72g',
-            'https://youtube.com/channel/UCDZkgJZDyUnqwB070OyP72g',
             'https://youtube.com/iwGFalTRHDA',
             'http://youtu.be/n17B_uFF4cA',
             'youtube.com/iwGFalTRHDA',
             'youtube.com/n17B_uFF4cA',
```
Larspolo: These are not valid YT links. Same goes for `https://youtube.com/iwGFalTRHDA`. Bit weird to then include them in the test

engelhamer (approved):

PR #1077, opened 2020-12-30
Only run linters for directories containing changes

Larspolo:
## Description
Only run linters when there are changes in those directories

Speedsup the linter check when only working in backend / frontend


.githooks/pre-commit
```diff
@@ -1,4 +1,10 @@
 #!/bin/sh
 
 make isort
-make check-linters
+if git diff-index --cached --name-only HEAD | grep "src/vue"; then
```
mjvkeulen, line 4: Possibly introduce a check for .js and .vue extensions. E.g. when only editing a sass file running the linter is not required?

mjvkeulen, line 4: I believe you could go one directory deeper with the check, src/vue/src

mjvkeulen (approved):

.githooks/pre-commit
Larspolo: This is the same as the actual linter now

engelhamer (approved):

PR #1079, opened 2021-01-06
Improved code coverage of test suite

Larspolo:
## Description
Added variety of tests to parts that needed some
- preset entry creation
- grading background task
- file retrieval with access id


src/django/VLE/utils/entry_utils.py
```diff
@@ -70,6 +70,8 @@ def create_entry_content(content_dict, entry, user):
             created_content = factory.make_content(entry, content, field)
 
             if field.type == field.FILE:
+                # QUESTION: when is this code reached?
+                # If its required, it already returns after the `check_fields` right?
```
mjvkeulen, line 74: Provided `validate_entry_content` is always called before `create_entry_content`, `VLEMissingRequiredField` should always be raised already and this block never reached.

src/django/test/factory/params.py
```diff
@@ -99,6 +104,43 @@ def _adjust_kwargs(cls, **kwargs):
...
+
+    @classmethod
+    def _adjust_kwargs(cls, **kwargs):
+        journal = kwargs.pop('journal')
+
+        if 'node' in kwargs:
+            node = kwargs['node']
+        else:
+            preset = test.factory.DeadlinePresetNode(format=journal.assignment.format)
+            node = preset.node_set.filter(journal=journal).first()
```
mjvkeulen, line 127: This is because of the one to one issue in the factory right? Possible change this to .get() iso first() for clarity.

src/django/test/factory/params.py
```diff
@@ -99,6 +104,43 @@ def _adjust_kwargs(cls, **kwargs):
...
+
+        kwargs['node_id'] = node.pk
+        kwargs['journal_id'] = journal.pk
+
+        template = preset.forced_template
+        kwargs['template_id'] = template.pk
+
+        author = kwargs.pop('author', random.choice(journal.authors.all()).user)
```
mjvkeulen, line 135: So in python function args are always evaluated. So even if an author is present in kwargs we will query the journal authors.

src/django/test/test_entry.py
```diff
@@ -450,11 +450,54 @@ def test_create_entry(self):
             'When the active LTI uplink is outdated no more entries can be created.'
         self.group_journal.assignment.active_lti_id = assignment_old_lti_id
         self.group_journal.assignment.save()
-
-        # TODO: Test for entry bound to entrydeadline
         # TODO: Test with file upload
         # TODO: Test added index
 
+    def test_create_invalid_preset_entry(self):
+        # Entries cannot be published after lockdate
```
mjvkeulen, line 457: Publishing entries?

src/django/test/test_entry.py
```diff
@@ -450,11 +450,54 @@ def test_create_entry(self):
...
+        node.preset.due_date = timezone.now() - timedelta(weeks=1)
+        node.preset.lock_date = timezone.now() - timedelta(weeks=1)
+        node.preset.save()
+        resp = api.create(self, 'entries', params=create_params, user=self.student, status=400)['description']
+        assert 'lock date' in resp, 'Node should be locked, and not available for new entry'
+        node.preset.due_date = timezone.now() + timedelta(weeks=1)
+        node.preset.lock_date = timezone.now() + timedelta(weeks=1)
+        node.preset.save()
+
+        # Template of that is not bound to node, should not be useable, even from the same assignment
```
mjvkeulen, line 471: `of that is not`

It should not be possible to create a deadline entry using a template other than the preset's template?

src/django/test/test_entry.py
```diff
@@ -450,11 +450,54 @@ def test_create_entry(self):
...
+        invalid_params['template_id'] = factory.Template(format=self.g_assignment.format).pk
+        resp = api.create(self, 'entries', params=invalid_params, user=self.student, status=400)['description']
+        assert 'Invalid template' in resp, \
+            'Template of that is not bound to node, should not be useable, even from the same assignment'
+
+        # Node must be progress node if 'node_id' is supplied
+        node.type = Node.PROGRESS
+        node.save()
+        resp = api.create(self, 'entries', params=create_params, user=self.student, status=400)['description']
+        assert 'EntryDeadline' in resp, 'Node must be progress node if "node_id" is supplied'
```
mjvkeulen, line 482: I don't quite follow this block. 

We try to create an entry for a progress node. 

> 'Node must be progress node if "node_id" is supplied'

The node is now a progress node right?

src/django/test/test_entry.py
```diff
@@ -450,11 +450,54 @@ def test_create_entry(self):
...
+
+        # Passed node already contains an entry, this should also be indicated correctly
+        api.create(self, 'entries', params=create_params, user=self.student)['description']
+        resp = api.create(self, 'entries', params=create_params, user=self.student, status=400)['description']
+        assert 'already contains an entry' in resp, \
+            'Passed node already contains an entry, this should also be indicated correctly'
+
+        # Passed fields should be from template
+        invalid_params = create_params.copy()
+        factory.Template(format=factory.Assignment().format)
```
mjvkeulen, line 494: Possible to reuse the assignment?

src/django/test/test_file_handling.py
```diff
@@ -132,7 +133,14 @@ def test_file_retrieve(self):
...
         api.get(self, 'files', params={'pk': file.pk}, user=self.student, status=200)
-        api.get(self, 'files', params={'pk': file.pk}, user=factory.Teacher(), status=403)
+        api.get(self, 'files', params={'pk': file.pk}, user=other_teacher, status=403)
+
+        # Test non related user on file without context (e.g. profile picture)
+        file = FileContext.objects.create(file=self.video, author=self.student, file_name=self.video.name)
+        api.get(self, 'files', params={'pk': file.pk}, user=self.student)
+        api.get(self, 'files', params={'pk': file.pk}, user=other_teacher, status=403)
+        api.get(self, 'files', params={'pk': file.pk, 'access_id': file.access_id}, user=other_teacher)
+        api.get(self, 'files/access_id', params={'pk': file.access_id}, user=other_teacher)
```
mjvkeulen, line 143: Possibly sprinkle some comments on why the individual cases differ and what the expected outcome is.

mjvkeulen: Makes me think of another param to api.get, msg -> use as assert msg when status does not match?

src/django/test/test_gradepassback.py
```diff
@@ -185,3 +185,21 @@ def test_replace_result_no_url(self):
             author.save()
 
         assert not grading.send_journal_status_to_LMS(self.lti_journal)['successful']
+
+    def test_grading_background_task(self):
+        journal = factory.Journal()
```
mjvkeulen, line 190: Possible to use the setup assignment?

src/django/test/test_gradepassback.py
```diff
@@ -185,3 +185,21 @@ def test_replace_result_no_url(self):
...
+        author = journal.authors.first()
```
mjvkeulen, line 191: We have a shorthand for this (which is also 1 query less), `journal.author`

src/django/test/test_gradepassback.py
```diff
@@ -185,3 +185,21 @@ def test_replace_result_no_url(self):
...
+            'When author is not member of journal, it should say so'
+
+        resp = grading.send_author_status_to_LMS(journal, author)
+        assert 'has no sourcedid' in resp['description'], \
+            'When author has no sourcedid url, it should say so'
+        author.sourcedid = 'asdf'
+        author.save()
+        resp = grading.send_author_status_to_LMS(journal, author)
+        assert 'has no grade_url' in resp['description'], \
+            'When author has no greade url, it should say so'
```
mjvkeulen, line 205: typo, clear comments to these tests tho ^^

mjvkeulen (changes requested):
Good stuff, some of these tests cover essential business logic. Silly they were not tested!

src/django/test/test_gradepassback.py
Larspolo: Its nowhere else needed though. This needs to be a non LTI journal specifically, the rest tests with LTI

src/django/test/test_file_handling.py
Larspolo: Currently doesn't exists right? Will do it in comments instead for now

src/django/test/test_gradepassback.py
Larspolo: `journal.author` gets the user (which is generally wanted), but now we want the AP

src/django/test/test_entry.py
Larspolo: Updated to "You should not be able to create a note in a PROGRESS node"

src/django/test/factory/params.py
Larspolo: Yes, did it

src/django/test/test_gradepassback.py
mjvkeulen: You know how often I make this mixup, damn our phrasing sometimes.

mjvkeulen (approved):

engelhamer (approved):

PR #1080, opened 2021-01-14
Fixed inconsistent entry title display

mjvkeulen:
Fixes some inconsistencies, as well as fixing Z#1478


src/vue/src/components/entry/EntryTitle.vue
```diff
@@ -0,0 +1,39 @@
...
+        {{ title }}
+    </h2>
+</template>
+
+<script>
+export default {
+    name: 'EntryTitle',
+    props: {
+        template: {
+            required: true,
```
Larspolo, line 12: [NIT] this doesnt have to be required right? If you are sure `node.entry.title` is set, then this works

src/vue/src/components/entry/EntryTitle.vue
```diff
@@ -0,0 +1,39 @@
...
+        },
+    },
+    computed: {
+        title () {
+            if (this.node.entry && 'title' in this.node.entry) {
+                return this.node.entry.title /* Set backend, this should always be correct */
+            }
+
+            if (this.node.deleted_preset) {
+                return this.node.entry.title
```
Larspolo, line 28: [NIT] This should not be possible to be a thing, if title in in `this.node.entry`, it should be earlier already accessed right?

Larspolo (approved):

src/vue/src/components/entry/EntryTitle.vue
mjvkeulen: True, but we have a linter rule which requires a default to be set when not required. Should I make it an empty object? This might also indicate the component just works without the template. Not too sure what to do, any suggestion?

Larspolo: Think a default could be set to an empty object, but I see that can also create "false hope" kinda. Think leaving as is is for now better

engelhamer (approved):

PR #1081, opened 2021-01-29
Add categories to templates, template and preset node views, assignment editor rework, import file bugfix

mjvkeulen:
- Categories can now be added via template edit
- Format editor reworked. Each component now has a display and edit mode. Each instance holds its own draft.
- Dedicated views to the models `PresetNode` and `Template`, each with their own validation front and backend.
- Files are now properly copied on import/copy, before removing a file from the source (e.g. in rich-text) would also delete the file for the target. 

## Summary of changes
- Fixes Z#1477 Partially created entry preset can be lost in FormatEditor
- Fixes Z#1454 The assignment save FAB appears without changes for a group assignment
- Fixes Z#1443 Expand permission validation to format update
- Fixes Z#55 Back end assignment format validation
- Fixes Z#90 Indication of incorrect changes upon saving edited assignment (remaining issue better encapsulated in #1482)

## Questions
- What still needs to be done to satisfy Z#95 (Improve assignment editor)?


src/django/VLE/serializers.py
```diff
@@ -414,6 +417,10 @@ class Meta:
         'attached_files',
     ]
 
+    order_by = [
+        'due_date',
+    ]
```
Larspolo, line 422: [NIT/Q]
```
            queryset=PresetNodeSerializer.setup_eager_loading(
                VLE.models.PresetNode.objects.order_by('due_date')
            ),
```
the order by is now no longer necessary right?

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
```diff
@@ -0,0 +1,80 @@
...
+        }),
+        ...mapMutations({
+            clearAssignmentDetailsDraft: 'assignmentEditor/CLEAR_ASSIGNMENT_DETAILS_DRAFT',
+            selectAssignmentDetails: 'assignmentEditor/SELECT_ASSIGNMENT_DETAILS',
+            setModeToRead: 'assignmentEditor/SET_ACTIVE_COMPONENT_MODE_TO_READ',
+        }),
+        save () {
+            if (!this.$refs.assignmentDetails.validateData()
+                || !this.$refs.assignmentDetails.$refs.assignmentDetailsDates.validateData()) {
+                return
```
Larspolo, line 68: [NIT] Somewhere an indication into which are still to be saved would make this PR even better. If not for now, I would defenitly see this as a future improvement

src/django/VLE/tasks/beats/cleanup.py
```diff
@@ -59,10 +61,21 @@ def remove_unused_assignment_files():
...
+
+    # Not in RT description, but via M2M 'attached_files'
+    fc_ids_attached_to_preset_nodes = (
+        VLE.models.PresetNode.attached_files  # M2M Manager
+        .through  # Directly acces the Django created M2M table of FC to PresetNode and vice versa
+        .objects  # Use the default manager for the M2M table
+        .values_list('filecontext__pk', flat=True)
+        .distinct()
+    )
+    ass_fcs_without_reference = ass_fcs_without_reference.exclude(pk__in=fc_ids_attached_to_preset_nodes)
```
Larspolo, line 76: Nice comments! Did this mean that before this change attached files to preset nodes would have been deleted?

src/django/VLE/utils/file_handling.py
```diff
@@ -183,14 +178,40 @@ def get_temp_files_from_rich_text(rich_text):
     return get_files_from_rich_text(rich_text).filter(is_temp=True)
 
 
+def copy_and_replace_rt_files(rich_text, user, assignment=None, category=None):
```
Larspolo, line 181: This function name feels a bit too brought for what it does. Currently only supports assignment / categories right?

Would either change the name, or widen the scope. Maybe import_utils.py:182 (inside `import_content`) could make use of it as well (and possibly add for non rt files then as well

src/django/VLE/utils/generic_utils.py
```diff
@@ -48,177 +45,62 @@ def optional_params(post, *keys):
...
 
 
-def required_typed_params(post, *keys):
-    if keys and not post:
-        raise VLE.utils.error_handling.VLEMissingRequiredKey([key[1] for key in keys])
+def cast_value(value, type, optional=False):
+    if value is None:
+        return None
+
+    if type == bool and value == 'false':
```
Larspolo, line 52: Nice function :)

[NIT] In `get_lti_params_from_jwt` we do ` == 'true'` for boolean fields. That could also make use of this

src/django/VLE/utils/import_utils.py
```diff
@@ -240,8 +242,13 @@ def bulk_import_assignment_categories(source_assignment, target_assignment, auth
...
 
-    # QUESTION: Copy any associated files? Might not be needed -> Edit in source means target still refs the source fc
-    # Provided access to the fc is only by access id, this should be fine?
+    # RT Categories require their respective 'category' attribute to be set, which is why we first have to create
+    # the new categories before we can copy and replace the RT files.
+    categories_with_copied_rt_description = []
+    for category in target_assignment.categories.all():
+        category.description = copy_and_replace_rt_files(category.description, author, category=category)
+        categories_with_copied_rt_description.append(category)
+    Category.objects.bulk_update(categories_with_copied_rt_description, ['description'])
```
Larspolo, line 251: [Q] would this be useful to all be put inside a `with transation`?

src/django/VLE/utils/import_utils.py
```diff
@@ -294,3 +301,28 @@ def copy_node(node, journal, entry=None):
...
+            is_temp=False,
+            in_rich_text=False,
+            assignment=assignment,
+            category=None,
+            content=None,
+            comment=None,
+            journal=None,
+        )
+        for old_fc in attached_files
+    ]
```
Larspolo, line 327: [Q] can this be done in a `bulk_create`?

src/django/VLE/views/assignment.py
```diff
@@ -183,63 +220,43 @@ def retrieve(self, request, pk=None):
...
-            unauthorized -- when the user is not logged in
-            not found -- when the assignment does not exist
-            forbidden -- User not allowed to edit this assignment
-            unauthorized -- when the user is unauthorized to edit the assignment
-            bad_request -- when there is invalid data in the request
-        On success:
-            success -- with the new assignment data
+    # QUESTION: There are a lot of step which change assignment data, wrap the method in an atomic decorator?
+    # This seems like it would be almost the desired default behaviour for most of our logic
+    # Worth spending some time looking into the advantages / disadvantages?
```
Larspolo, line 225: The shorter we lock, the sooner its available elsewhere right? Maybe we could start the transaction only after the permission check, but I still like the addition of it here over nothing

src/django/VLE/serializers.py
```diff
@@ -507,7 +486,10 @@ class Meta:
             'categories',
             # Not used / missing: active_lti_id, lti_id_set, assigned_groups, format
```
Larspolo, line 487: [NIT] I'm not sure when it was added, but this comment seem to now not be up to date, as format is now also added.
[Q] Why is this added? Should it be added to the read_only fields?

src/django/VLE/serializers.py
```diff
@@ -507,7 +486,10 @@ class Meta:
...
         )
-        read_only_fields = ()
+        read_only_fields = (
+            'active_lti_course',
+            'lti_id',
```
Larspolo, line 491: [Q] Should `id` also be included in this list? Or is it by default? Or is it needed from importing assignment?

src/django/VLE/views/assignment.py
```diff
@@ -475,13 +492,15 @@ def copy(self, request, pk):
         # Many to many field requires manual update, only set the course we are importing into
         assignment.courses.set([])
         assignment.add_course(course)
+        # Requires assignment with accurate pk
```
Larspolo, line 495: [Q] This funciton could also make use of a atomic transation thing right?

src/django/VLE/views/format.py
```diff
@@ -1,152 +0,0 @@
-"""
```
Larspolo, line 1: Nice :D

src/django/test/factory/assignment.py
```diff
@@ -98,3 +99,30 @@ def courses(self, create, extracted, **kwargs):
...
+
+    @classmethod
+    def _adjust_kwargs(cls, **kwargs):
+        assignment = kwargs.pop('assignment')
+        if not assignment:
+            assignment = factory.Assignment()
+
+        kwargs['pk'] = assignment.pk
+
+        kwargs['course_id'] = assignment.courses.first().pk
```
Larspolo, line 116: [NIT] could possibly change to sth like `        kwargs['course_id'] = kwargs.get('course_id', assignment.courses.first().pk)`

src/django/VLE/views/preset_node.py
```diff
@@ -0,0 +1,161 @@
...
+            preset.forced_template_id = template['id']
+            preset.unlock_date = unlock_date
+            preset.lock_date = lock_date
+
+        with transaction.atomic():
+            preset.save()
+            _update_attached_files(preset, request.user, attached_file_ids)
+            file_handling.establish_rich_text(author=request.user, rich_text=preset.description, assignment=assignment)
+
+        return response.success({
```
Larspolo, line 147: I was able to change the progress node's out of order:
image
This was not possible before and gave a validation error

Larspolo (changes requested):
Checked the backend code, and clicked through the frontend. Still frontend code to go, but I already like the addition of the store modules.

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentStartReadMode.vue
```diff
@@ -0,0 +1,39 @@
+<template>
```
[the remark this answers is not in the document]
mjvkeulen: This is caused by a switch component changing subcomponents via `v-if`. I will have to look into a solution, I think this approach is applied quite commonly by us, but I agree it is quite visible here.

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
```diff
@@ -0,0 +1,80 @@
...
+    methods: {
+        ...mapActions({
+            updateAssignment: 'assignment/update',
+        }),
+        ...mapMutations({
+            clearAssignmentDetailsDraft: 'assignmentEditor/CLEAR_ASSIGNMENT_DETAILS_DRAFT',
+            selectAssignmentDetails: 'assignmentEditor/SELECT_ASSIGNMENT_DETAILS',
+            setModeToRead: 'assignmentEditor/SET_ACTIVE_COMPONENT_MODE_TO_READ',
+        }),
+        save () {
```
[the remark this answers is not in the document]
mjvkeulen: Point one and two fixed, not sure how to address point 3. This is simply the router actually navigating to a different page. Would you like to see a different type of transition?

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -1,159 +1,219 @@
...
 
-            if (!this.validCategory) { return }
+            if (this.create) {
+                this.categoryCreate({ category: this.category, aID: this.$route.params.aID })
+                    .then((category) => { this.categoryCreated({ category }) })
+            } else {
+                this.categoryUpdate({ id: this.category.id, category: this.category, aID: this.$route.params.aID })
+                    .then(() => { this.categoryUpdated({ category: this.category }) })
+            }
+        },
```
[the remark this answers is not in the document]
mjvkeulen: What is the exact order to reproduce this? If I simply change the templates of a category, or the categories of a template and save I cannot reproduce this?

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -1,159 +1,219 @@
...
+            </b-button>
+        </b-row>
+
+        <template v-if="readMode">
+            <category-display
+
+                :id="`category-${category.id}-display`"
+                :editable="false"
+                :categories="[category]"
             />
```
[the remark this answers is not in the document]
mjvkeulen: Atm, I have been consistent in using our display components as true as possible to the normal views. Do you disagree with this design choice? Or just break it in this one instance?

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -37,7 +37,9 @@
                     <span v-else>This template can be freely used by students as often as they want</span>
                 </b-td>
             </b-tr>
-            <b-tr>
+
+            <!-- Question: @ Lars, hide this as well or hide nothing? -->
```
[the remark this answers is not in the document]
mjvkeulen: I am not in favor either way, so atm we have you on the one side and Engel on the other.

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,241 @@
...
+                    placeholder="Timeline display name"
+                    trim
+                    required
+                />
+            </b-form-group>
+
+            <preset-node-edit-select-and-preview-template
+                v-if="presetNode.type === 'd'"
+                :presetNode="presetNode"
+            />
```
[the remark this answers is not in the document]
mjvkeulen: Ah tricky, updating the template actually creates a new one (due to the archiving process). I would have to think about a clever solution since we cannot even rely on name as identifiers atm.

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,241 @@
...
+                </template>
+
+                <files-list
+                    :attachNew="true"
+                    :files="presetNode.attached_files"
+                    @uploading-file="uploadingFiles ++"
+                    @fileUploadSuccess="presetNode.attached_files.push($event) && uploadingFiles --"
+                    @fileUploadFailed="uploadingFiles --"
+                    @fileRemoved="(i) => presetNode.attached_files.splice(i, 1)"
+                />
```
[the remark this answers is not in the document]
mjvkeulen: Good catch, was some bad ordering of some filters (and clearly bad tests). 

What should the display be for attached files of a deadline preset node? Added a TODO for this point.

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,241 @@
...
+                :presetNode="presetNode"
+            />
+
+            <b-form-group>
+                <template
+                    v-if="presetNode.attached_files.length > 0"
+                    #label
+                >
+                    Files
+                </template>
```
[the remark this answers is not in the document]
mjvkeulen: I dont mind either way, removing the if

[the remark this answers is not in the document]
mjvkeulen: Tag the right line tho! :p

src/vue/src/store/modules/assignmentEditor.js
```diff
@@ -0,0 +1,417 @@
...
+    cancelPresetNodeEdit (context, { presetNode }) {
+        const presetNodes = context.rootGetters['presetNode/assignmentPresetNodes']
+        const originalPresetNode = presetNodes.find(elem => elem.id === presetNode.id)
+
+        context.commit('CLEAR_DRAFT', { drafts: context.state.presetNodeDrafts, obj: presetNode })
+        context.commit('SELECT_PRESET_NODE', { presetNode: originalPresetNode })
+    },
+    cancelTemplateEdit (context, { template }) {
+        const templates = context.rootGetters['template/assignmentTemplates']
+        const originalTemplate = templates.find(elem => elem.id === template.id)
```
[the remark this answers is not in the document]
mjvkeulen: Cancel edit should only occur on objects already created (and thus part of modules category/template etc). Thus hopefully the original should always be found. 

Weird error, hopefully, we get a better trace at some point?

src/vue/src/views/Course.vue
```diff
@@ -143,7 +143,7 @@ export default {
         },
         handleCreated (aID) {
             this.$router.push({
-                name: 'FormatEdit',
+                name: 'AssignmentEditor',
```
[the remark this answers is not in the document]
mjvkeulen: Good catch, easy fix but also have to move the assignmentAPI create function to the respective module to update local state. Will do that in a couple hours.

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentStartReadMode.vue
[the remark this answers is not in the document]
Larspolo: Maybe its also a css transition that is being applied? Thats what I was thinking. It is not on the other card

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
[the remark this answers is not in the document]
Larspolo: I would expect it to not have a transition at all, or at least a transition that doesn't display the breadcrumb etc twice

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
Larspolo: I totally agree with the design choice, but in this specific instance I would expand to also view the component as it would be viewed when clicking the "(i)"  icon

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
Larspolo: 1. Go to a category
2. Click edit
3. Select a template that was not selected before
4. Click save
5. Click anywhere on the breadcrumb
6. View the (unexpected) prompt

src/vue/src/components/template/TemplateEditSettings.vue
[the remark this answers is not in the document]
Larspolo: Lets put the final say to @engelhamer then

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
Larspolo: Think as is works fine. This is also the same as a student would view it

image

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentStartReadMode.vue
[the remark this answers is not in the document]
engelhamer: I will check it out now

src/django/VLE/models.py
```diff
@@ -1249,6 +1251,52 @@ def state_actions(cls, new, old=None):
...
+            raise ValidationError('Locks before due date.')
+
+        if self.format.presetnode_set.filter(unlock_date__gt=self.lock_date).exists():
+            raise ValidationError('One or more deadlines unlock after the assignment locks.')
+
+        if self.format.presetnode_set.filter(due_date__gt=self.lock_date).exists():
+            raise ValidationError('One or more deadlines are due after the assignment locks.')
+
+        if self.format.presetnode_set.filter(lock_date__gt=self.lock_date).exists():
+            raise ValidationError('One or more deadlines are locked after the assignment locks.')
```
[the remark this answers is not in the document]
engelhamer: [NIT]: would like this to be rephrased to 'have a lock date after..' (i.e. use 'lock date' is 'locks' everywhere). This way we use the exact same term the user sees and mitigate unneeded confusion from having to realise that 'locks' refers to this.

src/django/VLE/tasks/beats/cleanup.py
mjvkeulen: Yes, the current bug is quite severe, if any files of a copied assignment / imported template are changed/removed in the original, they are also removed from the copied assignment/template.

src/vue/src/components/format/timeline_controlled/assignment/AssignmentEdit.vue
```diff
@@ -0,0 +1,68 @@
+<template>
```
engelhamer: [NIT] Would `AssignmentDetailsEdit` not make more sense for a name? The whole format thing is actually assignment edit.

src/vue/src/components/format/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,227 @@
...
+                <icon name="trash"/>
+                Delete Preset
+            </b-button>
+
+            <b-button
+                class="green-button float-right"
+                @click.stop="finalizePresetNodeChanges()"
+            >
+                <icon :name="(edit) ? 'save' : 'plus'"/>
+                <!-- QUESTION: User facing it might make more sense to use 'deadline' over 'preset'? -->
```
engelhamer: Yes, if not already done (reviewing earlier commit) then this is a nice change to still make.

src/vue/src/components/template/TemplateEditSettings.vue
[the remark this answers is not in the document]
engelhamer: I would like the category stuff to be hidden until a category is created, to make it easier for new users to get started and not be intimidated by all options as much. This is also in line with some informal user feedback we have had in the past.

src/vue/src/components/template/TemplateImport.vue
```diff
@@ -1,138 +1,127 @@
...
+                    placeholder="Select A Template"
+                    class="multi-form"
+                    @select="() => {
+                        if (previewTemplate) {
+                            previewTemplate = null
+                        }
+                    }"
                 />
 
-                <category-display
```
engelhamer: I can imagine scenarios where the categories would be nice to copy as well. Maybe if we support importing categories from another assignment this should be reconsidered (create an issue)?

engelhamer (commented):
Until 04d124a

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
engelhamer: You have a template chain, which should be able to fix this issue?

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
engelhamer: That is true, but this 'view' of a category on it's own (so not part of the list for selecting it within an entry) is not viewable for students at all. I think it would not hurt also including the selected templates as well (since it now looks abit 'empty', right?).

src/vue/src/views/FormatEdit.vue
```diff
@@ -130,10 +142,11 @@ New nodes can be added via the '+' node. Click any node to view its contents.
 
         Promise.all(init).then(() => {
             this.loading = false
+            this.$intro().showHints()
 
             if (this.savedPreferences.show_format_tutorial) {
                 this.changePreference({ show_format_tutorial: false })
-                this.$intro().start()
+                // this.$intro().start()
```
engelhamer: Comment just to ensure this is re-enabled later.

src/vue/src/components/assignment_editor/AssignmentEditorBreadCrumb.vue
```diff
@@ -0,0 +1,18 @@
+<template>
```
engelhamer, line 1: I think that moving this to two nested components harms readability of the assignment editor. Why do you prefer this rather than keeping it in the same file? This method:
- introduces more lines than needed (overhed `template` tags, `export default` etc.)
- Requires you to click through to an additional file rather than where it is used
- Results in way more files, and this particular example has only 3 logical lines really?

engelhamer (commented):
All code only

src/vue/src/components/assignment/AssignmentDetailsDates.vue
```diff
@@ -0,0 +1,259 @@
...
+                } else {
+                    this.lockDateInputState = null
+                }
+            },
+        },
+    },
+    methods: {
+        firstAfterSecond (first, second) {
+            return first && second && Date.parse(first) > Date.parse(second)
+        },
```
Larspolo, line 253: [NIT] Woudl prefer removing this function, and flipping the parameters, this makes it for me more readable as you only have to look at the order of parameters, and not also first at the function name

src/vue/src/components/assignment/AssignmentDetailsDates.vue
```diff
@@ -0,0 +1,259 @@
...
+                    additionalConfig.minDate = node.lock_date
+                }
+            })
+
+            if (this.assignment.due_date && (!additionalConfig.minDate
+                || new Date(this.assignment.due_date) > additionalConfig.minDate)) {
+                // Assignment has due date later than current min.
+                additionalConfig.minDate = new Date(this.assignment.due_date)
+            } else if (this.assignment.unlock_date && (!additionalConfig.minDate
+                || new Date(this.assignment.unlock_date) > additionalConfig.minDate)) {
```
Larspolo, line 177: [NIT] I think here you can also make use of the `firstBeforeSecond` function right?

src/vue/src/components/assignment_editor/AssignmentEditorBreadCrumb.vue
Larspolo: A positive I see is that if we were to make further changes to the breadcrumb specific, then it will be handled in its specific file.
Another downside I see is that creating components may impact performance. That said, I don't think this single one makes much of a difference

Could go either way for me

src/vue/src/components/assignment_editor/timeline_controlled/SelectedTimelineComponentSwitch.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+    <div>
```
Larspolo, line 2: [NIT][ Is this div needed?

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEndReadMode.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+    <journal-end-card :assignment="assignment">
+        <template v-slot:edit-button>
+            <b-button
+                class="ml-2 orange-button float-right"
+                @click.stop="setModeToEdit()"
+            >
+                <icon name="edit"/>
+                Edit
```
Larspolo, line 9: [NIT] I would think that this edit button might cause confusion on what is being edited

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,240 @@
...
+                    Delete
+                </b-button>
+
+                <b-button
+                    class="green-button ml-auto"
+                    @click.stop="finalizePresetNodeChanges()"
+                >
+                    <icon :name="(edit) ? 'save' : 'plus'"/>
+                    <!-- QUESTION: User facing it might make more sense to use 'deadline' over 'preset'? -->
+                    {{ (edit) ? 'Save' : 'Add Preset' }}
```
Larspolo, line 127: [NIT] brackets shouldnt be necessary right?

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -0,0 +1,240 @@
...
+                    <icon name="trash"/>
+                    Delete
+                </b-button>
+
+                <b-button
+                    class="green-button ml-auto"
+                    @click.stop="finalizePresetNodeChanges()"
+                >
+                    <icon :name="(edit) ? 'save' : 'plus'"/>
+                    <!-- QUESTION: User facing it might make more sense to use 'deadline' over 'preset'? -->
```
Larspolo, line 126: Totally agree

src/vue/src/components/entry/Entry.vue
```diff
@@ -17,27 +17,23 @@
...
-                    </b-button>
-                    <b-button
-                        class="ml-2 orange-button float-right multi-form"
-                        @click="edit = true"
-                    >
-                        <icon name="edit"/>
-                        Edit
-                    </b-button>
-                </div>
             </template>
```
Larspolo, line 20: I can no longer create entries

image

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeReadMode.vue
```diff
@@ -0,0 +1,68 @@
...
+                    class="ml-auto orange-button"
+                    @click.stop="setModeToEdit()"
+                >
+                    <icon name="edit"/>
+                    Edit
+                </b-button>
+            </template>
+        </progress-node>
+
+        <!-- TODO Category: This does not display any attached files -->
```
Larspolo, line 21: Huh, this is quite the bug, no student can currently see the attached files. only in progress nodes

src/vue/src/store/modules/assignmentEditor.js
```diff
@@ -0,0 +1,417 @@
...
+        state.activeComponentMode = mode
+    },
+    CREATE_PRESET_NODE (state) {
+        if (state.newPresetNodeDraft) {
+            state.selectedPresetNode = state.newPresetNodeDraft
+        } else {
+            const newPreset = {
+                id: -1,
+                type: null,
+                template: '',
```
Larspolo, line 167: [Q] why si template '' and not null?

src/vue/src/store/modules/assignmentEditor.js
```diff
@@ -0,0 +1,417 @@
...
+            const newTemplate = {
+                field_set: [{
+                    type: 'rt',
+                    title: 'Content',
+                    description: '',
+                    options: null,
+                    location: 0,
+                    required: true,
+                }],
+                name: 'Entry',
```
Larspolo, line 127: [NIT] Maybe we should leave this empty. Seems only logical

src/vue/src/store/modules/assignment.js
```diff
@@ -1,31 +1,65 @@
...
+            return auth.get(`assignments/${id}`, { course_id: cID }, connArgs)
+                .then((response) => {
+                    context.commit('SET_ASSIGNMENT', { assignment: response.data.assignment })
+                    return response.data.assignment
+                })
+        }
+
+        return fromCache(context, 'retrieveCache', id, fn.bind(null), force)
+    },
+    update (context, { id, data, connArgs = auth.DEFAULT_CONN_ARGS }) { // eslint-disable-line
```
Larspolo, line 42: [NNIT]] could specify which rule you are disabeling so other stuff doesnt also become more ugly

src/vue/src/store/modules/assignment.js
```diff
@@ -1,31 +1,65 @@
...
+        function fn () {
+            return auth.update(`assignments/${id}`, data, connArgs)
+                .then((response) => {
+                    context.commit('SET_ASSIGNMENT', { assignment: response.data.assignment })
+
+                    return response.data.assignment
+                })
+        }
+
+        return fn()
```
Larspolo, line 52: [Q] Can this now just be return `auth.update`? Would prefer that. Dont think it can use cache anyway

src/vue/src/router/index.js
```diff
@@ -211,10 +211,14 @@ router.beforeEach((to, from, next) => {
...
 router.afterEach((to, from) => {
     if ('aID' in to.params) {
+        store.dispatch('assignment/retrieve', { id: to.params.aID })
+        store.dispatch('presetNode/list', { aID: to.params.aID })
         store.dispatch('category/list', { aID: to.params.aID })
+        store.dispatch('template/list', { aID: to.params.aID })
 
         if ('aID' in from.params && parseInt(to.params.aID, 10) !== parseInt(from.params.aID, 10)) {
-            store.commit('category/clearFilteredCategories')
+            store.commit('category/CLEAR_FILTERED_CATEGORIES')
```
Larspolo, line 220: [Q] Why is this not included in `            store.commit('assignmentEditor/RESET')`?

src/vue/src/helpers/global_mixins.js
```diff
@@ -0,0 +1,20 @@
...
+
+                    if (key.includes('State') && value === false) {
+                        const invalidFeedback = this._data[key.replace('InputState', 'InvalidFeedback')]
+                        this.$toasted.error(invalidFeedback)
+                        return true
+                    }
+                    return false
+                })
+
+                return !invalid
```
Larspolo, line 16: [NIT] would prefer to swap the variable name in the foreach to valid and returning false when sth is invalid

keep_testing.sh
```diff
@@ -14,8 +14,8 @@ test () {
     stty size | perl -ale 'print "-"x$F[1]'
```
Larspolo, line 14: Although not touched in this PR, Preset DB doesn't work (anymore)

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentStartReadMode.vue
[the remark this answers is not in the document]
engelhamer: Was quite a fun bug: caused by the load wrapper. Moved it to a more appropriate position now.

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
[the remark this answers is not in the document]
engelhamer: Point 3 was the same reason as the other fade thing with assignment details (load wrapper). Fixed now.

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
mjvkeulen: Whereas I don't think I had this issue, I did find some propagation issues where drafts were stale. Could you confirm this is resolved? These are steps I took to resolve/test it manually, if you find something, mind adding the route to this list? Will save us time when extending the frontend checklist.

# Test AssignmentEditor

Global rules:

- Selecting a component shows a central view in "read-mode"
- Only "create new" buttons start in edit mode by default
- "create mode" views do not show a "delete button" and their edit button is shown as "preview" (when applicable)
- Can only enter "edit mode" from read mode via edit button
- Can only save from edit mode
- Deleting a currently active component -> Go to the default view (assignment details)
- When navigating away from a component that holds changes (made in edit mode) without saving, reselecting the component
should now start in edit mode, showing the unsaved changes.
- Making any changes should only trigger the visual dirty flag for that component, e.g. updating a template's categories
should not trigger dirty on the respective categories.

## Draft M2M propagation

### Template perspective

#### Single draft manipulations
- Select template -> Edit -> Add category -> Save -> Observe template added to category and category to template; nothing dirty
- Select template -> DELETE any associated category directly via menu -> Observe draft is updated without dirty set; template component view remains active

#### Double draft manipulation
- Select template -> Edit -> Add 1 to name -> Select category -> Edit -> Add previous template -> Save -> Revisit template ->
- Remove 1 from name -> Observe no longer dirty -> Save -> Observe both template and category are not dirty.

### Category perspective

#### Single draft manipulations
- Select category -> Edit -> Add template -> Save -> Nothing is dirty -> Select added template -> Observe category is added; Template is not dirty
- Select category -> DELETE any associated template directly via menu -> Observe draft is updated without dirty set; category component view remains active

#### Double draft manipulation
- Select category -> Edit -> Add 1 to name -> Select template -> Edit -> Add previous category -> Save -> Revisit category ->
Remove 1 from name -> Observe no longer dirty -> Save -> Observe both template and category are not dirty

### TODO/Order
- Create a new assignment (without categories)
    - Observe: Timeline filter hidden; Categories menu collapsed (Create New Category is not visible); Template edit "fixed categories" option hidden
    - Add 1 category -> All above becomes visible
- Tests for timeline filter as one edits the assignment

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
mjvkeulen: I added the requested information, including a tooltip that the visual tag of the category is how most of the users will see the category.

image

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
mjvkeulen: The issue purely issues with syncing state across different objects frontend. This is slightly complicated as a template update can in fact create a new template (old archived, new version is created). Which needs to be synced with both the categories data and preset node forced templates.

Since most was kept sync already, this should be fixed in 457168fe6711b735ac4cc7cb34742bbcb4676127

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
mjvkeulen: So was this resolved?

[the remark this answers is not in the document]
mjvkeulen: > What should the display be for attached files of a deadline preset node? Added a TODO for this point.

Still not sure what to do here, entry preview does not include attached files, your screenie only shows a progress goal, what should the read mode be of an entry deadline?

src/vue/src/components/template/TemplateEditSettings.vue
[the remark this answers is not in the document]
mjvkeulen: I assume this is resolved then @Larspolo and remove the question

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: How would you like to see your proposal applied to the following example? (Which I think reads quite naturally as is)
`Deadline locks before the assignment is due`

src/django/VLE/models.py
```diff
@@ -2190,6 +2302,143 @@ class Meta:
...
+        validate_display_name()
+        validate_attached_files()
+        PresetNode.validate_due_date(data, assignment)
+
+        if type == Node.PROGRESS:
+            validate_progress_specific_fields()
+        elif type == Node.ENTRYDEADLINE:
+            validate_deadline_specific_fields()
+        else:
+            raise ValidationError('Unknown preset node type.')
```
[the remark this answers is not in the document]
mjvkeulen: Kinda sad it is basically circular imports blocking us from a gradual refactor right?

src/django/VLE/serializers.py
mjvkeulen: You are right, I also think adding ordering to template (by 'name') means we can remove it anywhere where the default manager is used.

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
mjvkeulen: Is this regarding https://zube.io/ejourn-al/ejournal/c/1482 (Highlight validation errors in the timeline), or something other?

For example, I could make the mixin `validateData` (global_mixins.js), also set all validation states to false on validation (whereas they could possibly still be null instead (indicating neither pass or fail)).

src/django/VLE/utils/file_handling.py
mjvkeulen: So currently it is used for `Assignment`, `Template`/`Field`, `PresetNode` and `Category`,

copy_rt_files_and_update_rt_references as new name? 

Content has both RT files and single content file fields, making it less suited to use this function (as it not just RT files).

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
Larspolo: I noticed later as well yeah. I think you could add the files component after the preset node description

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEdit.vue
Larspolo: No, thisis about some components that have a draft open to have an indication rhat they are open. We discussed rhis also in the scrum meeting

src/django/VLE/utils/file_handling.py
Larspolo: copy_rt_files_and_update_rt_references wouldn't fix it. `copy_assignment_related_rt_files`? Because it doesn't work for course files, entry files, comment files, which also have rt

src/django/VLE/utils/import_utils.py
mjvkeulen: This is related to the question I posed in assignment.py:

> There are a lot of steps which change assignment data, wrap the method in an atomic decorator?
This seems like it would be almost the desired default behavior for most of our logic
Worth spending some time looking into the advantages/disadvantages?

I believe it makes sense to wrap the entire `copy` assignment in a transaction block, it does not make sense to have half of a copied assignment ever. However, I still do not fully grasp the impact of the transaction block itself.

For example, if you use a transaction block and only execute read operations, nothing will be locked. If you do perform updates, I believe the full row is locked (not just the column of the updated fields), and I do not know if this lock lasts until the end of the entire transaction (I believe not, just until the end of the update), but I am not entirely sure. How this works is postgres (not Django) specific, and would require some research before I am fully comfortable using it in such large functions.

src/django/VLE/utils/import_utils.py
mjvkeulen: I believe so, but that would also mean we hold all files open in memory, not sure if that is preferred? (memory and fd wise)

src/django/VLE/views/assignment.py
mjvkeulen: So regarding the more general issue:

Do we want to invest some time into researching how postgres handles transactions by default? 
Do we want to make use of more transactions in the backend?

Move to issue?

src/django/VLE/serializers.py
mjvkeulen: Id is read-only by default, I removed all ids from the read-only list in a refactor a couple of months ago.

src/django/VLE/serializers.py
mjvkeulen: So I am able to update an assignment's format via the serializer with this setting, so good catch (provided we have _floating_ format somehwere). 

I don't believe we require the format (pk) to be serialized to begin with, will simply remove format.

src/django/VLE/views/assignment.py
mjvkeulen: Yes, let's aggregate this discussion into an issue.

src/django/VLE/views/preset_node.py
mjvkeulen: Should be fixed backend in: 3ea2d6ddb7cfd5e3851235f6289ca5e33ecd4e6f
Frontend in: 029300694f30ef9cb88ead4a1c2b357a0e6e3097

src/django/VLE/models.py
```diff
@@ -2574,6 +2853,18 @@ class Template(CreateUpdateModel):
 
     A template for an Entry.
     """
+    class Meta:
+        constraints = [
+            CheckConstraint(check=~Q(name=''), name='non_empty_name'),
+        ]
+        # NOTE: First requires solution for live data
```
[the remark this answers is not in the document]
mjvkeulen: What would you like to see? Issue?

src/django/VLE/utils/import_utils.py
mjvkeulen: Lets move the discussion here: https://zube.io/ejourn-al/ejournal/c/1486

src/vue/src/components/template/TemplateImport.vue
mjvkeulen: Think we have this issue (if that covers it) https://zube.io/ejourn-al/ejournal/c/1466

src/vue/src/components/assignment_editor/AssignmentEditorBreadCrumb.vue
mjvkeulen: I prefer compartmentalizing logic where possible, this is also how I started the `FormatEdit.vue` refactor. At the start `FormatEdit.vue` itself was 600 lines, including intro text scattered throughout even duplicated for the breadcrumb (as it appears twice). With the old breadcrumb taking up 15ish lines twice, whereas the current `AssignmentEditor.vue` template is 55 lines total.

This explains my initial approach, but now that the dust has settled does the assignment breadcrumb component hamper clarity / performs a false abstraction?

I don't believe so, at least not for the arguments u make here. What the assignment editor looks like functionally, is helped by the abstraction. You don't care about how the breadcrumb is structured / what styling is needed when looking at the assignment editor top down.

Later on, you make a valid point about a tour step being hidden, as it is now placed in a nested component. I believe this does show that the abstraction is leaky. Since we can't have multiple tour steps on the same component, I will flatten the structure to address this issue.

src/vue/src/views/AssignmentEditor.vue
```diff
@@ -1,51 +1,57 @@
 <template>
-    <load-wrapper :loading="loading">
```
mjvkeulen: @engelhamer is this not desired behaviour? Wrapping all of the content in a single load wrapper prevents the need to sprinkle additional guards throughout the template. Improves readability, making the logic slightly simpler to follow?

keep_testing.sh
mjvkeulen: Are you sure? Preset DB requires an updated faker version

src/django/VLE/views/assignment.py
mjvkeulen: https://zube.io/ejourn-al/ejournal/c/1486

src/vue/src/components/assignment_editor/timeline_controlled/SelectedTimelineComponentSwitch.vue
mjvkeulen: yes

src/vue/src/components/assignment_editor/timeline_controlled/assignment/AssignmentEndReadMode.vue
mjvkeulen: In read mode, you can see the end dates, e.g. lock date. Clicking edit will give you the opportunity to edit this.

src/vue/src/store/modules/assignmentEditor.js
mjvkeulen: I don't know, I think this was extracted during the rework, template can prob be null

mjvkeulen: image

Makes use of `''` to show the default text, and `null` to show the nonselectable option.

src/vue/src/store/modules/assignment.js
mjvkeulen: Would you prefer:
`line-of-logic` // eslint-disable-line object-curly-newline, max-len
`line-of-logic` // eslint-disable-line

Keep in mind line of logic can be lengthy

src/vue/src/store/modules/assignment.js
mjvkeulen: All of the noncached functions can drop the `fn` wrapper. 

However, adding a cache, for example to prevent multiple clicks triggering multiple updates, requires a diff of two lines atm.

src/vue/src/router/index.js
mjvkeulen: The list of filtered categories transcends the assignment editor (filter active in any other view, e.g. journal)

src/vue/src/helpers/global_mixins.js
mjvkeulen: Imo an "invalid" check works well with the nature of a for `some` loop.

src/vue/src/views/AssignmentEditor.vue
```diff
@@ -0,0 +1,245 @@
...
+New deadlines can be added via the '+' button in the timeline. Click any deadline to view its contents.
+`,
+            categoryIntroText: `
+<i>Categories</i> can be linked to entries and be used to filter the timeline.<br/><br/>
+
+You can choose to link specific templates to categories. When a student makes use of these templates
+to create an entry, the category will be linked to the entry by default.<br/><br/>
+
+Whether the student can edit which categories belong to an entry themselves, can be configured via the
+respective template setting "<i>Fixed Categories / Custom Categories</i>".
```
Larspolo, line 136:
```
It is also possible to allow students to add other existing categories to an entry. This can be turned on via the respective template setting "<i>Fixed Categories / Custom Categories</i>".
```

src/vue/src/views/AssignmentEditor.vue
```diff
@@ -0,0 +1,245 @@
...
+                }
+            })
+    },
+}
+</script>
+
+<style lang="sass">
+.dirty::after
+    content: " *"
+    color: $theme-orange
```
Larspolo, line 217: image
Dirty indicator is not for new stuff. Could be added / create an issue for it

mjvkeulen: Would be quite simple to add if you would like, not sure how to visually display it tho. E.g. what should the "dirty" state of the timeline add node look like?

src/vue/src/components/template/TemplateMenuItem.vue
```diff
@@ -0,0 +1,62 @@
...
+
+        <icon
+            name="trash"
+            class="trash-icon ml-2 float-right"
+            @click.native.stop="$emit('delete-template', template)"
+        />
+
+        <span
+            class="max-one-line"
+            :class="{ dirty: isTemplateDirty(template) }"
```
Larspolo, line 29: There is a bug with this indicator:
1. Click on template 1
2. Click edit,but do not edit anything
3. Click on template 2
4. Click edit
5. Edit sth and save
template 1 now has a dirty indicator, while it is not dirty

src/vue/src/views/AssignmentEditor.vue
mjvkeulen: The new <X> undefined looks like it needs a fix tho

src/vue/src/components/template/TemplateMenuItem.vue
mjvkeulen: Can not reproduce? :/

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
Larspolo: > Could you confirm this is resolved?

Can confirm the problem is solved, however, a new (I think closely related) problem occures now:

1. Edit template name that contains at least 1 category (also save)
2. Edit category name that is related to that template (also save)
3. View that the template now becomes dirty, this should not be the case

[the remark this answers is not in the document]
Larspolo: Select template -> Edit -> Add 1 to name -> Save -> Select related category -> Edit -> Add 1 to name -> Save -> Observe template becomes dirty

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEditSelectAndPreviewTemplate.vue
```diff
@@ -0,0 +1,101 @@
...
+                    <b-button
+                        :class="{
+                            'red-button': showTemplatePreview,
+                            'green-button': !showTemplatePreview,
+                        }"
+                        :disabled="!presetNode.template"
+                        @click="showTemplatePreview = !showTemplatePreview"
+                    >
+                        <icon :name="(showTemplatePreview) ? 'eye-slash' : 'eye'"/>
+                        {{ (showTemplatePreview) ? 'Hide Preview' : 'Show Preview' }}
```
Larspolo, line 41: Preview button can now be (re)moved as it shows more then just the template. Maybe like the template creation a preview btn top right? Otherwise a removal is also fine imo

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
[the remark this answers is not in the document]
Larspolo: Managed to reproduce it once, but can no longer, small enough issue RN for me to leave it be

src/vue/src/store/modules/assignment.js
Larspolo: First one, if its too long, you can do `// eslint-disable-next-line object-curly-newline, max-len`

src/vue/src/components/category/CategoryEdit.vue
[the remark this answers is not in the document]
mjvkeulen: > Edit template name that contains at least 1 category (also save)
Edit category name that is related to that template (also save)
View that the template now becomes dirty, this should not be the case

Could not reproduce this following the given steps.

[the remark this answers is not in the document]
mjvkeulen: > Select template X -> Edit -> Add 1 to name -> Save -> Select related category -> Edit -> Add 1 to name -> Save -> Observe template X becomes dirty

This happens only if an entry deadline exists which makes use of the template. Not sure why, will look into this.

src/vue/src/views/AssignmentEditor.vue
engelhamer: I think that this is desired behaviour: the load wrapper itself is a transition which fades-out the content on changing conditions. This should be used for portions of the app that require longer loading than others, but is not desired for whole router-pushed components because of the bug we saw.

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: `Deadline has a lock date before the assignment due date`

src/django/VLE/models.py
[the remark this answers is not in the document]
engelhamer: If you are not going to resolve this now an issue would be nice right?

src/vue/src/components/template/TemplateImport.vue
engelhamer: I think it can be expanded to also support copying template categories then 👍

engelhamer:
image

Creating a new LTI assignment fails due to a date error.

mjvkeulen:
> Creating a new LTI assignment fails due to a date error.

Fixed in fac353c9dd713cdbbe69feddd97e836ba25e19f5 (showing a wider problem, also resulting in ea1ad1bee3d2cb7fd5bfe5b34bd1edede361993e)

engelhamer:
image

Clicking 'add deadline' gives a back end error rather than the validation label. Only when the 'due date' input is clicked the validation line is shown, and clicking 'add deadline' then does give the right user-friendly error.

image

src/vue/src/components/template/TemplateMenuItem.vue
engelhamer: I can't reproduce either (chrome)

engelhamer:
image

There are no placeholders for an empty description or template list in the category display

Larspolo (approved):

src/vue/src/views/AssignmentEditor.vue
```diff
@@ -213,8 +213,14 @@ This tutorial can be consulted again by clicking the <i>info</i> sign.
 
 <style lang="sass">
 .dirty::after
-    content: " *"
+    content: '\2022'
```
mjvkeulen, line 216: I like the dot!

src/vue/src/views/AssignmentEditor.vue
```diff
@@ -213,8 +213,14 @@ This tutorial can be consulted again by clicking the <i>info</i> sign.
...
+    font-size: 2em
+    line-height: 10px
+    vertical-align: middle
+    margin-left: 4px
+    margin-right: 4px
     color: $theme-orange
+    text-shadow: 0 0 6px rgba(232,167,35,0.5), 0 0 6px rgba(232,167,35,0.8)
```
mjvkeulen, line 223: Q) Are we sure about the shadow?

src/vue/src/components/assignment/AssignmentDetails.vue
```diff
@@ -17,19 +17,17 @@
                         :state="nameInputState"
                     />
 
-                    <template #append>
```
mjvkeulen: It seems to render fine without, but is this not the desired slot?

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,75 +1,76 @@
...
+            class="float-right mb-3 ml-3"
+        />
+        <h2 class="theme-h2 field-heading multi-form">
+            Unlimited use
+        </h2>
+        Allow students to use this template for new entries in their journal.
+        When disabled, this template can still be used for deadlines you add to the timeline.
+        <template v-if="assignmentHasCategories">
+            <hr/>
+            <!-- TODO CATEGORY: Invert the notion of 'fixed categories' in the back end? This seems more natural. -->
```
mjvkeulen, line 26: Should be simple as the term is unique, what is a clear term for free form categories?

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,75 +1,76 @@
...
+            <h2 class="theme-h2 field-heading multi-form">
+                Allow custom categories
+            </h2>
+            Allow students to choose which categories they add to their entry when using this template.
+            <hr/>
+            <h2 class="theme-h2 field-heading">
+                Default categories
+            </h2>
+            Select which categories are added to entries using this template by default.
+            <category-select
```
mjvkeulen, line 52: Can't wait till we flatten some of our UI

image

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,75 +1,76 @@
...
-                        Students can not change the default categories for entries created with this template
-                    </span>
-                    <span v-else>
-                        Students can freely add or remove categories to entries created with this template
-                    </span>
-                </b-td>
-            </b-tr>
-        </b-tbody>
-    </b-table-simple>
+    <b-card class="no-hover">
```
mjvkeulen, line 2: Q) I thought we were moving away from the card in card layouts?

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,75 +1,76 @@
...
+                    class: 'red-button',
+                },
+            ]"
+            class="float-right mb-3 ml-3"
+        />
+        <h2 class="theme-h2 field-heading multi-form">
+            Unlimited use
+        </h2>
+        Allow students to use this template for new entries in their journal.
+        When disabled, this template can still be used for deadlines you add to the timeline.
```
mjvkeulen: Definitely looks better than the previous layout. 

We should consolidate our text with options menus into a reusable component. For example, it makes sense if the profile page uses the same component where applicable for a more uniform style:

image

image

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,75 +1,76 @@
...
-                    />
-                    Information
-                </b-th>
-            </b-tr>
-        </b-thead>
-
-        <b-tbody>
-            <b-tr>
-                <b-td class="template-availability">
-                    <b-button
```
mjvkeulen: Q) When do we prefer toggle buttons (True and False both visible), over a button which shows the current state:

image

The template edit menu now holds both.

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEditSelectAndPreviewTemplate.vue
```diff
@@ -0,0 +1,101 @@
...
+                </template>
+            </b-input-group>
+        </b-form-group>
+
+        <div class="multi-form">
+            <span
+                class="text-blue cursor-pointer"
+                @click.stop="createTemplate()"
+            >
+                {{ (templates.length === 0) ? 'Create a new template' : 'Or create a new template' }}
```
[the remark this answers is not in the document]
mjvkeulen: So currently we have:

1. + Create New Template
2. Enter desired template configuration
3. + Add Template
4. Template is saved and shown in read mode.

You would prefer to forgo the read mode, and take the user back to the state after step 1?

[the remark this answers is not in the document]
engelhamer: I think yes, because they created the template in that context. There is also a preview button there so the view mode can be easily achieved

src/vue/src/views/AssignmentEditor.vue
engelhamer: I think it attracts the attention a bit more like this

src/vue/src/components/assignment/AssignmentDetails.vue
engelhamer: I think that the slot makes sense for e.g. email verification (enter email, then press verify button in slot to verify it). However, the publish toggle of the assignment is not an action that uses the input field (name) here.

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: `allow_custom_categories`?

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: In the next update, but for now I like the consistency between them.

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: This screenshot is a toggle button (only see the active state, click to toggle it). I think this is useful for when the button contains text, because it otherwise becomes too large. If there are multiple  options for which icons only would suffice (so true/false) we can show both.

src/vue/src/components/template/TemplateEditSettings.vue
mjvkeulen: I understand the separation, but I am not sure it always works:

image

The card (for one field especially) breaks the uniformity of the editor. A simple solution would be to simply use form-groups with labels, and to remove the card:

image

I feel this is more consistent

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEditSelectAndPreviewTemplate.vue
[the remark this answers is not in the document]
mjvkeulen: I misunderstood the question, but this commit introduces what you are after: https://github.com/eJourn-al/eJournal/pull/1081/commits/357f1ae698ce5245a349372ee4d1431031357e73

The user is returned to editing the deadline if he chose to create a template during that workflow.

src/django/VLE/models.py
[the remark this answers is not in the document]
mjvkeulen: Z#1489

src/django/test/factory/assignment.py
mjvkeulen: [NFN] Not used atm, and your proposal would always execute a query (even when the id is provided).

src/vue/src/components/template/TemplateEditSettings.vue
mjvkeulen: Could work, fixed_categories is enabled by default. Might be an indication the name is fine for how we use it?

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: Fine by me! More consistent indeed. We can update that for the other instances in the UI branch but change this one now

engelhamer: Only thing that looks odd to me is the placement of the button below the label, for ibputs that makes sense but a right floating element not so much in my opinion

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: We also have e.g. `allow_custom_journal_name` and `allow_custom_journal_image` both off by default right?

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -27,20 +27,37 @@
...
-                <b-form-input
-                    v-model="category.name"
-                    autofocus
-                    placeholder="Name"
-                    class="theme-input"
-                    type="text"
-                    trim
-                />
-            </b-form-group>
+            <b-form-row class="mb-2">
```
mjvkeulen, line 30: NIT, work with multi-form or better define rule if not last child default multi-form spacing?

src/vue/src/components/category/CategoryEdit.vue
```diff
@@ -27,20 +27,37 @@
...
+                            v-model="category.name"
+                            autofocus
+                            placeholder="Name"
+                            class="theme-input"
+                            type="text"
+                            trim
+                        />
+                    </b-form-group>
+                </b-col>
+                <b-col cols="auto">
```
mjvkeulen, line 48: Q) Is column separation required if the input is of fixed width?

src/vue/src/components/category/CategoryEdit.vue
engelhamer: `multi-form` in effect is identical to `mb-2` and the latter is a default bootstrap class. In the UI branch I removed all so Ingot used to this.

src/vue/src/components/category/CategoryEdit.vue
engelhamer: Yes, since the default behaviour is equal division of width

src/vue/src/components/category/CategoryEdit.vue
mjvkeulen: Moved it to a not last-child rule

src/vue/src/components/template/TemplateEditSettings.vue
mjvkeulen: I stashed this solution as I don't like how the floating buttons interact with the next input row on wide devices (the buttons are twice the height of our text, adding additional margin-bottom when the text fits on a single line)

PR #1083, opened 2021-02-04
Added scroll controls to timeline category display

engelhamer:
timeline-category-scroll


src/vue/src/components/category/CategoryDisplay.vue
```diff
@@ -62,22 +76,50 @@ export default {
     computed: {
         infoModalID () { return `${this.id}-category-information` },
     },
+    methods: {
+        scrollLeft () {
+            this.$refs[`${this.id}-category-display`].scrollBy({
+                left: -this.$refs[`${this.id}-category-display`].clientWidth,
```
mjvkeulen, line 82: [NIT]: Combine methods and parametrize left or right as + or -

mjvkeulen (approved):

src/vue/src/components/category/CategoryDisplay.vue
```diff
@@ -18,6 +26,12 @@
                 $nextTick(() => { $bvModal.show(infoModalID) })
             "
         />
+        <icon
+            v-if="categories.length > 0"
+            class="category-right"
+            name="caret-right"
+            @click.native="scrollRight"
+        />
```
Larspolo, line 34: Arrows are still visible even when there is nothing to scroll through, would like to see them hidden if there is no overflow

engelhamer: Most naive solution would require `$refs` in computed prop, which is not desirable. Defering to issue for now.

src/vue/src/components/category/CategoryDisplay.vue
engelhamer: I like the explicit names a slight bit better (without a wrapper or sign)

PR #1084, opened 2021-02-05
Category colours based on theme colours

Larspolo:
## Description

First 9 are now set, after that it is random generated based on lighness levels between 0.2 and 0.5 (was 0.4 and 0.8, but I ofound that too bright)
image


src/vue/src/utils/colors.js
```diff
@@ -88,8 +88,24 @@ export default {
     },
 
     randomBrightRGBcolor () {
-        const hsl = this.randomBrightHSLColor()
+        const hsl = this.randomBrightHSLcolor()
         const rgb = this.hslToRgb(hsl.hue * 360, hsl.saturation, hsl.lightness)
         return this.rgbToHex(rgb.red, rgb.green, rgb.blue)
     },
+
+    themeColors: [
```
mjvkeulen, line 96: This should probably come from a source that is also accessible by the backend, so we can make use of this in factory/category.py and presetdb

Larspolo: Fine to move this to an issue?

Larspolo:
image

Cross is barely to not visible for red (ish) coloured categories

Larspolo:
image
Categories without description have a different size

src/vue/src/utils/colors.js
engelhamer: Now using the sass colormap of `themeColors` (which had the same name, so was quite confusing anyways) without the gray shades.

mjvkeulen: Moved my remark to Z#1542

mjvkeulen (approved):

PR #1094, opened 2021-02-10
Fixed issue where new entry notifications were not send

Larspolo:
## Description
Fixed issue where new entry notifications were not send

## Summary of changes
- Fixes S#1896

### DIFF
--- a/src/django/VLE/factory.py
+++ b/src/django/VLE/factory.py
@@ -225,7 +225,7 @@ def make_entry(template, author, node, category_ids=None):
     params = {}
     if not template.chain.allow_custom_categories or category_ids is None:
         params['category_ids'] = list(template.categories.values_list('pk', flat=True))
-    if category_ids:
+    elif category_ids:
         params['category_ids'] = category_ids
 
     # NOTE: because we use a custom variable (category_ids) we have to call the save function seperatly
--- a/src/django/VLE/factory.py
+++ b/src/django/VLE/factory.py
@@ -222,20 +222,15 @@ def make_node(journal, entry=None, type=VLE.models.Node.ENTRY, preset=None):
 
 
 def make_entry(template, author, node, category_ids=None):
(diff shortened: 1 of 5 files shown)


src/django/VLE/factory.py
```diff
@@ -222,20 +222,15 @@ def make_node(journal, entry=None, type=VLE.models.Node.ENTRY, preset=None):
...
-            VLE.models.EntryCategoryLink(entry=entry, category_id=id, author=author)
-            for id in category_ids
-        ]
-        VLE.models.EntryCategoryLink.objects.bulk_create(entry_category_links)
-
-        entry.node.entry = entry
-        entry.node.save()
+        params['category_ids'] = list(template.categories.values_list('pk', flat=True))
+    elif category_ids:
+        params['category_ids'] = category_ids
```
Larspolo, line 229: @mjvkeulen how is this line not hit by existing tests? Did you not test this? Or did I fuck up the conversion?

src/django/VLE/factory.py
```diff
@@ -222,20 +222,15 @@ def make_node(journal, entry=None, type=VLE.models.Node.ENTRY, preset=None):
 
 
 def make_entry(template, author, node, category_ids=None):
+    params = {}
     if not template.chain.allow_custom_categories or category_ids is None:
-        category_ids = list(template.categories.values_list('pk', flat=True))
-
-    with transaction.atomic():
```
mjvkeulen, line 228: I would be in favor of the following solution:

Remove `generate_new_entry_notifications` call from the save of `Entry`.
Call `transaction.on_commit(lambda: generate_new_entry_notifications.delay(entry.pk, node.pk))` in this block.

I feel the solution is more elegant and requires less of a rewrite. Do you see any paths where this solution will no longer generate notifications? 
Also one day we will move these factory functions to the model manager create function.

PR #1095, opened 2021-02-10
Add more user friendly entry creation validation messages

mjvkeulen:


src/django/VLE/utils/entry_utils.py
```diff
@@ -42,16 +42,16 @@ def check_fields(template, content_dict):
 
 def add_entry_to_node(node, template, author, category_ids=None):
     if not (node.preset and node.preset.forced_template == template):
-        raise VLEBadRequest('Invalid template for preset node.')
+        raise VLEBadRequest('Your educator has updated the template for this entry, please reload and try again.')
```
Larspolo, line 45: [NIT] we could check if the update date has changed recentrly (e.g. within the last hour) and if so, display this message, else the original message?

Larspolo (approved):

src/django/VLE/utils/entry_utils.py
mjvkeulen: I am not sure what the time window should be, I went for a different approach.

src/django/VLE/utils/entry_utils.py
```diff
@@ -40,18 +40,25 @@ def check_fields(template, content_dict):
...
-def add_entry_to_node(node, template, author, category_ids=None):
+def add_entry_to_deadline_preset_node(node, template, author, category_ids=None):
     if not (node.preset and node.preset.forced_template == template):
-        raise VLEBadRequest('Invalid template for preset node.')
+        if (
+            node.preset
+            and node.preset.forced_template.chain == template.chain
+            and template.archived
+        ):
+            raise VLEBadRequest('Your educator has updated the template for this entry, please reload and try again.')
```
engelhamer, line 50: [NIT] Educator not used elsewhere?

engelhamer (approved):

PR #1097, opened 2021-02-13
More intuitive default assignment format

mjvkeulen:


Larspolo (approved):

src/vue/src/components/template/TemplateField.vue
```diff
@@ -14,7 +14,7 @@
                 <b-input
                     v-model="field.title"
                     class="multi-form theme-input"
-                    placeholder="Field title"
+                    placeholder="Optional field title"
```
engelhamer, line 17: There is no 'required' asterisk here, implying this is optional already. We also don't put 'required template title'?

src/django/VLE/factory.py
```diff
@@ -187,29 +180,20 @@ def make_lti_groups(course):
             VLE.models.Group.objects.filter(course=course, lti_id=lti_id).update(name=name)
 
 
-def setup_default_assignment_layout(assignment, due_date, points_possible):
+def setup_default_assignment_layout(assignment):
     template = VLE.models.Template.objects.create(
-        name='Entry',
+        name='Template',
```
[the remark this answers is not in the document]
engelhamer: I liked the name Entry better because it also appeared in the timeline that way, but if the other PR is approved this is fine.

src/vue/src/components/template/TemplateField.vue
engelhamer: Same for optional description: also no indicator in the placeholder of it being optional (which I think is fine). This is a bit inconsistent.

src/django/VLE/factory.py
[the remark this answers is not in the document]
mjvkeulen: I renamed it to `Template` to better distinguish that the template name is just a label for the template itself. This is to further move away from the one template for each deadline setup we see so often.

src/vue/src/components/template/TemplateField.vue
mjvkeulen: Added it to the description as well. Since there are no field labels, it is not directly clear that the fields can be optional. 

Description is more intuitive to be an optional field than 'title', I think this is why we see users skip the description but not the title. For example for a template which only holds a single RT field to serve as the entry content.

engelhamer (approved):

PR #1101, opened 2021-02-23
Added the ability for students to set a custom entry title

mjvkeulen:
image
image
image
image

Continues on #1093 as it makes use of some shared logic.


Larspolo:
Given the description I didn't expect the title to be required for students. Imo students can override it, but shouldn't have to

Also nice and quick PR!

src/django/VLE/utils/entry_utils.py
```diff
@@ -53,7 +53,7 @@ def add_entry_to_node(node, template, author, category_ids=None):
     if node.preset.is_locked():
         raise VLEBadRequest('The lock date for this node has passed.')
 
-    entry = factory.make_entry(template, author, node, category_ids)
+    entry = factory.make_entry(template, author, node, category_ids, title)
```
Larspolo, line 56: Title on presetnodes does not get set to the jnput value during creation. Only update

Would expect tests to automatically test this in the futute

src/vue/src/components/entry/Entry.vue
```diff
@@ -190,8 +210,10 @@ export default {
             handler () {
                 if (this.node && this.node.entry) {
                     this.newEntryContent = { ...this.node.entry.content }
+                    this.title = this.node.entry.title
```
Larspolo, line 213: Presetnodes with filled entry with custom title still display the presetnode name in timeline. This should be the custom title (right?). It does get displayed correctly in the node title

Larspolo (changes requested):

src/django/VLE/utils/entry_utils.py
mjvkeulen: I don't follow. 

Workflow where student creates entry with custom title works as expected? (Same for patch)

src/vue/src/components/entry/Entry.vue
mjvkeulen: I didn't pass the title to a deadline entry during creation, should be fixed.

src/vue/src/components/entry/Entry.vue
```diff
@@ -42,6 +42,23 @@
...
             />
+
+            <b-form-group
+                v-if="node && template.allow_custom_title && (edit || create)"
+                label="Title"
+                class="required"
+            >
+                <template #label>
+                    Title
+                    <tooltip tip="Serves as the title for this entry as well as the display name in the timeline."/>
```
engelhamer, line 53: I would prefer to remove the 'display name' ambiguity here. Rather just 'how this entry appears in the timeline' maybe? We could also name it 'title' for the preset node for teachers, to have fewer terms to understand and make it a bit easier?

What do you think about this?

mjvkeulen: It dictates both the entry title (header of the entry content) as the label used in the timeline. 

I believe 'how this entry appears in the timeline' does not sufficiently reflect this. 

Terminology-wise, do you feel using title everywhere to represent both the timeline label and entry title is sufficiently obvious? I would be fine merging display name and title into some term (possibly title), since atm both (preset node display name and entry title) always dictate both the entry title as the timeline label. Should the chosen term be title, or can we find something which more obviously reflects both use cases?

engelhamer: Title sounds fine! I think it is sufficiently clear 👍

Larspolo:
image

I liked the tooltip, why is it now removed?

Larspolo (changes requested):
Title tooltip would be nice, rest is fine as is

Larspolo (approved):

src/vue/src/components/entry/EntryPreview.vue
```diff
@@ -19,6 +19,17 @@
             <files-list :files="presetNode.attached_files"/>
         </template>
 
+        <b-form-group
+            v-if="template.allow_custom_title"
+            label="Title"
```
engelhamer, line 24: Should have required asterisk?

engelhamer (approved):

PR #1103, opened 2021-02-27
Reduced query count for serializing permissions

Larspolo:
## Description
Found out that every call in the test script first logs in, this caused a lot of pointless queries, among others the permission serializer which is now a bit better, not good (it still loops over all connected courses  & assignments), but better


mjvkeulen (commented):
Could you add a test which shows an upper limit on the scaling of the queries? E.g. 4 queries for each user or smth?

src/django/VLE/permissions.py
```diff
@@ -137,8 +137,26 @@ def serialize_general_permissions(user):
...
+    permissions = {}
+    for course in assignment.courses.all():
+        perms = VLE.models.Role.objects.values(*VLE.models.Role.ASSIGNMENT_PERMISSIONS).filter(
+            role__user=user,
+            course=course,
+        ).first()
+
+        if perms:
+            for perm, value in perms.items():
+                if value or value not in permissions:
```
engelhamer, line 156: How does this work? I take it that `value` is the boolean stating whether the permission holds, but why is the permission added with value `True` if `value` is not in `permissions`?

Larspolo: Good catch. Its was wrong indeed. Added test to prevent this in the future. I think I just hoped there would be a test for this already.

mjvkeulen (approved):

engelhamer (approved):

PR #1107, opened 2021-03-04
Fixed an issue where file uploads could falsely trigger a pending changes indicator

mjvkeulen:
Fixes Z#1677


engelhamer (approved):

Larspolo (approved):
[NIT] would another (more globally effective) solution be to always remove it for file upload successes? Why keep the code_version and description anyway?

mjvkeulen:
> [NIT] would another (more globally effective) solution be to always remove it for file upload successes? Why keep the code_version and description anyway?

I moved the FileSerializer data behind the key 'file' on file creation

PR #1109, opened 2021-03-10
Fixed an issue which prevented entries from being posted

mjvkeulen:
Frontend still provided a title even when the template setting did not allow as such.

NOTE: A bandaid fix is already live (but overwritten on any deployment).


Larspolo (approved):
Works

engelhamer (approved):

PR #1111, opened 2021-03-11
Fixed an issue where a pending changes indicator could persist upon saving after a deadline date was removed

mjvkeulen:
Another issue where comparison of null and an empty string ('') caused an error

- Fixes Z#1687


src/vue/package.json
```diff
@@ -24,6 +24,7 @@
     "file-saver": "^2.0.2",
     "intro.js": "^2.9.3",
     "lodash.isequal": "^4.5.0",
```
Larspolo, line 26: Is this still needed?

Larspolo (approved):
Looks good, I think this is valid to also do frontend because we have it backend as well. Did not test all of the assignment editor though, only some parts that also use it

engelhamer (approved):

PR #1112, opened 2021-03-12
Django & httplib version bump

Larspolo:
## Description
Django & httplib version bump


mjvkeulen (approved):

requirements/base.txt
```diff
@@ -21,7 +21,7 @@ python-dateutil==2.8.0 # TODO Unsure of its dependants
 
 certifi==2019.9.11 # TODO figure out specific use cases
 chardet==3.0.4 # TODO figure out specific use cases
-httplib2==0.14.0 # TODO figure out specific use cases
+httplib2==0.19.0 # TODO figure out specific use cases
```
engelhamer, line 24: You bumped this, but still don't know where it is used?

engelhamer (approved):

requirements/base.txt
Larspolo: Yup. Bumped because of the security warnings

PR #1113, opened 2021-03-12
Added preview of group journal student names when joining

mjvkeulen:
- Fixes Z#1572


src/django/VLE/views/journal.py
```diff
@@ -281,9 +281,13 @@ def join(self, request, pk):
...
     def get_members(self, request, pk):
         """Get the list of members of the journal."""
-        journal = Journal.objects.get(pk=pk)
+        journal = Journal.objects.filter(pk=pk).select_related('assignment').get()
 
-        request.user.check_can_view(journal)
+        can_view_journal = request.user.can_view(journal)
+        is_assignment_participant = request.user.is_participant(journal.assignment)
+
+        if not (can_view_journal or is_assignment_participant):
```
engelhamer, line 289: This allows any member of a course (even without the `can_view_all_journals` permission to obtain the names of all people in the course (via enumerating journal `pk`s). I don't really like that change, and instead we should probably check whether this is a group assignment (only if that is the case we may allow this, but just for those users who can have a journal themselves or can view all journals).

engelhamer (changes requested):

src/django/VLE/views/journal.py
mjvkeulen: I don't believe that being able to see the names of other students in your assignment as a student is too forthcoming with information.

A reasonable example of this in practice is Canvas, where once given access to a course you can navigate to "people" and see a list of the course members.

engelhamer: Yes, but that holds only for students. There are also examples of course designers which should not be able to see students at all.

There is no benefit or other reason to change this functionality right? Why not keep the more secure state?

mjvkeulen: > Why not keep the more secure state?

We need to be less restrictive to allow students to see the members of a journal of which they are NOT (yet) a part of. 
can_view(journal) governs access to more than simply the current journal member's names.

We cannot tie this permission to `can_have_journal` as this API is used by both students and educators.

engelhamer: Indeed we need to be less restrictive, but _only for group assignments_. Hence I stated in my initial comment:

> instead we should probably check whether this is a group assignment (only if that is the case we may allow this, but just for those users who can have a journal themselves or can view all journals)

src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -82,7 +82,6 @@
         </b-row>
         <slot/>
         <div
-            v-if="assignment.is_group_assignment && $hasPermission('can_manage_journals')"
```
Larspolo, line 85: I dont think this should be removed. At least it needs to stay a group assignment

Larspolo: image

Example of a non group assignment

src/vue/src/components/journal/JournalMembers.vue
```diff
@@ -139,12 +139,18 @@ export default {
                 + 'will reset the journal and remove all its entries. This action cannot be undone.' : ''}`)) {
                 journalAPI.leave(this.journal.id)
                     .then(() => {
-                        this.$router.push({
-                            name: 'JoinJournal',
-                            params: {
-                                cID: this.assignment.course.id, aID: this.assignment.id,
-                            },
-                        })
+                        if (this.$route.name === 'Journal') {
```
Larspolo, line 142: [Q] when is this triggered?

Larspolo (changes requested):

src/vue/src/components/journal/JournalMembers.vue
engelhamer: When quitting a journal as a student, right?

engelhamer (approved):

src/vue/src/components/assignment/JournalCard.vue
engelhamer: Indeed has to be resolved

src/vue/src/components/journal/JournalMembers.vue
Larspolo: *when is this NOT triggered

mjvkeulen: never

Larspolo (approved):

PR #1117, opened 2021-03-19
Refactored timeline UI selection to pass nodes instead of indices

mjvkeulen:
Pave the way for future timeline related issues such as

- Z#1561
- Z#1238


src/vue/src/components/journal/JournalStudent.vue
```diff
@@ -192,12 +188,18 @@ export default {
         }
     },
     computed: {
-        addIndex () {
-            return this.nodes.findIndex((node) => node.type === 'a')
+        ...mapGetters({
+            currentNode: 'timeline/currentNode',
+            startNode: 'timeline/startNode',
+            addNodeSymbol: 'timeline/addNodeSymbol', // Add node is serialized by the backend
```
engelhamer, line 194: Shall we add it front end like the start and end nodes? Or maybe only use the FAB anymore?

engelhamer: This could be a good moment to include it in the refactor anyways?

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -28,9 +28,37 @@
...
+        },
+    },
     computed: {
+        ...mapGetters({
+            currentNode: 'timeline/currentNode',
+            endNode: 'timeline/endNode',
+            assignment: 'assignment/assignment',
+        }),
+        selected () {
+            const idKey = (this.workingWithPresetNodes) ? 'id' : 'nID'
```
engelhamer, line 51: Why do we keep this distinction between `id` and `nID` anyways? Maybe it is worth changing that instead of using the `workingWithPresetNodes` prop?

src/vue/src/components/journal/JournalStudent.vue
mjvkeulen: We could move to only using the FAB, this would take out the logic we have for the positioning of the add node. 

The add node is added backend, I believe this was for 2 reasons:
- Ensure the add node has access to the relevant templates. We now have access to the assignment/template store anyway, this would be simple to switch.
- The positioning of the add node is handled (and tested) backend. This is valid, especially due to the tests, but if we just move over to the FAB, this logic is no longer required.

@Larspolo opinion?

src/vue/src/components/timeline/TimelineNodeCircle.vue
mjvkeulen: Sure, probably keep `nID` as query param?

src/vue/src/components/journal/JournalStudent.vue
mjvkeulen: What would do with the assignment editor? Having just a FAB there might be ambiguous?

Larspolo: > Or maybe only use the FAB anymore?

I like it in the timeline, created the feeling of adding something to the timeline iso just submitting another thing. Same behavior for submitting unlimited & deadline entries

> The positioning of the add node is handled (and tested) backend.

Some frontend tests for this would sure be nice, as testing this manually is annoying

src/vue/src/components/timeline/TimelineNodeCircle.vue
Larspolo: Yeah, queryparam should be kept for also LTI. At least keep supporting nID as current submissions will still use that param.

I also tested if LTI still works, and it does :ok_hand:

src/vue/src/components/entry/Entry.vue
```diff
@@ -245,22 +245,24 @@ export default {
...
                         this.edit = false
-                        this.requestInFlight = false
-                    })
-                    .catch(() => {
-                        this.requestInFlight = false
                     })
+                    .finally(() => { this.requestInFlight = false })
             }
         },
+        clearDraft () {
```
Larspolo, line 252: Cancle button does not clear draft atm, this means that the edited content still persists, when it shouldnt
1.
image
2.
image
3.
Click Cancel and see (BUG):
image
4.
navigate forth and back to another node does not give a warning:
image

src/vue/src/components/entry/Entry.vue
```diff
@@ -299,8 +301,10 @@ export default {
         safeToLeave () {
             // It is safe to leave the entry if it is not currently being edited AND if no content for an entry to be
             // created is provided.
```
Larspolo, line 303: Title should be taken into account in the content. It currently isn't (neither for current deploy).

Either fix this now, or we should create an issue for it. Probably same for categories

src/vue/src/components/entry/EntryNonStudent.vue
```diff
@@ -271,7 +271,10 @@ export default {
                     },
                     { customSuccessToast },
                 )
-                    .then(() => {
+                    .then((entry) => {
+                        this.grade.grade = entry.grade.grade
+                        this.grade.published = entry.grade.published
```
Larspolo, line 276: [NIT] Would `this.grade = entry.grade` also work?

src/vue/src/components/journal/JournalStudent.vue
Larspolo: Relevant issue:
https://zube.io/ejourn-al/ejournal/c/1140

src/vue/src/store/modules/assignmentEditor.js
```diff
@@ -496,8 +488,6 @@ export default {
         },
         activeComponentMode: readSymbol,
 
-        selectedTimelineElementIndex: -1, /* See timeline.vue for mapping (due for refactor). */
```
Larspolo, line 499: Nice :)

Larspolo (changes requested):

src/vue/src/components/entry/Entry.vue
mjvkeulen: We have an outstanding issue for this already

src/vue/src/components/entry/Entry.vue
mjvkeulen: I believe we have an [issue](https://zube.io/ejourn-al/ejournal/c/1759) for the lacking navigation guard.

mjvkeulen: The nonJournal naviagation guard (safeToLeave), is addressed in this PR, where previously the pending grade was not reset on component destroy but via a watcher on node change.

The clear draft guard logic is missing checks, but that is the status quo nor the point of this PR.

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -48,17 +44,21 @@ export default {
...
-                    && this.node[idKey] === this.currentNode[idKey]
+                    && this.node.id && this.currentNode.id
+                    && this.node.id === this.currentNode.id
                 )
             )
         },
+        /* Boolean used to indicate the assignment is being edited, new preset nodes can be inserted
+         * which will not yet be saved / have an id.
+         * Entries are created one at a time and are always inserted after save (with id) */
+        workingWithPresetNodes () {
```
engelhamer, line 59: This property is not used anywhere in this component, right? Why do we need to add it here?

src/vue/src/components/entry/Entry.vue
engelhamer: Then make an issue for the cancel button @Larspolo?

src/vue/src/components/timeline/TimelineNodeCircle.vue
mjvkeulen: Left over, will remove

mjvkeulen: I think you are on old commit or smth?

src/vue/src/components/entry/Entry.vue
Larspolo: I think thats introduced in this PR, as this function should also reset title to '' or null?

mjvkeulen: Behaviour is identical to production (sadly), probably best to create an issue as this PR does not address this?

Larspolo (approved):

engelhamer (approved):

PR #1118, opened 2021-03-21
Added a filter for past deadlines

mjvkeulen:
image
image

The options menu probably requires some additional styling in combination with the upcoming UI rework.

- Fixes Z#1238


src/vue/src/components/timeline/Timeline.vue
```diff
@@ -126,17 +142,40 @@ export default {
...
+                this.$store.commit(
+                    'preferences/SET_HIDE_PAST_DEADLINES',
+                    { hidePastDeadlines: value, aID: this.assignment.id },
+                )
+            },
+            get () { return this.$store.getters['preferences/hidePastDeadlines'] },
+        },
+        nodesHoldPastDeadlines () {
+            return this.nodes.some((node) => (
+                node.type === 'd'
```
Larspolo, line 162: Why is this not also considering passed progress goals / filled entries? Now I can (as a teacher):
1. go to journal 1 without the OLD entry filled
2. hide passed deadlines,
3. navigate to journal 2 with the OLD entry filled
4. Notice that the OLD progress node is hidden
5. Notice that I now can no longer unhide the hidden deadlines

image

Would personally like to now be able to filter the "OLD" nodes

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -31,21 +31,35 @@
...
+                    class="extra-filter-options-container"
+                >
+                    <b-form-checkbox
+                        v-if="nodesHoldPastDeadlines"
+                        v-model="hidePastDeadlines"
+                        @change="filterNodes"
+                    >
+                        Hide past deadlines
+                    </b-form-checkbox>
+                </div>
```
Larspolo, line 54: Maybe sth for @engelhamer , but I find the styling of this a bit odd. Would expect it to left align with the "Timeline" heading. Also the shadow / border is now not alligned with other cards we have elsewhere

image

image

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -193,29 +233,50 @@ export default {
...
-                    } else if (node.type === 'd') {
-                        if ('entry' in node && node.entry) {
-                            return this.hasCategory(node.entry.categories, filters)
-                        /* Deadline without entry, filter on template categories */
-                        } else if (node.template.categories) {
-                            return this.hasCategory(node.template.categories, filters)
-                        }
-                    } else if (node.type === 'a' || node.type === 'p') {
-                        return true
+        filterNodes () {
```
Larspolo, line 262: Would like some comment somewhere about what exactly gets filtered. Also maybe we need a difference between teachers & students.

Student -> hide progress nodes & filled entries
Teacher -> hide progress nodes & unfilled entries?

Larspolo (changes requested):

src/vue/src/components/timeline/Timeline.vue
mjvkeulen: I don't think we need to differentiate between student and teachers. Lets use the example of the JIR for someone redoing a year, a teacher would still like to be able to filter all the missed deadlines (which are irrelevant as the student imported them).

Can add a comment, but I would prefer making it obvious from the code instead.

src/vue/src/components/timeline/Timeline.vue
mjvkeulen: I believe this is addressed in the UI branch, the checkbox as is will fit quite well in the upcoming look.

src/vue/src/components/timeline/Timeline.vue
Larspolo: ATM I dont find it clear, having to look over multiple functions with many if's in each, but if you could make it more clear, I'm also fine with that. Think that having a comment somewhere on what to test when rewriting (parts of) this code would be usefull

engelhamer (approved):

src/vue/src/components/timeline/Timeline.vue
engelhamer: > Student -> hide progress nodes & filled entries

I would never want to filter posted entries (except for with the category filter). What do you meean by this?

Larspolo (approved):

PR #1119, opened 2021-03-23
Added various file handling improvements

mjvkeulen:
## Description
Added a progress bar indicator to file uploads
image

Display file size limits in MB
image

RT editor should now properly remove files that do not pass validation, tested via paste, drag, and the image upload menu. 

One remaining issue is that the image upload menu does not correctly display error messages (it did not do so before either). This [post](https://ittone.ma/ittone/tinymce-5-7-failure-callback-returns-object-object-no-matter-what/) describes our issue well, I propose we move this to a dedicated issue.

## Summary of changes
- Fixes Z#1688
- Fixes Z#1740
- Fixes Z#1686
- Fixes Z#1352
- Fixes Z#1695


src/vue/src/components/assets/TextEditor.vue
```diff
@@ -276,53 +276,44 @@ export default {
...
+                )
             } else {
                 formData.append('in_rich_text', true)
                 formData.append('file', file)
                 this.uploading += 1
                 this.$emit('startedUploading')
-                auth.uploadFile('files', formData)
+                auth.uploadFile('files', formData, {
+                    onUploadProgress: (e) => {
+                        progress((e.loaded / e.total) * 100)
```
Larspolo, line 306: [NIT] Would like to see a rounding on here as well, this now may result in odd numbering. Would even round it to whole numbers

image

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
```diff
@@ -2,12 +2,22 @@
...
             class="fileinput mb-2"
             @change="fileHandler"
         />
+        <b-progress
+            v-if="uploading"
+            :value="loaded"
+            :max="total"
+            show-progress
+            animated
+            class="multi-form"
```
Larspolo, line 18: [NIT] Can this be styled to match our other bar (top right)?
image

Larspolo, line 18: [NIT] I would prefer a % after the number, but idk if bootstrap allows for that

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
```diff
@@ -120,43 +134,70 @@ export default {
         }
     },
     methods: {
+        resetUploadState () {
+            this.uploading = false
+            this.total = 100
```
Larspolo, line 139: [NIT] Could this be stuck at ~90% (specifically for bonus points)? Because after the upload, the unpacking etc might also take some time, would like for it not to show 100% during that period

Larspolo (approved):

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
mjvkeulen: Removing animated would yield this:

image

But I quite like the striped/animated look, it fits the live update feel

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
mjvkeulen: As in, cap the actual upload itself at 90%? 

The bonus upload endpoint does require some processing time itself, but I am not sure what a clean solution would be. I doubt the actual processing time is sufficiently static (independent of the input file) to hardcode a solution around it (nor worth the time imo).

Larspolo: >As in, cap the actual upload itself at 90%? 

Yes, or just any arbitrary number to better, but still not truthfully reflect the actual status

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
Larspolo: Maybe sth for @engelhamer 

I would prefer without the stipes

mjvkeulen: Do not forget that its not just stripes or not, it is the animation which is tied to stripes as well.

src/vue/src/components/assets/file_handling/BonusFileUploadInput.vue
mjvkeulen: Currently, it accurately represents the upload status. I believe this is a good thing. 

If we also want some indication which represents that the server is still "processing" we could add that after the upload indicator.

mjvkeulen: The label is now "Processing..." when the upload is finished but the server has not responded yet.

engelhamer (approved):

PR #1122, opened 2021-04-01
Added drafts for student entries

Larspolo:
## Description
Added drafts for student entries
CREATION:
image

DISPLAY:
image

EDITING:
image

TIMELINE:
image

## Summary of changes
- Fixes Z#62


src/django/VLE/factory.py
```diff
@@ -205,7 +205,7 @@ def make_node(journal, entry=None, type=VLE.models.Node.ENTRY, preset=None):
     return VLE.models.Node.objects.get_or_create(type=type, entry=entry, preset=preset, journal=journal)[0]
 
 
-def make_entry(template, author, node, category_ids=None, title=None):
+def make_entry(template, author, node, category_ids=None, title=None, is_draft=None):
```
mjvkeulen, line 208: It would be more straightforward to default `is_draft` to False (matching the DB default). None is not valid input and need not be considered as a value?

src/django/VLE/models.py
```diff
@@ -1969,13 +1972,17 @@ def reset(self):
...
-    def get_sorted_nodes(self):
+    def get_sorted_nodes(self, author=None):
         """
         Get all the nodes of a journal in sorted order.
 
         Sorts first by preset due date, defaulting to entry creation date when dealing with an unlimited entry.
         """
-        return self.node_set.select_related(
+        nodes = self.node_set
+        if not author or not self.authors.filter(user=author).exists():
```
mjvkeulen, line 1982: Q) Why does the presence of author function as a check to exclude entry drafts? Should be adopted in the docstring, or a more sensible argument passed?

src/django/VLE/models.py
```diff
@@ -2699,10 +2714,24 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
-        if is_new and not isinstance(self, TeacherEntry):
+        if self.should_send_new_entry_notification(is_new, draft_before):
             generate_new_entry_notifications.apply_async(
                 args=[self.pk, self.node.pk], countdown=settings.WEBSERVER_TIMEOUT)
 
+    def should_send_new_entry_notification(self, is_new, draft_before):
+        print(draft_before, self.is_draft)
```
mjvkeulen, line 2722: Left over

src/django/VLE/models.py
```diff
@@ -2679,6 +2690,10 @@ def set_categories(self, new_category_ids, author):
 
     def save(self, *args, **kwargs):
         is_new = not self.pk
+        if not is_new and not self.is_draft:
+            draft_before = Entry.objects.get(pk=self.pk).is_draft
```
mjvkeulen, line 2694: draft_before as a variable name to me implies a reference to an entry instance itself, possibly was_draft/is_draft instead?

src/django/VLE/timeline.py
```diff
@@ -30,19 +30,19 @@ def get_nodes(journal, author=None):
...
+                needs_add_node = False
 
             node_list.append(get_progress(journal, node))
         elif node.is_entry:
             node_list.append(get_entry_node(journal, node, author))
         elif node.is_deadline:
             node_list.append(get_deadline(journal, node, author))
 
-    if can_add and not journal.assignment.is_locked():
+    if needs_add_node:
```
mjvkeulen, line 45: Q) How come the journal.assignment.is_locked() check can be omitted?

src/django/VLE/views/entry.py
```diff
@@ -40,11 +40,12 @@ def create(self, request):
         """
         journal_id, template_id, content_dict = utils.required_params(
             request.data, 'journal_id', 'template_id', 'content')
-        node_id, category_ids, title = utils.optional_typed_params(
+        node_id, category_ids, title, is_draft = utils.optional_typed_params(
```
mjvkeulen, line 43: Would it not make more sense to make is_draft a required param? An entry is either a draft or not, the backend can then always simply use the provided flag.

src/django/VLE/views/entry.py
```diff
@@ -128,8 +132,9 @@ def partial_update(self, request, *args, **kwargs):
         if title and not entry.template.chain.allow_custom_title:
             return response.forbidden('Entry does not allow for a custom title.')
 
-        # Check for required fields
-        entry_utils.check_fields(entry.template, content_dict)
+        # Check for required fields only when it was already not a draft OR it is now no longer a draft
+        if not entry.is_draft or not is_draft:
```
mjvkeulen, line 136: Q) Why does patch care about the current draft state of the entry? Does not only the requested state dictate whether validation should run?

src/django/VLE/views/entry.py
```diff
@@ -180,6 +185,9 @@ def partial_update(self, request, *args, **kwargs):
         entry.last_edited_by = request.user
         entry.last_edited = timezone.now()
         entry.title = title
+        # Entries should not be able to be posted
```
mjvkeulen, line 188: Q) I don't quite follow the comment in regards to the code block?

src/django/VLE/views/entry.py
```diff
@@ -180,6 +185,9 @@ def partial_update(self, request, *args, **kwargs):
...
+        if not is_draft:
+            entry.is_draft = False
```
mjvkeulen, line 190: Making is_draft a required param would streamline such logic as well.

src/django/VLE/views/journal_import_request.py
```diff
@@ -79,7 +79,7 @@ def partial_update(self, request, *args, **kwargs):
 
         source_entries = Entry.objects.none()
         if not jir_action == jir.DECLINED:
-            source_entries = Entry.objects.filter(node__journal=jir.source)
+            source_entries = Entry.objects.filter(node__journal=jir.source, is_draft=False)
```
mjvkeulen, line 82: Would it make sense to let the default Entry manager exclude drafts? I believe this would make working with draft less error prone, as one has to always explicitly reason about working with draft entries or not, similar to how we work with journals?

mjvkeulen (changes requested):
Nice swift PR!

- How does a user go about explicitly clearing or restoring a draft?
- Drafts should be autosaved saved so users can profit from this nice feature when they encounter an error or simply make a mistake by closing a tab or something

src/django/VLE/timeline.py
Larspolo: Moved to inside the can_add function

src/django/VLE/views/entry.py
Larspolo: Defaulted to nothing to easily implement it with existing tests and teacher entry creation/updating

src/django/VLE/views/entry.py
Larspolo: is_draft param could be set to true when the entry is already not a draft anymore. This would omit the call, but in actual fact, it would still need to be validated because it is posted entry

src/django/VLE/views/journal_import_request.py
Larspolo: Going to the case of excluding teacher journals. That is a way stronger case of them being excluded. Entry drafts imo should act like just any other entry, except for cases as needs_marking, notifications and JIR.

Also excluding it in the default manager, still exposes it to sth like Node.objects.filter(entry...) which is also used  often. 

I think even when by default excluded, we will still need to think about weather or not we want to include drafts

Larspolo:
> How does a user go about explicitly clearing or restoring a draft?

Its all the same as clearing or restoring any other entry

src/django/VLE/factory.py
Larspolo: Could, but `optional_params` would set it to None nonetheless. This would then still require more work

src/django/VLE/models.py
Larspolo: Adopted in the docstring, its used to check "can_see_drafted_entries"

mjvkeulen:
>> How does a user go about explicitly clearing or restoring a draft?

> Its all the same as clearing or restoring any other entry

When a user saves a draft and then later wants to reset, I don't believe this is possible right now? Edit -> cancel would still restore to the latest draft state stored remotely as far as I understand it?

src/django/VLE/views/entry.py
mjvkeulen: is_draft should not be optional when posting an entry, a user always intentionally posts an entry as ready for submission or explicitly as a draft. This should be reflected here.

mjvkeulen: The benefit is that we also don't have to deal with None as a possible value. 

I hope fixing this in the tests should be as straightforward as adding one field to the entry params factory.

src/django/VLE/models.py
mjvkeulen: The can_see_drafted_entries makes a lot of sense. 

Possibly move this out of this function as it does not align with getting sorted nodes? A solution would be to apply the exclude on the return value directly and keep this function pure.

src/django/VLE/views/entry.py
mjvkeulen: So this is an interesting situation, did you block a user from saving a new draft when editing an already posted entry?

When a student now edits an entry that has already been created, then saves its edit as a draft. What does the teacher see? Is this combination not possible?

Larspolo:
> When a user saves a draft and then later wants to reset, I don't believe this is possible right now? Edit -> cancel would still restore to the latest draft state stored remotely as far as I understand it?

With the delete button bottom left. Is a name change to "discard draft" wanted?

src/django/VLE/views/entry.py
Larspolo: > did you block a user from saving a new draft when editing an already posted entry?

No, only entries that are already un-drafted (aka posted) cannot be re-drafted

src/django/VLE/views/entry.py
Larspolo: Fixed by allowing `optional_typed_params` to also take defaults as a thrid value, now it never is `None`. I like the API not needing a draft value, where it will be undrafted by default

src/django/VLE/models.py
Larspolo: > Can we encapsulate that in the manager/queryset?

I dont see where / how exactly. Moved it to a separate function.

mjvkeulen:
To summarize critique and earlier discussions:

- _Entry drafts should be auto-saved_: this will help users who encounter an error / close the tab by accident etc.
We agreed to implement this at a later date.

- _It is not possible to create a draft as an improvement to an already submitted entry_. E.g. a teacher marks an entry for improvement, a student can no longer use the draft functionality to improve his/her submission.
I believe we agreed to implement this at a later date?

- _The action of saving a draft should not take the user out of edit mode_. It is likely they might want to save now, but continue editing their work.
Under discussion

- Builds on point 2: _Users should be presented with a (subtle) indication that their draft is saved_ (it is now safe to leave the page without posting the entry)
image
Under discussion

- _Discarding a draft should be explicitly communicated as such_. Currently, we still phrase discarding as a successful deletion. E.g. "Are you sure that you want to delete this entry?" and "Entry successfully deleted".

src/django/VLE/views/entry.py
mjvkeulen: I believe we disagree on this, and we now have a bit more complexity (but also more flexibility) with optional params with configurable defaults.

I am fine with resolving as is.

Larspolo:
I'm fine with implementing / discussing these points (preferably not over text), but atm it feels a lot like we are postponing a useful feature for students only because we want it more useful.

src/django/VLE/models.py
```diff
@@ -1977,13 +1980,26 @@ def reset(self):
...
+    def get_sorted_nodes(self, user=None):
         """
         Get all the nodes of a journal in sorted order.
 
         Sorts first by preset due date, defaulting to entry creation date when dealing with an unlimited entry.
+        If "user" is set, it checks if the user can see drafted entries, if so, also return drafted entries
         """
-        return self.node_set.select_related(
+        nodes = self.node_set
+        if not self.can_see_drafted_entries(user):
```
engelhamer, line 1999: If `user` is not set the `can_see_drafted_entries` call also returns `False`, so the negation makes it so that drafts are included if no user is set. This contradicts the docstring, and I think it would be better to adhere to the situation described there.

src/django/VLE/models.py
```diff
@@ -2687,6 +2707,10 @@ def set_categories(self, new_category_ids, author):
 
     def save(self, *args, **kwargs):
         is_new = not self.pk
+        if not is_new and not self.is_draft:
+            was_draft = Entry.objects.get(pk=self.pk).is_draft
+        else:
+            was_draft = True
```
engelhamer, line 2713: Also marks edited non-draft entries as `was_draft`.

i.e. create an entry, post it (no draft), edit it -> `was_draft` true.

engelhamer (approved):

mjvkeulen:
Discarding draft still triggers progress will not be saved if you leave warning

mjvkeulen:
It is possible to create a draft for an empty entry

src/vue/src/components/entry/Entry.vue
```diff
@@ -302,7 +376,7 @@ export default {
             // It is safe to leave the entry if it is not currently being edited AND if no content for an entry to be
             // created is provided.
             return (
-                !this.edit
+                (!this.node || !this.node.entry || this.$_isEqual(this.newEntryContent, this.node.entry.content))
```
mjvkeulen, line 379: The entry title is not part of the entry content, so it is now possible to navigate away without warning after entering a title.

mjvkeulen (approved):

PR #1125, opened 2021-04-11
Fixed issue where institution wide only teachers would become course teachers

Larspolo:
## Description
Fixed issue where institution wide only teachers (so not also a teacher in the course itself) would become course teachers

### DIFF
--- a/src/django/VLE/settings/base.py
+++ b/src/django/VLE/settings/base.py
@@ -85,14 +85,16 @@ EMAILS = Emails()
 LTI_SECRET = os.environ['LTI_SECRET']
 LTI_KEY = os.environ['LTI_KEY']
 ROLES = OrderedDict({
-    'Teacher': ['instructor', 'administrator'],
-    'TA': ['teachingassistant'],
+    'Teacher': ['urn:lti:role:ims/lis/instructor', 'urn:lti:role:ims/lis/contentdeveloper', 'administrator'],
+    'TA': ['urn:lti:role:ims/lis/mentor', 'urn:lti:role:ims/lis/teachingassistant'],
     'Student': ['learner']
 })
 LTI_ROLES = OrderedDict({
-    'instructor': 'Teacher',
+    'urn:lti:role:ims/lis/instructor': 'Teacher',
+    'urn:lti:role:ims/lis/contentdeveloper': 'Teacher',
     'administrator': 'Teacher',
-    'teachingassistant': 'TA',
+    'urn:lti:role:ims/lis/mentor': 'TA',
+    'urn:lti:role:ims/lis/teachingassistant': 'TA',
     'learner': 'Student',
 })
 LTI_TEST_STUDENT_FULL_NAME = 'Test student'
--- a/src/django/VLE/lti_launch.py
+++ b/src/django/VLE/lti_launch.py
@@ -57,12 +57,15 @@ def roles_to_list(params):
     roles = list()
     if 'roles' in params:
         for role in params['roles'].split(','):
-            roles.append(role.split('/')[-1].lower())
+            roles.append(role.lower())
     return roles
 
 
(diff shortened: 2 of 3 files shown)


src/django/VLE/settings/base.py
```diff
@@ -85,14 +85,16 @@ class Emails:
 LTI_SECRET = os.environ['LTI_SECRET']
 LTI_KEY = os.environ['LTI_KEY']
 ROLES = OrderedDict({
-    'Teacher': ['instructor', 'administrator'],
-    'TA': ['teachingassistant'],
+    'Teacher': ['urn:lti:role:ims/lis/instructor', 'urn:lti:role:ims/lis/contentdeveloper', 'administrator'],
```
mjvkeulen, line 88: No test exists for contentdeveloper

src/django/VLE/settings/base.py
```diff
@@ -85,14 +85,16 @@ class Emails:
...
+    'TA': ['urn:lti:role:ims/lis/mentor', 'urn:lti:role:ims/lis/teachingassistant'],
```
mjvkeulen, line 89: No test exists for mentor

Larspolo: Also not for teaching assistant, added now

src/django/VLE/settings/base.py
Larspolo: Also not for teaching assistant, added now

mjvkeulen (approved):

src/django/VLE/settings/base.py
engelhamer, line 88: Why is `administrator` not prefixed? Should that not be `urn:lti:role:ims/lis/Administrator` or `urn:lti:instrole:ims/lis/Administrator` then?

src/django/VLE/settings/base.py
```diff
@@ -85,14 +85,16 @@ class Emails:
...
     'Student': ['learner']
 })
 LTI_ROLES = OrderedDict({
-    'instructor': 'Teacher',
+    'urn:lti:role:ims/lis/instructor': 'Teacher',
+    'urn:lti:role:ims/lis/contentdeveloper': 'Teacher',
     'administrator': 'Teacher',
```
engelhamer, line 95: Same as above.

src/django/VLE/settings/base.py
```diff
@@ -85,14 +85,16 @@ class Emails:
...
-    'teachingassistant': 'TA',
+    'urn:lti:role:ims/lis/mentor': 'TA',
+    'urn:lti:role:ims/lis/teachingassistant': 'TA',
     'learner': 'Student',
```
engelhamer, line 98: Also here there should be a prefix? This will not match.

src/django/VLE/settings/base.py
```diff
@@ -85,14 +85,16 @@ class Emails:
 LTI_SECRET = os.environ['LTI_SECRET']
 LTI_KEY = os.environ['LTI_KEY']
 ROLES = OrderedDict({
-    'Teacher': ['instructor', 'administrator'],
-    'TA': ['teachingassistant'],
+    'Teacher': ['urn:lti:role:ims/lis/instructor', 'urn:lti:role:ims/lis/contentdeveloper', 'administrator'],
+    'TA': ['urn:lti:role:ims/lis/mentor', 'urn:lti:role:ims/lis/teachingassistant'],
     'Student': ['learner']
```
engelhamer, line 90: Also a prefix here?

engelhamer (changes requested):

src/django/VLE/settings/base.py
engelhamer: Why was `contentdeveloper` added?

engelhamer (approved):

PR #1130, opened 2021-04-21
Split models into separate files

mjvkeulen:
Logic grouping of models, in a models module following https://docs.djangoproject.com/en/2.2/topics/db/models/#organizing-models-in-a-package

- [x] Incorporate new models.py updates before merge with develop: Use common ancestor with develop to determine changes made to models.py since (git merge-base develop models-split -> git diff bd95e0152f724fba9bcb2890686ef4180f8ddc39..develop src/django/VLE/models.py)


src/django/VLE/models/__init__.py
```diff
@@ -0,0 +1,26 @@
+from .assignment import Assignment  # noqa: F401
```
Larspolo, line 1: (relating to the LTI comment) You also didnt find sth better then this? Or did you not (yet) look into this?

Larspolo (approved):

src/django/VLE/models/__init__.py
mjvkeulen: Sorry the LTI comment?

I did not find a way around importing all models in the init file, I believe this the recommend approach (see link in description)

Larspolo: https://github.com/eJourn-al/eJournal/pull/1123/files/2ca5324182850679a8d01cf31d33a8947d8124cf#diff-905d14326e220699e7e8ada18222953f5ea917434f85df5772971da920c6e783?w=1

engelhamer (approved):

PR #1131, opened 2021-04-23
Added support for Kaltura via video embed

mjvkeulen:
A video field is added with all hosts enabled by default.

If Kaltura is an allowed host, the video input becomes a text area. This is to convey that more than a simple video link is expected:
image
image

Otherwise, the input remains a single line input as a simple video link is expected
image
image

If any validation errors occur during input the user is presented with an information icon which opens a modal for instructions:
image
image

- Fixes Z#1480


src/django/test/test_utils.py
```diff
@@ -283,3 +283,29 @@ def test_validate_youtube_url_with_video_id(self):
...
+            '<iframe src="https://api.eu.kaltura.com/<anything_but_space_goes>" <any tags are allowed but not captured',
```
Larspolo, line 290: Although the current solution works, I would feel more save if we would remove/disallow any non iframe tag as well. Just in case we later on change sth / forget sth

src/vue/src/components/entry/EntryVideoFieldDisplay.vue
```diff
@@ -0,0 +1,64 @@
...
+        kalturaAllowed () { return this.field.options.split(',').includes('k') },
+    },
+    methods: {
+        validatedSrc () {
+            const youtubeId = genericUtils.parseYouTubeVideoID(this.data)
+            const kalturaEmbedCodeSrc = genericUtils.praseSrcFromKalturaEmbedCode(this.data)
+
+            if (this.youtubeAllowed && this.kalturaAllowed && (youtubeId || kalturaEmbedCodeSrc)) {
+                if (youtubeId) {
+                    return `https://www.youtube.com/embed/${youtubeId}?rel=0&amp;showinfo=0`
```
Larspolo, line 41: [Q] Why is "https://www.youtube.com/embed/" etc not also returned in the `parseYouTubeVideoID` function? If for no real reason, I would prefer to not have to add that here as well

src/vue/src/components/entry/EntryVideoFieldInput.vue
```diff
@@ -0,0 +1,160 @@
+<template>
```
Larspolo, line 1: Nice tutorial :)

src/vue/src/components/entry/EntryVideoFieldInput.vue
```diff
@@ -0,0 +1,160 @@
...
+            } else if (this.youtubeAllowed) {
+                return 'Enter a YouTube video URL'
+            } else if (this.kalturaAllowed) {
+                return 'Enter a Kaltura video embed code'
+            } else {
+                this.$store.commit('sentry/CAPTURE_SCOPED_MESSAGE', {
+                    msg: 'Video host incorrectly configured.',
+                    extra: { field: this.field, data: this.value },
+                })
+                return 'Unknown video host selected, please contact us'
```
Larspolo, line 129: Would remove ", please contact us" in the placeholder tbh, doesnt really do much. Maybe "Your teacher has selected an invalid video host, if this issue persists, contact your teacher." as a warning somewhere

Larspolo (changes requested):
Looks good, just a few minor things, open for discussion

src/django/test/factory/content.py
```diff
@@ -38,8 +40,10 @@ def gen_valid_non_file_data(field):
     if field.type == Field.TEXT:
         return factory.Faker('text').generate()
     if field.type == Field.VIDEO:
-        # According to our current validators this can be any url
-        return 'https://www.youtube.com/watch?v=lJMNA7UcpxE'
+        if field.youtube_allowed:
+            return 'https://www.youtube.com/watch?v=lJMNA7UcpxE'
```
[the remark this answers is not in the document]
mjvkeulen: I will do so

src/django/test/test_utils.py
mjvkeulen: The additional tags are part of the Kaltura embed code. As in, these will always be pasted by the user as part of the embed code. 

The regex specifically captures only the **src** attribute. However, it is important the regex only accepts the iframe embed code, because a default video link to a Kaltura resource link misses certain params which would allow the media to be played from another origin.

src/vue/src/components/entry/EntryVideoFieldDisplay.vue
mjvkeulen: I don't feel strongly about this, the function (as we have it currently), explicitly captures the video id from a valid YouTube url. What is wrong with the existing approach?

src/django/test/factory/field.py
```diff
@@ -21,6 +21,8 @@ def _select_location(g_field):
 def _gen_options(g_field):
     if g_field.type == Field.SELECTION:
         return '["a", "b", "c"]'
+    elif g_field.type == Field.VIDEO:
+        return f'{Field.YOUTUBE},{Field.KALTURA}'
```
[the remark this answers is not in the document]
mjvkeulen: I could see the benefit of changing the DB field type to JSON or ARRAY.

src/vue/src/components/entry/EntryVideoFieldDisplay.vue
Larspolo: Because `parseYouTubeVideoID` does return the entire URL, mostly just consistency, not required

src/django/test/test_utils.py
Larspolo: Would storing only the src part work as well? After it is stripped from the iframe tag

mjvkeulen: We could but I don't know exactly which of all these params (I experimented and the flashvars were relevant :() are required, so we would still have one ugly as link.

The regex is used to confirm that the user pastes the entire embed code (iframe and all), would you then use a different regex front and back?

If we use the same regex to parse the source from the embed code to strip, as we use to display the source anyway, there is no actual security gain.

Larspolo (approved):

src/django/test/test_utils.py
engelhamer: Storing only source would make editing the entry confusing: the input is not the original embed paste anymore.

src/vue/src/components/entry/EntryVideoFieldInput.vue
```diff
@@ -0,0 +1,162 @@
+<template>
+    <!-- eslint-disable vue/attribute-hyphenation -->
+    <b-form-group
+        :state="state"
+    >
+        <b-form-textarea
```
engelhamer, line 6: Would like to see a preview of the kaltura video maybe? Now it is a bit hard to tell what I am really embedding.

src/vue/src/components/entry/EntryVideoFieldInput.vue
```diff
@@ -0,0 +1,162 @@
...
+                            <b>Copy the link</b>: Click the <b>Copy</b> button to copy a link to the video.
+                        </li>
+                    </ol>
+                </template>
+
+                <template v-if="kalturaAllowed">
+                    <h2>Kaltura</h2>
+
+                    <ol>
+                        <li>Navigate to your Kaltura environment.</li>
```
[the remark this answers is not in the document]
engelhamer: Also because Kaltura is semi-whitelabel (no logo or name on the page, custom url, ..)

src/vue/src/components/template/TemplateField.vue
```diff
@@ -75,6 +75,17 @@
                         @input="selectedExtensionType = ' '"
                     />
                 </div>
+                <theme-select
```
engelhamer, line 78: I would like to see some text here also when something is selected. See e.g. the group filter on the assignment page.

engelhamer: Currently it's empty, and with both options selected by default the user may not know what this dropdown is for until they click on it.

engelhamer (changes requested):

src/vue/src/components/entry/EntryVideoFieldInput.vue
```diff
@@ -0,0 +1,162 @@
...
+            v-else
+            v-model="data"
+            :placeholder="placeholder"
+            class="theme-input"
+            type="text"
+            trim
+        />
+        <template #invalid-feedback>
+            {{ invalidFeedback }}
+            <icon
```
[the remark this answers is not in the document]
mjvkeulen: Field headings are optional, I moved the icon to the append slot of the input

src/django/test/test_utils.py
Larspolo: True, but when editing, we could also re-add the <iframe> tags

src/vue/src/components/entry/EntryVideoFieldInput.vue
engelhamer: @mjvkeulen how about this?

mjvkeulen: I am not sure here, a decent preview takes up quite some space. Our current process doesn't provide a preview either, and the addition of Kaltura as a supported video host does not change this. 

Possibly something to move to an issue if you feel strongly?

engelhamer (approved):

src/vue/src/components/entry/EntryVideoFieldInput.vue
engelhamer: Okay, then maybr an issue for 'video embed preview'

engelhamer (approved):

PR #1132, opened 2021-04-24
Remove specific test user full_name check

mjvkeulen:
A test user can be launched with different names, something we learned after observing a test user launch with the full_name 'Cusist toetsen' instead of the expected 'Test user'.

Whilst the params for a test user on first launch are LMS depedent, for Canvas we can move towards using the following solution:

If you are using the instructor student view, the LTI custom parameter Canvas.masqueradingUser.id will be set.  This alone won't let you know its the student view, but would let you LTI know its a masquerade.  There might be other LTI parameters that would help tie it down.  For example, lis_person_sourcedid seems to be missing for student view, but is there for an admin masquerading, so that might be a way of distinguishing the cases.

For now we have no normal users without an email adress, so loosening the check for lack of email value will suffice for now.

- Fixes #1832


src/django/VLE/views/lti.py
```diff
@@ -115,11 +115,16 @@ def get_finish_state(user, assignment, lti_params):
...
-       and 'custom_user_full_name' in params and params['custom_user_full_name'] == settings.LTI_TEST_STUDENT_FULL_NAME:
+    if not user and 'custom_user_email' in params and params['custom_user_email'] == '':
         lti_id, username, full_name, email, course_id = utils.required_params(
             params, 'user_id', 'custom_username', 'custom_user_full_name', 'custom_user_email', 'custom_course_id')
+
+        if len(username) <= 8:
+            with push_scope() as scope:
+                scope.set_context('params', params)
+                capture_message(
+                    'Test student created with unexpectedly short username (could be student)', level='warning')
```
Larspolo, line 126: Great idea!

src/django/test/factory/params.py
```diff
@@ -23,9 +22,9 @@ class Meta:
 
 
 class JWTTestUserParamsFactory(JWTParamsFactory):
+    # NOTE: custom_user_full_name can vary, we have seen 'Test student' but also 'Cursist X'
     custom_user_email = ''
     user_id = factory.Sequence(lambda x: "305c9b180a9ce9684ea62aeff2b2e97052cf2d4b{}".format(x))
-    custom_user_full_name = settings.LTI_TEST_STUDENT_FULL_NAME
```
Larspolo, line 28: [NIT] IMO we can still leave this in / at least test that if given, the full_name is used properly. Same for `TestUserFactory`

Larspolo (approved):

engelhamer (commented):
I find this too risky: if not manually detected soon enough, a genuine student's work could be discarded just because they and one of their fellow students happen to not have an email address set. Also: launching the student view as a teacher could cause the same to happen to a student's work.

The name constraint was necessary in my opinion. If we cannot enforce it anymore, I would like to add an explicit consent in the LTI launch process where a 'normal' user would set a password. This should state that no email address is set, and if the user is a student they should first set an email address in their LMS (or maybe set one with us, that is yet to be verified). Here can be also a button that marks the current user as a 'student view', with the warning that the account should not be used for real work and can be overwritten at any time by a new test student. This should be in the same way as e.g. terms and conditions are agreed upon for some web services (with a checkbox).

src/django/test/factory/params.py
mjvkeulen: Name is not removed from the class as it inherits from JWTParamsFactory?

src/django/VLE/views/lti.py
```diff
@@ -233,6 +233,8 @@ def lti_launch(request):
             refresh = TokenObtainPairSerializer.get_token(user)
             user.last_login = timezone.now()
             if 'custom_user_email' in params and params['custom_user_email']:
+                if user.is_test_student:
+                    raise VLEProgrammingError('A test student is launched with a valid email address')
```
Larspolo, line 237: Would change this error to sth a user would also understand.
"You are trying to add an email address to a test student account. This should not be possible. Please contact support@ejournal.app if you are not a test student."

engelhamer (approved):

PR #1137, opened 2021-05-02
Fixed an issue where TODOs would not show for TAs

mjvkeulen:


Larspolo (approved):
I assume this is tested, never knew this operator, it looks really good for other places as well

engelhamer (approved):

PR #1139, opened 2021-05-03
Fixed an issue where multiple errors would toast on token timeout

mjvkeulen:
Quick fix, not pretty


Larspolo (approved):

engelhamer (approved):
Maybe add comment stating that this will be updated

PR #1140, opened 2021-05-03
Fixed an issue where the timeline assignment could be undefined

mjvkeulen:
This fix is mostly a bandaid on top of the multiple calls to assignment retrieve with different arguments throughout the app.

A proper solution requires a refactor: isolate expensive assignment serializer logic to dedicated APIs. E.g. stats, journals, and courses.


engelhamer (approved):

engelhamer (approved):

src/vue/src/components/timeline/Timeline.vue
```diff
@@ -61,10 +61,14 @@
                 aria-expanded="false"
                 aria-controls="timeline-container"
             >
-                <timeline-nodes
-                    :filteredNodes="filteredNodes"
-                    :allNodes="nodes"
-                />
+                <!-- Nested components make use of the store assignment, this ensures both are available -->
```
Larspolo, line 64: How do I test that users now dont end up with an ever loading icon?

Larspolo (approved):

PR #1142, opened 2021-05-04
Changed edit entry layout to display optional flag iso required

Larspolo:
## Description
OLD:
image

NEW:
image

- Fixes Z#1906


engelhamer (changes requested):
I would propose using the same approach as the `required` css styling here, so with selectors and `:after`

Larspolo:
Done

engelhamer:
How about the other fields (why have a star for e.g. password forms etc, but optional here)?

Z#1906 is not fixed, since fields without a title are still not indicated as required or optional.

Larspolo:
I did not go over every input field we have, thag can be done as well, but was not the goal of this PR. 

This PR originated from the many users that fill it in even when not needed

There are two types of forms. One where all fields are expected to be optional unless stated otherwise, and one where all fields are ezpected to be required unless stated otherwise.  Entry fields are clearly taken as everything is required by the userbase. On the otherhand we have an assignment creation where the teacher ezcepts to not have to fill everything in, here a required star is imo more favourable

Larspolo:
> Z#1906 is not fixed, since fields without a title are still not indicated as required or optional.

It does, see the screenshot below (top to bottom inc title: optional, required, required, optional):
image

src/vue/src/sass/partials/text.sass
```diff
@@ -37,6 +37,10 @@ $font-family-sans-serif: 'Roboto', sans-serif
     &.required::after
         content: " *"
         color: $theme-red
+    &.optional::after
```
mjvkeulen, line 40: Move to class or duplicate and add to the .form-group (&.optional, &[optional]) block so we can use this with form groups as well.

Would this also allow you to remove some of the `<span class="field-heading optional">` blocks?

Larspolo: I added it in the sass, but idk how/where that would reduce the span

mjvkeulen (approved):

engelhamer (approved):

PR #1143, opened 2021-05-06
Execute click on dropdown select

Larspolo:
## Description
Execute click on dropdown select

### DIFF
--- a/src/vue/src/components/assets/DropdownButton.vue
+++ b/src/vue/src/components/assets/DropdownButton.vue
@@ -115,6 +115,7 @@ export default {
         emitChangeOption (option) {
             this.isOpen = false
             this.$emit('change-option', option)
+            this.$emit('click', option)
         },
     },
 }


src/vue/src/components/assets/DropdownButton.vue
```diff
@@ -115,6 +115,7 @@ export default {
         emitChangeOption (option) {
             this.isOpen = false
             this.$emit('change-option', option)
+            this.$emit('click', option)
```
engelhamer, line 118: How does this work timing-wise? Could the change-option listener be not completed when this fires?

engelhamer (commented):
How

src/vue/src/components/assets/DropdownButton.vue
Larspolo: Think so

mjvkeulen: The events should be processed in order: options change should be handled before the click, there is no concurrency at work here.

But how to truly show this idk

Larspolo: Yeah, its ordered

engelhamer:
Can be closed as we are opting for another solution?

PR #1144, opened 2021-05-06
Added auto-save for entry drafts

Larspolo:
## Description
Added the ability for students to auto save

## Summary of changes
- Fixes Z#1810


src/django/test/factory/user.py
```diff
@@ -9,7 +9,7 @@ class Meta:
         model = 'VLE.User'
 
     username = factory.Sequence(lambda x: f"{10000000 + x}")
-    full_name = factory.Faker('name_nonbinary')
+    full_name = factory.Faker('name')
```
mjvkeulen, line 12: Q) What was the difference?

src/django/VLE/models.py
```diff
@@ -555,6 +555,9 @@ class Preferences(CreateUpdateModel):
     auto_proceed_next_journal = models.BooleanField(
         default=False
     )
+    auto_save_drafts = models.BooleanField(
```
mjvkeulen, line 558: I believe the user should not be confronted with a choice here, no need to add any complexity when auto-saving is just ideal behaviour for the user. 

If you insist on making this a setting: I propose moving this to the profile settings so the user not confronted with this option when posting an entry each time.

src/vue/src/components/entry/Entry.vue
```diff
@@ -288,8 +307,54 @@ export default {
...
+                }
+            },
+        },
+        'preferences.auto_save_drafts': {
+            immediate: true,
+            handler () {
+                this.doAutoSave()
+            },
+        },
+    },
```
mjvkeulen, line 332: You can combine this into a single watcher for three variables (unless deep conflicts with this somehow).

src/vue/src/components/entry/Entry.vue
```diff
@@ -288,8 +307,54 @@ export default {
...
+        newEntryContent: {
+            deep: true,
+            handler () {
+                if (this.preferences.auto_save_drafts) {
+                    this.doAutoSave()
+                }
+            },
+        },
+        title: {
+            deep: true,
```
mjvkeulen, line 319: Title is not a nested object so deep can be omitted

src/vue/src/components/entry/Entry.vue
```diff
@@ -288,8 +307,54 @@ export default {
...
     },
     methods: {
+        ...mapMutations({
+            updatePreferences: 'preferences/CHANGE_PREFERENCES',
+            setCurrentNode: 'timeline/SET_CURRENT_NODE',
+        }),
+        doAutoSave () {
+            window.clearTimeout(this.autoSaveTimeout)
+
+
```
mjvkeulen, line 344: How does the linter not detect this.

src/vue/src/components/entry/Entry.vue
```diff
@@ -381,7 +454,7 @@ export default {
             // For newly created entries simply check if any field or the title has any content
             if (this.create) {
                 return this.template.field_set.some((field) => this.newEntryContent[field.id])
-                    || this.title !== this.node.entry.title
+                    || this.title !== ''
```
mjvkeulen, line 457: Are you sure this still works for a preset entry where the template allows a custom title, but the teacher did specify a default display name?

mjvkeulen (changes requested):
Looking forward to this one!

Currently crashes when creating an entry for a deadline (does not even show entry component).

src/vue/src/components/entry/Entry.vue
```diff
@@ -288,8 +307,54 @@ export default {
...
+        doAutoSave () {
+            window.clearTimeout(this.autoSaveTimeout)
+
+
+            if (this.entryContentHasPendingChanges()) {
+                this.autoSaveMessage = ''
+                this.allowPreview = false
+                const func = this.create ? this.createEntry : this.saveChanges
+                this.autoSaveTimeout = window.setTimeout(() => {
+                    this.autoSaveMessage = 'Saving...'
```
mjvkeulen, line 350: I think this is not visible as it is cleared on the next keypress. You should play around with the ordering here or clear the autoSaveMessage in a finally block of func

src/django/test/factory/user.py
Larspolo: non binary didnt work. Figured this was the quickest fix

src/vue/src/components/entry/Entry.vue
Larspolo: Thats on purpose. Wanted to not have a long "Saving...", Only when actually saving

src/django/VLE/models.py
Larspolo: Its not much of a confrontation, and some users have less internet and/or PC specs, think for those its nice to quickly turn it off

src/vue/src/components/entry/Entry.vue
Larspolo: How?

src/vue/src/components/entry/Entry.vue
Larspolo: Kinda, now it the student does not automatically start filling in the title, the teacher's title gets put into that field. On creation the title starts empty anyway. This could be changed, but that is kept default. Unless you want to check `this.title !== '' && this.title !== this.node.display_name` for the specific cases where the student copy pastes the original title and doesnt want to auto save the draft

src/vue/src/components/entry/Entry.vue
mjvkeulen: https://stackoverflow.com/questions/42737034/vue-js-watch-multiple-properties-with-single-handler

Again I don't know how deep will factor into this

src/vue/src/components/entry/Entry.vue
mjvkeulen: But currently, it does not show when actually saving afaik. That would require the autosave message to be cleared in the .then of the request itself?

mjvkeulen: That makes sense

src/vue/src/components/entry/Entry.vue
Larspolo: The clean option is Vue 3 only. I think for the Vue 2, having a computed variable is not ideal either. Are you sure you want that solution?

mjvkeulen: If this would become:

computed: newEntry (or autoSaveOnUpdate if we keep the pref)
watch
    newEntry: do stuff 

That seems fine with me, but it's not as clean as the Vue3 solution

src/vue/src/components/entry/Entry.vue
```diff
@@ -362,11 +429,18 @@ export default {
                     category_ids: categoryIds,
                     title: this.template.allow_custom_title ? this.title : null,
                     is_draft: isDraft,
-                }, { customSuccessToast: 'Entry successfully posted.' })
+                }, { customSuccessToast: this.preferences.auto_save_drafts
+                    ? 'Created draft entry.'
+                    : 'Entry successfully posted.' })
```
mjvkeulen, line 434: I believe if it's an autosave call the user should not see a toast. Again following the idea that auto-saving should be the default and as nondisruptive as possible

src/vue/src/components/entry/Entry.vue
```diff
@@ -46,14 +46,28 @@
...
-                    @click="clearDraft()"
-                >
-                    <icon name="ban"/>
-                    Cancel
-                </b-button>
+                <template v-if="edit">
+                    <b-button
+                        v-if="preferences.auto_save_drafts"
+                        class="ml-2 green-button float-right"
+                        :class="{ 'input-disabled': !allowPreview }"
```
mjvkeulen, line 53: [NIT]: Add a little tooltip for why preview is disabled, eg. "Saving draft" or something similar.

src/vue/src/components/entry/Entry.vue
mjvkeulen: Lets leave it as a nit then!

src/vue/src/components/entry/Entry.vue
Larspolo: There is a lot of visual change: comment section is added, preview button is added, node is added to timeline.

This created the behaviour of not knowing what happened when I tested this with an end user. Thats why I added the toast, for when users do act suppriced, they have a toast to display them what happened. After this there are no toasts, because the visual change is also minimal (only the "saving..." / "saved" text)

src/vue/src/components/entry/Entry.vue
Larspolo: When I attempted this, the tooltip kept showing when it was enabled again, so maybe if you know the fix we can add it, but else Im gonna leave it for now

src/vue/src/components/entry/Comments.vue
```diff
@@ -42,6 +42,10 @@ export default {
             type: Boolean,
             default: false,
         },
+        isDraft: {
+            type: Boolean,
+            default: false,
+        },
```
mjvkeulen, line 48: NIT: eID -> Entry object

src/vue/src/components/entry/Entry.vue
```diff
@@ -46,11 +46,11 @@
...
                     </template>
                 </template>
-                <template v-if="edit">
+                <template v-if="edit || create">
                     <b-button
-                        v-if="preferences.auto_save_drafts"
+                        v-if="preferences.auto_save_drafts || !entryContentHasPendingChanges()"
                         class="ml-2 green-button float-right"
-                        :class="{ 'input-disabled': !allowPreview }"
+                        :class="{ 'input-disabled': requestInFlight || create }"
```
mjvkeulen, line 53: Q) This reads as preview disabled when creating?

src/vue/src/components/entry/Entry.vue
```diff
@@ -345,23 +342,29 @@ export default {
...
-                            this.autoSaveMessage = 'Failed to save'
-                        })
-                }, this.create ? 5000 : 1000)
-            } else {
+                    if (this.create) {
+                        this.createEntry()
+                    } else {
+                        this.autoSave()
+                    }
+                }, this.create ? 1000 : 1000)
```
mjvkeulen: Left over from refactor (silly linter tbh)

src/vue/src/components/template/TemplateCard.vue
```diff
@@ -0,0 +1,87 @@
...
+        width: calc(100% - 60px)
+        min-height: 45px
+        flex-direction: column
+        padding-left: 20px
+        .max-one-line
+            display: block
+            width: 100%
+            text-overflow: ellipsis
+            white-space: nowrap
+            overflow: hidden
```
mjvkeulen, line 76: Prob have to cleanup/extract some of this css?

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+    <div v-if="templates.length > 1">
```
mjvkeulen, line 2: NIT: I believe the div is not needed due to the v-if / else construct here

src/vue/src/components/entry/Entry.vue
```diff
@@ -161,12 +175,19 @@
...
+                    Auto save
+                </b-form-checkbox>
                 <dropdown-button
-                    v-if="edit"
+                    v-if="!preferences.auto_save_drafts"
                     :selectedOption="draftPostEntrySetting"
                     :options="{
                         p: {
-                            text: 'Post',
+                            text: 'Submit',
```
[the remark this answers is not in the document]
engelhamer: Then change the button option letters too.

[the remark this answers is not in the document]
engelhamer: However, we do say 'posting entries' everywhere

src/vue/src/components/entry/Entry.vue
engelhamer: That's right and fine behaviour in my opinion.

src/vue/src/components/template/TemplateCard.vue
engelhamer: Doing that

src/vue/src/components/template/TemplateSelection.vue
engelhamer: Template root cannot be a `template` tag and `v-if` cannot be combined with `v-for`, so I think this solution is the supposed setup.

engelhamer (approved):

src/vue/src/components/assets/TextEditor.vue
```diff
@@ -2,14 +2,37 @@
...
+                    name="link"
+                    class="ml-2"
+                />
+            </template>
+            <b-card-body>
+                <br/>
+                <br/>
+                <br/>
+            </b-card-body>
+        </b-card>
```
mjvkeulen, line 35: NIT: Makes sense to move this block into its own component. I can see us tweaking this to more closely resemble the RT editor in the future.

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -198,40 +194,55 @@ export default {
...
 <style lang="sass">
 .timeline-node-circle
-    width: 55px
-    height: 55px
-    border-radius: 50% !important
+    width: 50px
+    height: 50px
+    border-radius: 50%
+    position: absolute
+    left: -2px
```
mjvkeulen, line 201: NIT: Magic number fiesta

mjvkeulen (approved):

engelhamer (approved):

src/vue/src/sass/partials/cards.sass
```diff
@@ -36,7 +36,7 @@
         position: sticky
         @include lg
             z-index: 2 // PDF controls
-            top: 0px
+            top: -30px
```
mjvkeulen, line 39: Comment / variable?

PR #1145, opened 2021-05-11
Improved manage course groups experience 

mjvkeulen:
- Increase the speed at which groups and their members load
- Added search functionality to both groups and if needed, members of the groups themselves
- Improved the edit course groups UI (more in line with the new UI, was due for an update)

image
image
image

# Fixes
- Z#1918
- Z#1917


src/vue/src/components/course/CourseGroupEditor.vue
```diff
@@ -1,138 +1,202 @@
...
+            class="full-width mb-2"
+            placeholder="Search by group name or user name/username"
+            debounce="500"
+        />
+
+        <b-table
+            :fields="[
+                { key: 'name', label: 'Group', tdClass: 'w-75', sortable: true },
+                { key: 'members', sortable: true },
+                'actions'
```
Larspolo, line 46: The filter icons look like they are for the field title on the left. Would prefer it to not be floated to the right, but instead directly after the title, especially with the large fields
image

src/django/VLE/views/group.py
```diff
@@ -187,3 +188,20 @@ def assigned_groups(self, request):
...
+        check_can_view_groups(request.user, course)
+        request.user.check_permission('can_view_course_users', course)
+
+        return response.success({
+            'groups': GroupAndMembersSerializer(
+                GroupAndMembersSerializer.setup_eager_loading(course.group_set),
+                many=True,
+                context={'user': request.user},
+            ).data,
+        })
```
Larspolo, line 207: This call could also be put into the list, with optional param being "get_members", and when it is true, we change the serializer. Less duplicate code & extend the functionality of this call.

src/django/test/test_group.py
```diff
@@ -229,3 +231,39 @@ def test_get_assigned_groups(self):
...
+            data = api.get(
+                self, 'groups/groups_and_members', params={'course_id': self.course.pk}, user=self.course.author)
+
+        check_permission_mock.assert_called_with('can_view_course_users', self.course)
+        check_can_view_groups_mock.assert_called_with(self.course.author, self.course)
+
+        for group_data in data['groups']:
+            group = Group.objects.get(pk=group_data['id'])
+            group_usernames = set(group.participation_set.values_list('user__username', flat=True))
+            assert len(group_data['members']) == len(group_usernames)
```
Larspolo, line 267: Missing assert messages

src/django/test/test_group.py
```diff
@@ -229,3 +231,39 @@ def test_get_assigned_groups(self):
...
+        # Also add a group with no members
+        factory.Group(course=self.course)
+
+        with assert_num_queries_less_than(4, msg='1x fetch groups and 2x prefetch (participations and users'):
+            data = GroupAndMembersSerializer(
+                GroupAndMembersSerializer.setup_eager_loading(self.course.group_set),
+                many=True,
+                context={'user': self.course.author},
+            ).data
+            assert len(data) == 4, '4 groups exist and are serialized'
```
Larspolo, line 253: Missing the "why", why should 4 groups exist and be serialized? Why did you want to test this specifically?

src/vue/src/components/course/CourseGroupEditor.vue
```diff
@@ -1,138 +1,202 @@
...
+
+            <template #cell(members)="data">
+                {{ data.item.members.length }}
+            </template>
+
+            <template #cell(actions)="row">
+                <icon
+                    name="cog"
+                    class="edit-icon"
+                    @click.native="row.toggleDetails"
```
Larspolo, line 95: I lose the overview too quick right now. I think this is caused by any of the following:
- There is no clear border between groups, this makes it unclear when one ends and the other starts
- There are borders within an opened groups, which for me act like thats where a new group starts
- There can be multiple opened groups at the same time, which cause for confusion for which you are editing in currently

I propose (any of) the following:
- Allow for only opened group
- Add a more visual separator between an opened groups and others (think: shadow, thicker border, and/or background colour change)

For reference:
image

src/vue/src/components/course/CourseGroupEditor.vue
```diff
@@ -1,138 +1,202 @@
...
+        <hr/>
+
+        <b-input
+            v-model="filter"
+            class="full-width mb-2"
+            placeholder="Search by group name or user name/username"
+            debounce="500"
+        />
+
+        <b-table
```
Larspolo, line 42: [Q] Why is this table not iterating between white and gray background, and the inner one is?

src/vue/src/components/group/GroupMemberEdit.vue
```diff
@@ -0,0 +1,111 @@
...
+            class="full-width mt-2"
+            placeholder="Search by user name or username"
+            debounce="500"
+        />
+
+        <b-table
+            :fields="[
+                { key: 'full_name', sortable: true },
+                { key: 'username', sortable: true },
+                'action',
```
Larspolo, line 40: [NIT] This action field is missaligned with the outer actions field. Not sure if intended or not, if not, I would prefer aligned action fields
image

src/vue/src/components/group/GroupMemberEdit.vue
```diff
@@ -0,0 +1,111 @@
...
+    },
+    methods: {
+        removeUser (member) {
+            groupAPI.removeMember(
+                this.group.id,
+                member.id,
+                { customSuccessToast: `Removed ${member.full_name} from ${this.group.name}` },
+            )
+                .then(() => {
+                    this.group.members = this.group.members.filter((elem) => elem.id !== member.id)
```
Larspolo, line 93: Think `this.group = response.group` would be preferred, just in case something also also changed in the backend. This would then properly set those changes. Same goes for the other calls

Larspolo (changes requested):
Looking good!

src/vue/src/components/course/CourseGroupEditor.vue
mjvkeulen: This is default bootstrap, I am not sure if this should be changed

src/django/test/test_group.py
mjvkeulen: These are two checks where the assert msg is:  Data matches DB state. But adding them nonetheless

src/vue/src/components/course/CourseGroupEditor.vue
mjvkeulen: Found this more confusing than helpful because of the nested table

src/vue/src/components/group/GroupMemberEdit.vue
mjvkeulen: These are nested tables, which should not align as these are not the same columns, or even part of the same table really. (also note column header spelling)

src/vue/src/components/course/CourseGroupEditor.vue
Larspolo: I still found it confusing, think then others will find as well. What does @engelhamer think about it?

src/vue/src/components/course/CourseGroupEditor.vue
Larspolo: What do you propose?

> these are only expanded after explicit user interaction?

The difference here is that it is properly used when opening multiple groups, however, if the UI then becomes unclear because of proper use, we have a UI problem, not a "well dont do this then" problem. If the solution to unclear UI would be to close other groups, then I suggest doing that automatically. Its not often that a users wants to edit two groups at the same time anyway

src/vue/src/components/group/GroupMemberEdit.vue
Larspolo: Figured, I would still expect it to be at the same location, but this is fine by me then

src/vue/src/components/group/GroupMemberEdit.vue
Larspolo: Sure, can be resolved, leaving it open for Engel to read

src/vue/src/components/course/CourseGroupEditor.vue
mjvkeulen: > Its not often that a users wants to edit two groups at the same time anyway

Searching for a user and looking at which groups they are a part of (including which other group members) seems like a good reason to allow multiple groups to display their members at the same time.

Larspolo: Fair, but then I would like to have a better divide, after opening up multiple, I quickly lose the overview

mjvkeulen: image

Larspolo (approved):

src/vue/src/components/course/CourseGroupEditor.vue
engelhamer: Not bothered by it.

src/vue/src/components/course/CourseGroupEditor.vue
engelhamer: I like it.

src/vue/src/components/group/GroupMemberEdit.vue
engelhamer: Would still look  nicer if they were aligned. This is why I tend to use `b-table-simple`.

engelhamer (approved):

PR #1146, opened 2021-05-11
Improved email style compatibility

engelhamer:
## Description
Emails are now consistent with the rest of the design again. Also some issues related to display are now fixed.

Support:
image

Same, but in dark mode and other client (outlook):
image

Digest:
image

Push notification:
image

Email verification:
image

## Summary of changes
- Fixes Z#1322
- Fixes Z#1762
- Fixes Z#1911
- Email style is now consistent with new UI


src/django/templates/email_templates/digest.html
```diff
@@ -1,49 +1,38 @@
...
+        </td>
+    </tr>
+{% endblock content %}
+
+{% block footer %}
+    <p style="color: grey; text-align: center">
+        Sent by  <a href="https://www.eJournal.app" style="color: grey; text-decoration: none"><span style="color: #216cce">e</span>Journal</a>
+        - Don't like these emails? <a href="{{ email_data.profile_url }}" style="color: #15c; text-decoration: underline">Update your preferences</a>.
+    </p>
+{% endblock footer %}
```
Larspolo, line 37: Could be moved to a component as well

`{% block footer %}  {% include "default_footer.html" %} {% endblock footer %}` or sth

src/vue/src/sass/global.sass
```diff
@@ -147,6 +147,11 @@ hr
             border-radius: 10px 10px 0px 0px
             color: white
         .card-body
+            border: 1px solid $border-color
+            border-top-width: 0px
+            &:last-child
+                border-radius: 0px 0px 10px 10px
+        .card-footer
```
Larspolo, line 154: Why is this here? Maybe left over from another branch?

engelhamer: This is required for the bottom border of the login card. I was apparently removed somewhere, but is encountered through the emails with sign up link etc. That's why I changed it here.

engelhamer:
Email preview:
image

Larspolo (approved):

mjvkeulen (changes requested):
Looking good! 

For me the only remaining item is the logo image, not all clients will render this by default and it's on quite a prominent spot.

If we simply use text instead we won't have to deal with this at all

engelhamer:
But then again that does not look as appealing and will be of a completely
different font?

Op ma 21 jun. 2021 12:46 schreef mjvkeulen ***@***.***>:

mjvkeulen:
To me, that's preferable to our outlook users (mostly teachers) receiving a
notification with a broken image link in a prominent location (logo /
header).

engelhamer:
> To me, that's preferable to our outlook users (mostly teachers) receiving a notification with a broken image link in a prominent location (logo / header).

image

This is the worst case. Most users will just see the normal logo. It is not obtrusive to me and very easy for those who experience it to fix by the click of a button.

All students on gmail will just see a fancy logo instead of the old text.

mjvkeulen (approved):

PR #1147, opened 2021-05-13
Removed the needs for unique emails

Larspolo:
## Description
https://zube.io/ejourn-al/ejournal/c/1924

## Summary of changes
- Fixes Z#1924


src/django/VLE/views/email.py
```diff
@@ -37,7 +37,13 @@ def forgot_password(request):
     identifier, = utils.required_params(request.data, 'identifier')
 
     # We are retrieving the username based on either the username or email
-    user = User.objects.get(Q(email=identifier) | Q(username=identifier))
+    user_qry = User.objects.filter(Q(email=identifier) | Q(username=identifier))
+    if user_qry.count() > 1:
+        return response.bad_request(
+            description='There are multiple accounts attached to this email, please use the username instead.'
+        )
```
mjvkeulen, line 44: This is restrictive, could we not just send an email?

Larspolo: I prefer this, so that you know which accoutn your are resetting, and we dont need to send the username with it

mjvkeulen: What if we just iterate `send_password_recovery_link` for each user?

Larspolo: I know we could, but I prefer this. Iterating over multiple will result in the user receiving multiple emails as well. Then it will be unclear which password the user is resetting when clicking on the email. That would result in the user having to reset all the passwords to the same password

mjvkeulen: Username can be forgotten, the email address is more likely to be remembered. I still prefer iterating the reset link. Leaving this up to @engelhamer as the decider.

Larspolo: How about we make a (imo P5) issue. As this is not just a for loop to be properly implemented?

engelhamer: The username is pretty much always known at this point, as it is the same as the LMS id.
At the same time I don't really like sharing that multiple accounts exist for an address though (reveals _some_ info about the user..)

I would propose to add the username to the email, and indeed send multiple mails in the (rare) case of multiple separate accounts existing:
image

Since both remarks are minor and don't happen often, I would indeed propose to make an issue and accept the status quo.

engelhamer (approved):

engelhamer:
See [Z#1983](https://zube.io/ejourn-al/ejournal/c/1983)

mjvkeulen (approved):

PR #1148, opened 2021-05-15
Added the ability to limit educator journal access to group members only

Larspolo:
## Description
Added the ability for course coordinators to limit TA's to only viewing students from inside their group

Also, the 2 hours didn't go as planned. 5 hours now

## Summary of changes
- Fixes Z#1929


src/django/VLE/models/participation.py
```diff
@@ -7,13 +8,53 @@
...
+            return self.none
+
+        with_correct_role = Q(
+            role__can_view_all_journals=True,
+        )
+        # When there are no groups connecting the participants, we can simply ommit this check
+        if groups != self.NOT_SUPPLIED and groups:
+            with_correct_role.add(Q(
+                role__can_view_journals_in_group=True,
+                groups__in=groups,
```
mjvkeulen, line 33: I believe this will yield a syntax error in Django 3+, groups=groups should work (also on Django 2 I believe)

src/django/VLE/models/participation.py
```diff
@@ -7,13 +8,53 @@
...
+            ), Q.OR)
+
+        if courses is not -1:
```
mjvkeulen: Reuse NOT_SUPPLIED

src/django/VLE/migrations/0087_role_can_view_journals_in_group.py
```diff
@@ -0,0 +1,28 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0086_better_default_demo_instance_name'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='role',
+            name='can_view_journals_in_group',
```
mjvkeulen, line 22: Towards a user, it makes more sense to make the permissions can_view_journals_in_group and can_view_all_journals mutually exclusive. 

To highlight this distinction I would propose to rename the permission to `can_only_view_journals_in_group`

src/django/VLE/models/user.py
```diff
@@ -162,6 +162,24 @@ def check_can_view(self, obj):
...
+    def can_view_user(self, obj):
+        if self == obj:
+            return True
+        if VLE.permissions.is_user_supervisor_of(self, obj):
+            return True
+        if VLE.permissions.is_user_supervisor_of(obj, self):
+            return True
+        if Journal.objects.filter(authors__user__in=[self]).filter(authors__user__in=[obj]).exists():
+            return True
+
```
mjvkeulen, line 198: NIT: Prob add return false iso the implicit return None

src/django/VLE/permissions.py
```diff
@@ -102,11 +100,27 @@ def get_supervisors_of(journal, with_permissions=[]):
...
+    return get_supervisors_of_multiple(
+        groups=journal.authors.values_list('user__groups', flat=True),
+        courses=journal.assignment.courses.all(),
+        with_permissions=with_permissions,
+    )
+
+
+def get_supervisors_of_multiple(groups=None, courses=None, with_permissions=[]):
+    # List conversion is necessary because of a Django bug.
+    # The nested many to many inside a SubQuery causes "IndexError: pop from empty list"
```
mjvkeulen, line 112: Hah I remember this one

src/django/VLE/views/role.py
```diff
@@ -82,7 +82,9 @@ def retrieve(self, request, pk=0):
 
             if user != request.user:
                 # TODO: P Add a permission for this
-                request.user.check_permission('can_view_all_journals', course)
+                # QUESTION: when is this useful? Frontent assignment_id is never set (anymore?)
+                # Could this also be a is_supervisor check?
```
mjvkeulen, line 86: If it is not used frontend, remove all conditional logic here and let us worry about it when we do need it.

src/django/VLE/migrations/0087_role_can_view_journals_in_group.py
Larspolo: How many hours is this change worth to you?

mjvkeulen: A few I guess? What is the expected result of giving a user both the permissions `can_view_journals_in_group` and `can_view_all_journals`?

Larspolo: My guess would be 2 to 3 hours

src/django/VLE/models/user.py
Larspolo: Which do you mean?

mjvkeulen: The block has no final return value

src/django/VLE/models/user.py
```diff
@@ -162,6 +162,40 @@ def check_can_view(self, obj):
...
+        # Owners of the journal
+        return self.has_permission('can_have_journal', obj.assignment)
+
+    def can_view_user(self, obj):
+        if self == obj:
+            return True
+        if VLE.permissions.is_user_supervisor_of(self, obj):
+            return True
+        if VLE.permissions.is_user_supervisor_of(obj, self):
+            return True
```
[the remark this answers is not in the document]
Larspolo: Not introduced in this PR, and dont find it necessarily beter

src/django/VLE/models/role.py
```diff
@@ -96,23 +98,26 @@ def save(self, *args, **kwargs):
         if self.can_edit_course_user_group and not self.can_view_course_users:
             raise ValidationError('A user needs to be able to view course users in order to manage user groups.')
 
-        if self.can_view_all_journals and self.can_have_journal:
+        if self.can_view_all_journals and not self.can_view_journals_in_group:
+            self.can_view_journals_in_group = True
```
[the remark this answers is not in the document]
Larspolo: I still like it this way, feel like whenever someone can view all journal, this can simply be put to true, there is no other way it was intended

src/django/VLE/permissions.py
```diff
@@ -102,11 +100,29 @@ def get_supervisors_of(journal, with_permissions=[]):
...
-    return VLE.models.User.objects.filter(pk__in=VLE.models.Participation.objects.filter(
-        role__can_view_all_journals=True,
-        role__course__in=journal.assignment.courses.all(),
-        **{f'role__{permission}': True for permission in with_permissions},
-    ).values('user')).distinct()
+    return get_supervisors_of_multiple(
+        groups=journal.authors.values_list('user__groups', flat=True),
+        courses=journal.assignment.courses.all(),
+        with_permissions=with_permissions,
+    )
```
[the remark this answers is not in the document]
Larspolo: Good luck!

src/django/VLE/views/assignment.py
```diff
@@ -419,6 +419,7 @@ def add_bonus_points(self, request, *args, **kwargs):
         except Exception:
             csv_reader = csv.reader(csv_file.splitlines())
 
+        journals = Journal.objects.filter().viewable_for(request.user)
```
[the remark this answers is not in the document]
Larspolo: With the empty filter `viewable_for` becomes available to be used

src/django/VLE/serializers.py
```diff
@@ -630,8 +630,10 @@ def get_stats(self, assignment):
         relevant_stat_users_all = list(relevant_stat_users['all'])
         relevant_stat_users_own = list(relevant_stat_users['own'])
 
-        all_j = VLE.models.Journal.objects.filter(assignment=assignment, authors__user__in=relevant_stat_users_all)
-        own_j = VLE.models.Journal.objects.filter(assignment=assignment, authors__user__in=relevant_stat_users_own)
+        all_j = VLE.models.Journal.objects.filter(
+            assignment=assignment, authors__user__in=relevant_stat_users_all)
+        own_j = VLE.models.Journal.objects.filter(
+            assignment=assignment, authors__user__in=relevant_stat_users_own)
```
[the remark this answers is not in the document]
Larspolo: Probably the second

src/django/VLE/views/assignment.py
```diff
@@ -428,7 +429,10 @@ def add_bonus_points(self, request, *args, **kwargs):
                 username, bonus = row
                 bonus = float(bonus)
                 user = User.objects.get(username=str(username))
-                journal = Journal.objects.get(assignment=assignment, authors__user=user)
+                journal = journals.get(
+                    assignment=assignment,
+                    authors__user=user,
+                )
```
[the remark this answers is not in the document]
Larspolo: also probaly from refactory

src/django/VLE/views/grade.py
```diff
@@ -105,7 +107,8 @@ def publish_all_assignment_grades(self, request):
 
         request.user.check_permission('can_publish_grades', assignment)
 
-        journals = Journal.objects.filter(assignment=assignment, unpublished__gt=0).distinct()
+        journals = Journal.objects.filter(
+            assignment=assignment, unpublished__gt=0).viewable_for(request.user).distinct()
```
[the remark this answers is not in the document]
Larspolo: Problem is that we cannot add it without a function call, because we do not have access to the user. But I am not opposed to having to type `viewable_for(...)` for every query. I would even like to see expanding this with `editable_by` and `deletable_by` etc. to reduce query counts, but that is for some other day.

src/django/VLE/views/journal.py
```diff
@@ -80,12 +80,19 @@ def list(self, request):
...
 
         request.user.check_can_view(assignment)
-        if not assignment.is_group_assignment:
-            request.user.check_permission('can_view_all_journals', assignment)
+        journal_query = Journal.objects.filter(assignment=assignment).for_course(course)
+        if not assignment.is_group_assignment or Participation.objects.filter(
+            user=request.user,
+            course=course,
+            role__can_have_journal=False,
+        ).exists():
```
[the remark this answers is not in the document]
Larspolo: Cannot think of anything existing as of now

src/django/VLE/views/journal.py
```diff
@@ -311,6 +321,7 @@ def add_members(self, request, pk):
         journal = Journal.objects.get(pk=pk)
 
         request.user.check_permission('can_edit_assignment', journal.assignment)
+        request.user.check_can_view(journal)
```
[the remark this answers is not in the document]
Larspolo: join: no, the `can_view_journal` function only works for journal authors or teachers / observers

deletion: we could. Added.

src/django/test/test_participation.py
```diff
@@ -150,7 +150,7 @@ def test_participation_serializer(self):
         assignment = factory.Assignment()
         course = assignment.courses.first()
         teacher = assignment.author
-        queries_per_additional_user = 4  # User serializer
+        queries_per_additional_user = 6  # User serializer
```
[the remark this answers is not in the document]
Larspolo: Dont know anymore, changed it back, seemed to work local, lets see what CI thinks about it

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
mjvkeulen: As in: can you not simply do Journal.objects.viewable_for(request.user)?

src/django/VLE/models/participation.py
```diff
@@ -25,7 +25,7 @@ def supervisors_only(self, groups=NOT_SUPPLIED, courses=NOT_SUPPLIED):
         # When there are no courses connecting the participants and supervisor,
         # we can shortcut to something that will never yield true
         if courses is None:
-            return self.none
+            return self.none()
```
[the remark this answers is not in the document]
Larspolo: We want a queryset to be returned, the other returned a function

src/django/VLE/views/journal.py
```diff
@@ -379,6 +391,7 @@ def kick(self, request, pk):
...
```
engelhamer, line 394: I don't think it is needed to have this check at all here (because what really matters is the permission above, and nothing is shared?)

src/django/VLE/views/journal.py
```diff
@@ -379,6 +391,7 @@ def kick(self, request, pk):
         journal = Journal.objects.get(pk=pk)
 
         request.user.check_permission('can_edit_assignment', journal.assignment)
```
engelhamer, line 393: Shouldn't this be `can_manage_journals`?

src/django/VLE/views/journal.py
```diff
@@ -400,6 +413,7 @@ def lock(self, request, pk):
         journal = Journal.objects.get(pk=pk)
 
         request.user.check_can_view(journal.assignment)
+        request.user.check_can_view(journal)
```
engelhamer, line 416: Same, see the permission check directly below. Why do you want this here? Same goes for assignment.

src/django/VLE/views/journal_import_request.py
```diff
@@ -19,6 +19,7 @@ def list(self, request):
         journal_target_id, = utils.required_typed_params(request.query_params, (int, 'journal_target_id'))
 
         journal_target = Journal.objects.get(pk=journal_target_id)
+        request.user.check_can_view(journal_target)
```
engelhamer, line 22: Why?

src/django/VLE/views/journal.py
Larspolo: Without this TA's with this permission could kick users from other journals

src/django/VLE/views/journal.py
Larspolo: Without this TA's with this permission could kick users from other journals. Dont think that is preferred

src/django/VLE/views/journal_import_request.py
Larspolo: So that TA's can view JIR's from journals they cannot see

src/django/VLE/views/journal.py
engelhamer: But how do we decide which groups a group journal is in? If one of the members is in the group? What if they are kicked?

Larspolo: Yeah, either
- a member of the journal,
- or in the assignment with the can_view_all_journals permission,
- or in the assignment with same groups as any of the authors and can_view_journals_in_own_group permission

src/django/test/test_participation.py
```diff
@@ -201,3 +201,10 @@ def test_assignment_participation_serializer(self):
             ).data
 
         assert len(context_post) - len(context_pre) <= queries_per_additional_user
+
+    def test_supervisors_only_no_courses(self):
+        course = factory.Course()
+        assert Participation.objects.supervisors_only(courses=Course.objects.filter(pk=course.pk)).exists(), \
+            'By default the supervisors_only should return the teacher'
```
mjvkeulen, line 208: In line with your assert remarks @Larspolo shouldn't this actually check if the teacher of the course is returned? (course author)

src/vue/src/components/assets/PdfDisplay.vue
```diff
@@ -64,33 +64,15 @@
                     @click.native="downloadLink.click()"
                 />
             </div>
-            <pdf
```
mjvkeulen, line 67: Is this correct? Unexpected on this branch

src/vue/src/components/course/UserRoleConfiguration.vue
```diff
@@ -53,7 +53,13 @@
                         v-for="role in roles"
                         :key="`${role}-${permission}`"
                     >
+                        <template v-if="disabledPermission(role, permission)">
```
mjvkeulen, line 56: image

The tooltip itself is clear, but I would expect the style to be similar to how a teacher cannot remove can edit course roles/details:

image

Selected checkbox which is disabled for input

src/vue/src/main.js
```diff
@@ -174,19 +174,27 @@ new Vue({
                 },
             }
 
-            if (this.$hasPermission('can_view_all_journals', 'assignment', assignment.id)) {
+            console.log(this.$hasPermission('can_view_journals_of_own_groups', 'assignment', assignment.id))
```
mjvkeulen, line 177: Left over

src/vue/src/main.js
```diff
@@ -174,19 +174,27 @@ new Vue({
...
+
+            if (this.$hasPermission('can_view_journals_of_own_groups', 'assignment', assignment.id)) {
                 if (!assignment.is_published && !assignment.is_group_assignment) { // Teacher not published route
-                    route.name = 'AssignmentEditor'
```
mjvkeulen, line 179: Q) What should the route name be if these user does not have `can_edit_assignment`? The behaviour is now slightly different but I can't recall why this was required in the first place.

src/vue/src/views/Assignment.vue
```diff
@@ -1,6 +1,6 @@
 <template>
```
mjvkeulen, line 1: When filtering groups on the assignment page I get the following error:

image

image


Group filter is also not updated because of it

mjvkeulen (changes requested):
`edit_entry` in category.py is missing a journal specific check

src/django/VLE/views/journal.py
mjvkeulen: Does `request.user.check_can_view(journal.assignment)` not always yield True if `request.user.check_can_view(journal)` yields True? So the first check can be omitted?

src/django/VLE/permissions.py
```diff
@@ -102,11 +100,20 @@ def get_supervisors_of(journal, with_permissions=[]):
...
+def get_supervisors_of_multiple(groups=None, courses=None, with_permissions=[]):
+    return VLE.models.User.objects.filter(
+        pk__in=VLE.models.Participation.objects.supervisors_only(
+            courses=courses,
+            groups=groups,
+        ).has_permissions(with_permissions).values_list('user__pk', flat=True).distinct()
+    )
 
 
 def can_edit(user, obj):
```
mjvkeulen, line 119: `_can_edit_entry` needs to filter journals based on viewable journals right?

src/django/VLE/serializers.py
mjvkeulen, line 636: Q) Should TODOs not also be limited to those viewable specifically by the user?

src/django/VLE/models/assignment.py
```diff
@@ -466,11 +466,11 @@ def conflicting_lti_link(self):
     def can_unpublish(self):
         return not (self.has_entries() or self.has_outstanding_jirs())
```
mjvkeulen, line 468: Q) Is this function  `get_all_users` of `models/assignment.py` not also impacted by the stricter filtering requirements introduced in this PR?

src/django/VLE/models/user.py
```diff
@@ -162,6 +162,40 @@ def check_can_view(self, obj):
...
+        return Journal.objects.filter(pk=journal.pk).viewable_by(self).exists()
+
+    def can_view_user(self, user):
+        if self == user:
+            return True
+        if VLE.permissions.is_user_supervisor_of(self, user):
+            return True
+        if VLE.permissions.is_user_supervisor_of(user, self):
+            return True
+        if Journal.objects.filter(authors__user__in=[self]).filter(authors__user__in=[user]).exists():
```
mjvkeulen, line 180: Q) Is the `__in` syntax required here?

src/vue/src/components/course/UserRoleConfiguration.vue
Larspolo: How about changing the disabled teacher input to the newer style?

src/vue/src/main.js
Larspolo: The navigation to home was to redirect users to the homepage if an external user send them a link to an assignment they were not part of. If we didnt redirect them, it showed weird stuff

src/vue/src/views/Assignment.vue
Larspolo: (how) is this introduced in this PR?

mjvkeulen: Haven't looked, but not happening on develop :shrug:

src/django/VLE/models/user.py
Larspolo: Yes, tests fail if I change it to:
`Journal.objects.filter(authors__user=self).filter(authors__user=user).exists()`

src/django/VLE/models/assignment.py
Larspolo: Missed that, added now

src/django/VLE/permissions.py
Larspolo: Its kinda already done by checking if the user is inside the journal

src/vue/src/components/course/UserRoleConfiguration.vue
Larspolo: Changed to my suggestion

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: No
```
>       journals = Journal.objects.viewable_by(request.user)
E       AttributeError: 'JournalManager' object has no attribute 'viewable_by'
```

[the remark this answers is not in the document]
mjvkeulen: Q) Would Journal.objects.all() also work as a starting point? (We just need a query set instance of Journal right)? Not that that's better than an empty filter

src/django/VLE/views/grade.py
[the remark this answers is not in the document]
mjvkeulen: I think this touches on how there are too many alternative paths to check for a permission. Let's leave it as is for now and streamline our permissions once we introduce peer reviews (which will require object-level permissions, e.g. on a per `entry` basis).

src/django/VLE/migrations/0093_role_can_view_journals_in_group.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.2.10 on 2021-05-14 16:41
```
mjvkeulen, line 1: Migration file name needs to also be updated to match the permission

src/django/VLE/models/assignment_participation.py
```diff
@@ -34,6 +36,22 @@ def bulk_create(self, aps, *args, create_missing_journals=True, new_assignment_n
...
+    ):
+        return self.annotate(
+            can_view=Subquery(
+                # Get all the supervisors from the courses that are connected to the journal
+                # Then check if the supplied supervisor is in that set by filtering on the supervisor and getting the PK
+                VLE.models.Participation.objects.supervisors_only(
+                    groups=groups,
+                    courses=courses,
+                ).distinct('user').filter(user=supervisor).values('pk')
+            )
```
mjvkeulen, line 52: Wrap the subquery in an `Exists()` like we do in `annotate_has_unpublished_overall_assessment`, this allows us to work with a boolean as the annotation `can_view` would suggest.

src/django/VLE/models/assignment_participation.py
```diff
@@ -34,6 +36,22 @@ def bulk_create(self, aps, *args, create_missing_journals=True, new_assignment_n
...
+        ).filter(can_view__isnull=False).annotate(can_view=F('pk')).distinct()
```
mjvkeulen, line 53: This can then just become can_view=True

mjvkeulen: Q) Why is this required? `.annotate(can_view=F('pk'))`

src/django/VLE/models/journal.py
```diff
@@ -38,6 +38,23 @@ def for_course(self, course):
...
+        return self.annotate(
+            can_view=Subquery(
+                # Get all the supervisors from the courses that are connected to the journal
+                # Then check if the supplied supervisor is in that set by filtering on the supervisor and getting the PK
+                VLE.models.Participation.objects.supervisors_only(
+                    groups=groups,
+                    courses=courses,
+                ).distinct('user').filter(user=supervisor).values('pk')
+            )
+        ).filter(can_view__isnull=False).annotate(can_view=F('pk')).distinct()
```
mjvkeulen, line 56: Similar thing here

src/django/VLE/models/user.py
mjvkeulen: This confused me so I double-checked.

Without the `__in` syntax, all tests passed!

What does worry me a bit (unrelated sort of), is that I first made the mistake of writing `authors__user=self` 2x instead of once, and all tests still passed!

src/django/VLE/views/role.py
```diff
@@ -60,31 +59,16 @@ def retrieve(self, request, pk=0):
...
+        course = Course.objects.get(pk=course_id)
 
-            request.user.check_can_view(assignment)
+        request.user.check_participation(course)
 
-            if user != request.user:
-                # TODO: P Add a permission for this
-                request.user.check_permission('can_view_all_journals', course)
+        if user != request.user:
+            # TODO: P Is this the right permission
```
mjvkeulen, line 68: So when is someone allowed to request the permissions of another user for a specific course.

@engelhamer?

src/django/VLE/models/assignment_participation.py
Larspolo: Removed again with the `Exists` option its no longer needed, but it was the same system as `allowed_journals` where multiple journals were returned

src/vue/src/components/course/UserRoleConfiguration.vue
mjvkeulen: I preferred the old disabled checkbox (with tooltip), to me that is a bit more clear at first glance that this permission is enabled. However, this is nonblocking for me!

src/django/VLE/models/role.py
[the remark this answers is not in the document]
mjvkeulen: So the only legitimate way someone can differ is by messing with the request/frontend themselves (now that this is enforced frontend (with tooltip and all)). This is why when someone does so I would expect an error not a silent change of the provide value (prone to confusion when refactoring something later on). 

Again nonblocking, but hopefully you agree with the argument ^^

src/vue/src/components/course/UserRoleConfiguration.vue
engelhamer: I find this ambiguous: the grey info icon does not tell the user whether it is active or not without hovering.

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
engelhamer: Indeed `all()`

src/django/VLE/views/role.py
engelhamer: When they need to edit the permissions I guess? So the course edit > users tab.

engelhamer (commented):
For example the `can_manage_course_user_groups` is not very sensible in combination with this: it is possible to place users in another group, and then they are visible again. I think we should make this option only possible if a role has almost no management permissions such as this.

src/vue/src/components/course/UserRoleConfiguration.vue
Larspolo: The tooltip does

mjvkeulen (approved):

src/vue/src/components/course/UserRoleConfiguration.vue
engelhamer: Consider extending this to the `can_have_journal_` combo: that value must be false for those who `can_view_all_journals`, and this would ideally be also displayed here. Then you would have the same symbol but for one it can mean `True` and for the other `False`>

Larspolo: I have no intensions of extending this to the `can_have_journal` combo

engelhamer (approved):

PR #1149, opened 2021-05-16
Fixed make setup

mjvkeulen:
Minor tweaks to get the build script running again


Makefile
```diff
@@ -94,7 +94,7 @@ setup-no-input:
 	(sudo apt-cache show python3.6 | grep "Package: python3.6") || \
 	(sudo add-apt-repository ppa:deadsnakes/ppa -y; sudo apt update) || echo "0"
 
-	sudo apt install nodejs -y
+	sudo apt install nodejs npm -y
```
engelhamer, line 97: Npm should come with nodejs

mjvkeulen: I had to add this to make the setup script work on a clean install

engelhamer: Hm seems like a PPA may be the cause of differences here. Laat time I checked it was not needed. https://www.digitalocean.com/community/tutorials/how-to-install-node-js-on-ubuntu-18-04

Which version do you get from this install?

Larspolo: Ubuntu 20.04 could be the difference here

mjvkeulen: Setup for a fresh Ubuntu 20.04.2, npm version 7.13.0

Larspolo: Problem is that we github acitons stil required dataclasses, I still manually comment is out because of that

mjvkeulen: Cant we just move dataclasses to ci.txt?

requirements/ci.txt
```diff
@@ -20,3 +20,4 @@ pytest-split==0.1.5 # Pytest with GitHub Actions speedup
 pytest-xdist==1.30.0 # Run Pytest in parallel
 deepdiff==5.0.2 # Testing deep difference of dictionary and list combinations
 django-test-migrations==0.3.0
+dataclasses==0.8 # Email sender config
```
Larspolo, line 23: Think this is Python version dependend, production still has an outdated version which should also be upgraded for it to not be needed there anymore

Larspolo: After #1150 is merged, this change is acceptable, but then it needs to be removed here as well

Larspolo: @mjvkeulen this should now be removed as well

Larspolo (changes requested):

requirements/ci.txt
engelhamer: It is removed

Larspolo (approved):
`nodejs` is no longer a valid command after this. I think its changed to just `node`

engelhamer (approved):

PR #1150, opened 2021-05-17
Upgraded Python and Django

mjvkeulen:
Upgraded Django and quite some dependencies

Still some deprecation warnings showing when running tests (in external libs), should be able to filter these but can't quite find the syntax (yet)

- Fixes Z#1123


.github/workflows/build.yml
```diff
@@ -55,10 +55,10 @@ jobs:
           token: ${{ secrets.ACTIONS_TOKEN }}
           submodules: recursive
 
-      - name: Set up Python 3.6.9
+      - name: Set up Python 3.9.5
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6.9
+          python-version: 3.9.5
```
Larspolo, line 61: With this, dataclasses requirement should no longer be necessary anymore. We do have to watch out with deploying this though. All requirements need to be upgraded as well which doesnt happen automatically

Makefile
```diff
@@ -50,7 +50,7 @@ check-linters-back:
 	${venv_activate} \
 	&& flake8 ./src/django \
 	&& ./src/django/manage.py check --fail-level=WARNING \
-	&& isort src/django -c \
+	&& isort src/django/ -rc -c \
```
Larspolo, line 53: [NIT] newest isort makes this argument depricated

src/django/VLE/utils/responses.py
```diff
@@ -146,6 +146,7 @@ def file(file_path, filename):
 
     if settings.ENVIRONMENT == 'LOCAL' or settings.ENVIRONMENT == 'CI_CD':
         try:
+            # NOTE: The file will be closed automatically, so don’t open it with a context manager.
```
Larspolo, line 149: Feel like there is a good story behind this comment

setup.cfg
```diff
@@ -7,7 +7,7 @@ max_parameters_amount = 16
 
 [tool:pytest]
 python_files = test_*.py
-addopts = -n auto -vvl --cov=VLE --cov-report term-missing --cov-config ./setup.cfg
+addopts = -Wa -n auto -vvl --cov=VLE --cov-report term-missing --cov-config ./setup.cfg
```
Larspolo, line 10: [Q] What is this change?

Larspolo (commented):
Waiting for models merge before the actual review

setup.cfg
mjvkeulen: Display all warnings (e.g. deprecation warnings). Depending on the version of pytest this is the default as well I think.

Makefile
mjvkeulen: This currently throws an error. Let's address this once we update isort?

I deliberately chose not to update isort in this PR due to the size of the diff it would generate. Best suited for its own dedicated PR.

Larspolo (approved):

engelhamer (approved):

PR #1154, opened 2021-05-23
Select template via name in addition to the add button

mjvkeulen:
I feel this is more intuitive, first search criteria is the template name. It makes sense that one can simply click on said template name to actually select it.

image


engelhamer:
I prefer the more explicit previous implementation

Larspolo:
Same, although I will agree that clicking on the template name to add would be an improvement as well. We could stil include that, but revert the actions -> action change?

mjvkeulen:
> Same, although I will agree that clicking on the template name to add would be an improvement as well. We could stil include that, but revert the actions -> action chang

This works for me, I've pushed the change

Larspolo (approved):

engelhamer (approved):

PR #1159, opened 2021-05-30
Split assignment copy to more comprehensible functions

Larspolo:
## Description
PR useful for copying existing formats to assignments. Although not full (e.g. api endpoint) support yet


mjvkeulen (approved):

src/django/VLE/views/assignment.py
```diff
@@ -28,6 +28,102 @@
 from VLE.utils.file_handling import copy_assignment_related_rt_files
 
 
+def create_a_to_be_copied_assignment(source, author, course):
```
engelhamer, line 31: Naming confusese me here. This copies the assignment already, and does not create a 'to be copied' assigmnent? The 'to be copied' assignment would be the source.

src/django/VLE/views/assignment.py
```diff
@@ -494,94 +590,29 @@ def copy(self, request, pk):
...
-        assignment.active_lti_id = None
-        assignment.lti_id_set = []
-        set_assignment_dates(assignment, months_offset)
-
-        # One to one fields needs to be updated before save else we would have a duplicate key
-        assignment.format = format
-        assignment.save()
-
+        new_assignment = create_a_to_be_copied_assignment(assignment_source, request.user, course)
         assignment_source = Assignment.objects.get(pk=assignment_source_id)
```
engelhamer, line 602: Why is it needed to have this again here?

src/django/VLE/views/assignment.py
Larspolo: I indeed had trouble with the assignment name. The assignment gets copied partially, not yet the whole format
What do you suggest?

src/django/VLE/views/assignment.py
Larspolo: Dont know, maybe @mjvkeulen knows

src/django/VLE/views/assignment.py
Larspolo: To reduce complexity (and overview) of the function. Especially with LTI mixed in as well, it was too unclear what was happening where, this was an attempt at clarifying it

src/django/VLE/views/assignment.py
mjvkeulen: `assignment_source` is used as the template for `new_assignment`, some of its values are changed and is saved to become `new_assignment`. Whereas this process maintains the reference to the `assignment_source` its values are no longer correct (e.g. a new `pk` as its now effectively `new_assignment`).

engelhamer: Worth a comment as it is unclear.

engelhamer (changes requested):

engelhamer (approved):

PR #1162, opened 2021-06-04
Remove django prometheus from project

mjvkeulen:
Was causing issues when upgrading to python3.9


engelhamer (approved):

Larspolo (approved):

Larspolo (changes requested):
Config module change should not be changed here

Larspolo (approved):

PR #1163, opened 2021-06-05
Added rubrics and the ability to provide a journal with an overall assessment

mjvkeulen:
# Rubrics 
Rubrics can be created at the assignment level in line with the other assignment editor objects. Rubrics can be imported from other assignments and are copied alongside an assignment's format.

image

image

A valid rubric consists of at least one criterion with two or more levels (one for full score, and one for the minimum score).

It is possible to create a Rubric with a varied number of levels per criteria. E.g. a criterion with a pass or fail level, and a criterion with a 5-star grading scale.

In order to speed up the grading process, a Teacher can provide initial feedback snippets on a per-level basis. These are used to provide the initial feedback for the criterion.

Preset DB contains a fully functional rubric for "Written Assignments", this in line with our goal to slowly bring more high fidelity objects to our test env.

image

# Overall journal assessment

This PR also introduces the ability to provide a journal with an overall assessment. An overall assessment can be combined with our standard workflow where each individual part of the journal is evaluated/graded, or used on its own. 

An assessment works with drafts and autosave. Once an assessment is published, a copy will always be available as part of the assessment history. The assessment history can also be used to restore a previous version.

An overall assessment can be started via the journal grading menu:

image

Once started, an assessment draft is created and any overall feedback can be supplied.

image

It is possible to add a rubric to an assessment. Or to configure a default overall assessment rubric at the assignment level in order to speed up the grading process.

image

An overall assessment provides a final grade for the journal. By default, this is the sum of any individual parts (graded entries, bonus points, assessment rubric). This grade is dynamic, so if a teacher decides to update an individual entry grade, even after having published an overall assessment, the assessment grade will update.

image

However, it is also possible to explicitly override the default and provide a manual grade.

image

In this scenario the assessor is sufficiently warned that this grade is an override, and any changes to any of the scored parts are no longer dynamic.


src/vue/src/components/journal/JournalNonStudent.vue
```diff
@@ -44,7 +44,7 @@
                     :bonusPoints="journal.bonus_points"
                 />
             </load-wrapper>
-        </template>
+        </div>
 
         <template #right>
```
Larspolo, line 49: Missing this right one

mjvkeulen: There is guaranteed content available for the slot when component is created

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -6,17 +6,73 @@
                 End of assignment
             </h2>
         </template>
+
+        <load-wrapper :loading="loading">
```
Larspolo, line 10: This looks great, and is a good place to have the overall assessment. However, I do think some extra indication for both students and teachers would be nice to hint that the overall assessment is placed here

Larspolo (changes requested):

src/vue/src/components/rubric/edit/LevelEdit.vue
```diff
@@ -0,0 +1,130 @@
...
+                    class="trash-icon float-right mt-3 ml-2"
+                    :class="{ 'input-disabled': criterion.is_in_use }"
+                    name="trash"
+                    @click.native="removeLevel()"
+                />
+            </div>
+
+            <icon
+                class="edit-icon float-right mt-3"
+                name="comments"
```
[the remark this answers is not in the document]
mjvkeulen: Clicking on the icon once yields this:

image

So that is not sufficiently self-explanatory?

src/vue/src/components/rubric/CriterionRow.vue
```diff
@@ -0,0 +1,233 @@
...
+                :description="scoreDescription"
+            >
+                <b-form-input
+                    v-model="criterionSubmission.points"
+                    class="score-input"
+                    type="number"
+                    size="2"
+                    placeholder="-"
+                    :formatter="floatFormatter"
+                    :disabled="!grading || !criterionSubmission.level_id"
```
Larspolo, line 70: Would prefer it to switch to just text iso readonly. Especially for students this now looks a bit weird.

src/vue/src/components/assessment/Assessment.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+    <div>
+        <assessment-rubric
+            :assessment="assessment"
+            :edit="edit"
+            :busyRequests="busyRequests"
+            :assignment="assignment"
+        />
+
+        <b-form-group label="Feedback">
```
Larspolo, line 10: [NIT] "Overall feedback" or "Extra feedback" might be a better title

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -0,0 +1,356 @@
...
+        addCriterion () {
+            this.rubric.criteria.push({
+                id: this.criteriaMinId - 1,
+                name: `Criteria ${this.rubric.criteria.length + 1}`,
+                description: '',
+                long_description: '',
+                location: this.rubric.criteria.length,
+                levels: [
+                    {
+                        name: 'Full marks',
```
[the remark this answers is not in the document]
mjvkeulen: Full marks for a single criterion are not related to the maximum number of points available for the assignment.

src/vue/src/components/journal/JournalAssessment.vue
```diff
@@ -0,0 +1,370 @@
...
+    <b-button
+        v-if="$hasPermission('can_grade')"
+        variant="link"
+        class="blue-button"
+        @click="showOverallAssessment"
+    >
+        <icon name="check"/>
+        Overall Assessment
+
+        <b-modal
```
Larspolo, line 11: I think redirection this to the end node iso a modal would feel more natural. Even if it comes with the downside of modals being able to be wider.
The same width will be shown in the rubric editor, for student, and for teachers in the "end of assignment" node anyway.
We already have an issue open to improve the width restraint, which could imo be bumped after this PR

src/vue/src/components/journal/JournalAssessment.vue
```diff
@@ -0,0 +1,370 @@
...
+                    v-if="assessment"
+                    :class="{
+                        'input-disabled': busy || loading || blankDraft,
+                        'red-button': !assessment.is_draft,
+                        'orange-button': assessment.is_draft ,
+                    }"
+                    @click="deleteAssessment"
+                >
+                    <icon :name="assessment.is_draft ? 'undo': 'trash'"/>
+                    {{ assessment.is_draft ? nonDraft ? 'Restore' : 'Reset' : 'Delete' }}
```
Larspolo, line 110: [NIT] When it says "Restore", it shortly switches to "Reset" during loading, and only then changes to "Delete"

Larspolo, line 110: [Q] when is "Reset" shown?

Larspolo, line 110: [NIT] I still find this naming a bit confusing, but I also dont have the golden solution. How does changing "restore" to "Reset Draft" sound? Or "Remove pending changes"?

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,166 @@
...
+                0
+                + this.rubricPoints
+                + this.bonusPoints
+                + this.entryPoints
+            )
+        },
+        overridingDefaultGrade () { return this.edit && this.assessment.points !== null },
+        gradeOverrideTooltip () {
+            if (this.overridingDefaultGrade) {
+                return 'Back to default.'
```
Larspolo, line 124: [NIT] "Back to automatically determining grade."

src/django/VLE/utils/generic_utils.py
```diff
@@ -152,5 +153,7 @@ def gen_url(node=None, journal=None, assignment=None, course=None, user=None):
             url += '/Journal/{}'.format(journal.pk)
             if node:
                 url += '?nID={}'.format(node.pk)
+            elif assessment:
```
Larspolo, line 156: Think you are missing something here. this now checks if the import is not None. Also add a test for this ;)

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: - Teacher publishes an overall assessment -> node is selected and assessment is displayed
- Student receives a notification, by default this is push, via the email button link the student is redirected to their journal, and the end node is selected displaying the overall assessment.

Both teacher and student are introduced to the location of the overall assessment quite naturally. What would you propose differently?

src/django/VLE/utils/file_handling.py
```diff
@@ -178,7 +182,7 @@ def get_temp_files_from_rich_text(rich_text):
     return get_files_from_rich_text(rich_text).filter(is_temp=True)
 
 
-def copy_assignment_related_rt_files(rich_text, user, assignment=None, category=None):
+def copy_rt_files(rich_text, user, assignment=None, category=None, assessment=None):
```
Larspolo, line 185: [NIT] Found the original name better. It cannot handle profile pictures or other non assignment related files if there are any

src/django/VLE/utils/grading.py
```diff
@@ -16,6 +16,8 @@ def publish_all_journal_grades(journal, publisher):
     - journal: the journal in question
     - publisher: the publisher of the grade
     """
+    Assessment.objects.unpublished_assessment_for_journal(journal).update(is_published=True)
```
Larspolo, line 19: [NIT] really nit, but would expect this after the other grades are published

src/django/VLE/utils/import_utils.py
```diff
@@ -272,6 +273,43 @@ def bulk_import_assignment_categories(source_assignment, target_assignment, auth
     return zip(source_categories, target_categories)
 
 
+def copy_rubrics(rubrics, target_assignment):
+    """
+    Copies a set of rubrics to the target assignment, including any nested criteria and levels.
+
+    NOTE: Prefetch the criteria and levels for performance
+    """
+    criteria = []  # [[C1, C2], [C3]]
```
Larspolo, line 282: left over code?

src/django/VLE/views/assessment.py
```diff
@@ -0,0 +1,205 @@
...
+        journal=assessment.journal,
+        sent=False,
+        type=Notification.NEW_OVERALL_JOURNAL_ASSESSMENT,
+    ).delete()
+
+
+def handle_draft_to_non_draft(draft, author, publish):
+    if publish:
+        generate_new_assessment_notifications.apply_async(args=[draft.pk], countdown=settings.WEBSERVER_TIMEOUT)
+        grading.task_journal_status_to_LMS.apply_async(args=[draft.journal.pk], countdown=settings.WEBSERVER_TIMEOUT)
```
Larspolo, line 28: This needs to go after the rest in case that crashes, we dont want the notifications to be send anyway

src/django/test/factory/content.py
```diff
@@ -7,17 +7,26 @@
...
+
 IMG_FILE_PATH = '../vue/public/journal-view.png'
 PDF_FILE_PATH = './VLE/management/commands/dummy.pdf'
 
 if not (os.path.exists(IMG_FILE_PATH)):
     IMG_FILE_PATH = './src/vue/public/journal-view.png'
     PDF_FILE_PATH = './src/django/VLE/management/commands/dummy.pdf'
 
+# For staging env just use the commands folder
+if not (os.path.exists(IMG_FILE_PATH)):
```
Larspolo, line 26: This check should include `for staging env` as well then?

src/django/VLE/models/preferences.py
```diff
@@ -43,6 +43,11 @@ class Preferences(CreateUpdateModel):
         choices=FREQUENCIES,
         default=PUSH,
     )
+    new_overall_journal_assessment_notifications = models.TextField(
+        max_length=1,
+        choices=FREQUENCIES,
+        default=PUSH,
```
Larspolo, line 49: [D(iscussion)] Do we wish to change the `new grade` to daily now? Because this is now more important

src/django/VLE/models/rubric.py
```diff
@@ -0,0 +1,225 @@
...
+    for instance_data in data:
+        name, = generic_utils.required_typed_params(instance_data, (str, 'name'))
+
+        if name in names:
+            raise ValidationError(f'{instance_label} name: {name} is not unique.')
+
+        names.add(name)
+
+
+def validate_fks(data, attr, expected_fk, instance_label):
```
Larspolo, line 37: [NIT] Feels like a utils functions to me (also unique name and locations (for templates)). Would like a docstring as well. When is this useful?

src/django/VLE/models/rubric.py
```diff
@@ -0,0 +1,225 @@
...
+    for instance_data in data:
+        actual_fk, = generic_utils.required_typed_params(instance_data, (int, attr))
+        if actual_fk > 0 and actual_fk != expected_fk:
+            raise ValidationError(
+                f'{instance_label} is unexpectedly linked to a different {attr}. `{actual_fk}` iso `{expected_fk}`.')
```
Larspolo, line 42: [NIT] if the user sees this error its a bit weird.

src/django/VLE/settings/base.py
```diff
@@ -119,6 +119,8 @@ class Emails:
 # Names we have encountered used for test students
 LTI_TEST_STUDENT_FULL_NAMES = {'Test student'}
 
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
```
Larspolo, line 122: [Q] I cannot find this anywhere in the code, when is this useful?

src/vue/src/components/journal/JournalAssessment.vue
```diff
@@ -0,0 +1,370 @@
...
+            ref="journalAssessmentModal"
+            size="xl"
+            noEnforceFocus
+            footer-class="d-block"
+        >
+            <template #modal-header>
+                <h5 class="modal-title">
+                    Overall Assessment
+                    <icon
+                        v-b-tooltip:hover="'Overall assessments trump any individual entry grades'"
```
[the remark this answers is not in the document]
mjvkeulen: This is a bit outdated, now the grade is reactive unless overwritten explicitly by the user. See the description

Larspolo (changes requested):

src/vue/src/components/rubric/edit/RubricEdit.vue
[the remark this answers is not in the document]
mjvkeulen: min (10, assignmentMax)

src/vue/src/components/journal/JournalAssessment.vue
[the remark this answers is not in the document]
mjvkeulen: will remove tooltip

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Will change end of assignment title to assessment if present

src/vue/src/components/rubric/CriterionRow.vue
mjvkeulen: Will switch to text for both student and teacher in read mode

src/vue/src/components/journal/JournalAssessment.vue
mjvkeulen: Decided to move to zube issue in order to find a better solution

src/vue/src/components/journal/JournalAssessment.vue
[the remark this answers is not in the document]
mjvkeulen: Updated tooltip

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        # As student I should receive only published non archived assessments for a journal
+        with mock.patch('VLE.models.User.can_view') as can_view_mock:
+            api.get(self, 'assessments', params={'journal_id': self.journal.pk}, user=self.journal.author)
+            can_view_mock.assert_called_with(self.journal)
+        resp = api.get(self, 'assessments', params={'journal_id': self.journal.pk}, user=self.journal.author)
+
+        serialized_assessments_pks = {data['id'] for data in resp['assessments']}
+        assert self.assessment.pk in serialized_assessments_pks, 'A student should receive published assessments'
+        assert self.assessment_archived.pk not in serialized_assessments_pks, \
+            'A student should not see archived assessments unless specifically requested'
```
Larspolo, line 229: [Q] Is the student able to see archived assessments? How? Didnt see it frontend at least

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        assert self.assessment_draft.pk not in serialized_assessments_pks, 'A student not receive drafts'
```
Larspolo, line 230: A student should not receive drafted assessments

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+
+        # The ability to see archived assessments is governed by `can_view_grade_history`
+        params = {'journal_id': self.journal.pk, 'get_history': True}
+        with mock.patch('VLE.models.User.check_permission') as check_permission_mock:
+            resp = api.get(self, 'assessments', params=params, user=self.assignment.author)
+            serialized_assessments_pks = {data['id'] for data in resp['assessments']}
+            assert self.assessment_archived.pk in serialized_assessments_pks
```
Larspolo, line 237: Missing message

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+    def test_assessment_create(self):
+        journal = factory.Journal(assignment=self.assignment, entries__n=0)
+        params = {'journal_id': journal.pk}
+
+        with mock.patch('VLE.models.User.check_permission') as check_permission_mock:
+            resp = api.create(self, 'assessments', params=params, user=self.assignment.author)
+            check_permission_mock.assert_called_with('can_grade', self.assignment)
+
+        assessment_draft = Assessment.objects.get(pk=resp['assessment']['id'])
+        assert resp['assessment'], 'A created assessment is returned'
```
Larspolo, line 263: "When creating an assessment via de API, it should respond with the created assessment."

Although the statement above is more checking that then this statement is. This just checks if there is a response

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        api.create(self, 'assessments', params=params, user=self.assignment.author, status=400)
+
+        # A journal or assessment id is required to create an assessment
+        api.create(self, 'assessments', params={}, user=self.assignment.author, status=400)
+
+        # When a fellow teacher has already started a draft (data was stale) ask the user to reload
+        teacher2 = factory.Teacher(add_to=self.assignment)
+        params = {'journal_id': journal.pk}
+        resp = api.create(self, 'assessments', params=params, user=teacher2, status=400)
+        assert self.assignment.author.full_name in resp['description']
```
Larspolo, line 287: "The author that already started their assessment should be given in the response"

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        archived_assessment = factory.Assessment(is_archived=True, journal=journal)
+        pre_archived_assessment_data = AssessmentSerializer(archived_assessment).data
+
+        params = {'assessment_id': archived_assessment.pk}
+        resp = api.create(self, 'assessments', params=params, user=self.assignment.author)
+
+        archived_assessment.refresh_from_db()
+        post_archived_assessment_data = AssessmentSerializer(archived_assessment).data
+        assert equal_models(pre_archived_assessment_data, post_archived_assessment_data), \
+            'When starting a draft based via the assessment histroy, the source assessment should not be modified'
```
Larspolo, line 338: history*

src/django/test/test_assignment.py
```diff
@@ -774,6 +793,41 @@ def test_assignment_import(self):
...
+            RubricSerializer.setup_eager_loading(assignment_copy.rubrics.all()),
+            many=True,
+            context={'user': assignment_copy.author},
+        ).data
+
+        assert equal_models(
+            source_rubric_data,
+            target_rubric_data,
+            ignore_keys=['creation_date', 'id', 'assignment', 'criterion', 'level', 'rubric'],
+        )
```
Larspolo, line 829: "The source and target rubric should contain the same information after copy"

src/django/test/test_assignment.py
```diff
@@ -825,6 +879,52 @@ def test_assignment_copy_templates_and_categories(self):
...
+        source_assignment.assessment_model.default_rubric = rubric
+        source_assignment.assessment_model.save()
+
+        resp = api.post(self, 'assignments/{}/copy'.format(source_assignment.pk), params={
+            'course_id': target_course.pk,
+            'months_offset': 0,
+        }, user=self.teacher)
+
+        assignment_copy = Assignment.objects.get(pk=resp['assignment_id'])
+        assert assignment_copy.rubrics.count() == source_assignment.rubrics.count()
```
Larspolo, line 897: "All rubrics should be copied during an assignment copy"

src/django/test/test_notification.py
```diff
@@ -504,3 +504,19 @@ def test_jir_notification(self):
...
+    @override_settings(CELERY_TASK_ALWAYS_EAGER=True)
+    def test_assessment_notification_content(self):
+        journal = factory.Journal(entries__n=0)
+        assignment = journal.assignment
+        assessment = factory.Assessment(journal=journal)
+
+        generate_new_assessment_notifications(assessment.pk)
+        notification = Notification.objects.get(assessment=assessment)
+
+        assert '?assessment' in notification.url, \
```
Larspolo, line 517: Needs check that this doesnt exist on notifications that are not about assessment, but still contain journal

src/django/test/test_rubric.py
```diff
@@ -0,0 +1,488 @@
...
+            'Updating a rubric without making changes to the data should yield the same rubric'
+
+        params['pk'] = self.full_rubric.pk
+
+        new_rubric_params = factory.RubricCreationParams(assignment_id=self.assignment.pk, n_criteria=3)
+
+        updated_data = deepcopy(params)
+        updated_data['name'] = 'Some new name'
+        updated_data['criteria'][1]['name'] = 'Some new name'
+        updated_data['criteria'][1]['levels'][1]['name'] = 'Some new name'
```
Larspolo, line 304: Would prefer these names to be different, just in case the name gets wrongly changed

src/django/test/test_rubric.py
```diff
@@ -0,0 +1,488 @@
...
+                delete_criterion_params['criteria'].remove(criterion_data)
+                break
+        else:
+            assert False, 'Test must remove a criterion from the data'
+
+        resp = api.patch(self, 'rubrics', params=delete_criterion_params, user=self.assignment.author, status=400)
+        assert 'cannot remove' in resp['description'], \
+            'It is not possible to remove a criterion which is used in an assessment'
+        assert Criterion.objects.filter(pk=criterion_submission.criterion.pk).exists(), \
+            'It is not possible to remove a criterion which is used in an assessment'
```
Larspolo, line 350: "The criterion should still exists when trying to remove it while it is in use"

src/django/test/test_rubric.py
```diff
@@ -0,0 +1,488 @@
...
+                        for i, level_data in enumerate(criterion_data['levels']):
+                            level_data['location'] = i
+                        break
+        assert deleted, 'Test must remove a level from the level data'
+
+        resp = api.patch(self, 'rubrics', params=delete_level_params, user=self.assignment.author, status=400)
+        assert 'cannot remove' in resp['description'], \
+            'It is not possible to remove a level which is used in an assessment'
+        assert Level.objects.filter(pk=criterion_submission.level.pk).exists(), \
+            'It is not possible to remove a level which is used in an assessment'
```
Larspolo, line 380: "The level should still exists when trying to remove it while it is in use"

src/django/test/test_rubric.py
```diff
@@ -0,0 +1,488 @@
...
+        rubric3 = factory.Rubric(assignment__author=teacher, assignment=rubric2.assignment)
+        rubric_from_assignment_where_teacher_is_student = factory.Rubric()
+        student_role = Role.objects.get(
+            name='Student', course__assignment=rubric_from_assignment_where_teacher_is_student.assignment)
+        factory.Participation(user=teacher, role=student_role)
+
+        resp = api.post(self, 'rubrics/importable', params={'assignment_id': source_assignment.pk}, user=teacher)
+
+        assert {rubric['id'] for rubric in resp['rubrics']} == {rubric.pk, rubric2.pk, rubric3.pk}, \
+            'All rubrics are serialized where the used can edit the assignment'
```
Larspolo, line 488: This does not check if rubrics are returned multiple times. Would go for a length check as well

Larspolo (changes requested):
Dont mind all the test comments too much. I wanted to test if I could only review the test code based on the assert statements alone, managed pretty well

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Also added show overall assessment to journal details

src/vue/src/components/journal/JournalAssessment.vue
mjvkeulen: When working on a draft and no older version is available (from the same working session, so excluding history)

src/django/VLE/utils/generic_utils.py
mjvkeulen: What do you mean?

This works and is tested in `test_assessment_notification_content`

src/django/VLE/utils/file_handling.py
mjvkeulen: profile picture is not a `rich_text` file, this can be easily extended and works beyond specific assignment scope now

src/django/VLE/views/assessment.py
mjvkeulen: I moved it, but this a downside of how we approach transactions.

Could just be 
```
transaction.on_commit(lambda: generate_new_assessment_notifications.delay(draft.pk))
transaction.on_commit(lambda: grading.task_journal_status_to_LMS.delay(draft.journal.pk))
```

But this would require us to inherit from `TransactionTestCase` for tests which uses stuff like `on_commit`. Which truncates all tables after a test, which might not be compatible with `setUpTestData(cls)`.

src/django/test/factory/content.py
mjvkeulen: Could be added, it's a fallthrough check tho. If no file found look here, moved to elif to skip unneeded os path checks.

src/django/VLE/models/preferences.py
mjvkeulen: I would not change anything since you don't know which model will be used in advance.

src/django/VLE/models/rubric.py
mjvkeulen: Has to intentionally mess with data

src/django/VLE/settings/base.py
mjvkeulen: Is migration setting from Django 2 to 3 (no behavioural change, just an explicit definition of what our pks are)

src/django/VLE/views/grade.py
```diff
@@ -105,7 +106,17 @@ def publish_all_assignment_grades(self, request):
 
         request.user.check_permission('can_publish_grades', assignment)
 
-        journals = Journal.objects.filter(assignment=assignment, unpublished__gt=0).distinct()
+        journals = (
+            Journal.all_objects  # Default manager has no access to `annotate_has_unpublished_overall_assessment`
```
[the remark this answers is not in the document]
mjvkeulen: Because our default manager is no created from `JournalQuerySet`

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        assert not RubricSubmission.objects.filter(assessment=self.assessment_draft).exists(), \
+            'A draft is not linked to a rubric submission by default'
+        assert not self.assessment_draft.feedback, 'A draft contains no feedback by default'
+
+        assert self.assessment_archived.is_archived, 'Setting archived works'
+        assert not self.assessment_archived.rubric_submission, 'No rubric submission is generated if so set'
+
+    def test_active_assessment_for_journal_query(self):
+        assert Assessment.objects.active_assessment_for_journal(journal=self.journal).get() == self.assessment, \
+            '''
```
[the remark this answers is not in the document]
mjvkeulen: No active assessment is generated for the journal, which is quite explicitly defined here no?

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+
+        assert self.journal.assessment == self.assessment, \
+            'Journal has an assessment property which returns the active assessment'
+
+    def test_unpublished_assessment_for_journal_query(self):
+        journal = factory.Journal(assignment=self.assignment, entries__n=0)
+        factory.Assessment(as_empty_draft=True, journal=journal)
+        assessment = factory.Assessment(journal=journal, is_published=False)
+        assert Assessment.objects.unpublished_assessment_for_journal(journal=journal).get() == assessment, \
+            '''
```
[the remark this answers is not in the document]
mjvkeulen: Ditto here

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+
+        assessment_history_queried = list(Assessment.objects.assessment_history_for_journal(journal))
+
+        assert all(expected == queried for expected, queried in zip_equal(
+            assessment_history, assessment_history_queried,
+        )), 'Assessment history returns all non draft assessments in order of oldest to newest'
+
+        assessment_history.sort(key=lambda x: x.update_date)
+        assert assessment_history == assessment_history_queried
+        assert assessment_history[0].update_date < assessment_history[1].update_date
```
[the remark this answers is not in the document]
mjvkeulen: These all check the same thing: `Assessment history returns all non draft assessments in order of oldest to newest`

These just validate it using python sort as well

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        # It should not be possibly to publish a draft
+        assessment.is_published = True
+        self.assertRaises(VLEProgrammingError, assessment.save)
+
+        # It should not be possible to archive a draft
+        assessment.is_published = False
+        assessment.is_archived = True
+        self.assertRaises(VLEProgrammingError, assessment.save)
+
+        # Only published assessments can be archived
```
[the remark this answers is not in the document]
mjvkeulen: I believe its correct: Only when an assessment is published can it be archived (both True)

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        assessment.save()
+
+    def test_can_edit_assessment(self):
+        teacher = self.assessment.author
+        student = self.journal.author
+        journal = factory.Journal(assignment=self.assignment, entries__n=0)
+        assessment = factory.Assessment(journal=journal, rubric_submission__rubric=self.full_rubric)
+        rubric_submission = assessment.rubric_submission
+        criterion_submission = rubric_submission.criterionsubmission_set.first()
+        assert criterion_submission
```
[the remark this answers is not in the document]
mjvkeulen: Kinda small seeing `first()` and being part of the init block (assert is still part of the init block)

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+
+        assessment.is_draft = True
+        assessment.is_published = False
+        assessment.is_archived = False
+        assessment.save()
+        assert teacher.can_edit(assessment), 'Assigned teachers can always edit drafts'
+        assert teacher.can_edit(rubric_submission), 'Rubric submissions can only be edited for assessment drafts'
+        assert teacher.can_edit(criterion_submission), 'Criterion submissions can only be edited for assessment drafts'
+        assert not student.can_edit(assessment), 'Student cannot edit the assessment'
+        assert not student.can_edit(rubric_submission), 'Student cannot edit the assessment'
```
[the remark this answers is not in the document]
mjvkeulen: This was intentional as its tied to the assessment

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        assert not student.can_edit(criterion_submission), 'Student cannot edit the assessment'
```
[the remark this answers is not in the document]
mjvkeulen: Likewise

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,533 @@
...
+        rubric_submission = assessment.rubric_submission
+        criterion_submission = rubric_submission.criterionsubmission_set.first()
+        assert criterion_submission
+
+        assessment.is_draft = True
+        assessment.is_published = False
+        assessment.is_archived = False
+        assessment.save()
+        assert teacher.can_edit(assessment), 'Assigned teachers can always edit drafts'
+        assert teacher.can_edit(rubric_submission), 'Rubric submissions can only be edited for assessment drafts'
```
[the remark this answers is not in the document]
mjvkeulen: Sorry split the message?

src/django/test/test_assessment.py
mjvkeulen: If given the permission grade history

src/django/test/test_assessment.py
mjvkeulen: The block has the comment: `The ability to see archived assessments is governed by `can_view_grade_history`` Will repeat it

src/django/test/test_assessment.py
mjvkeulen: That combination is fine no? We know the assessment is created and serialized since we got assessment the line above

mjvkeulen: I believe context prefix is good: `When creating an assessment via de API` but this generally puts mosts asserts over two lines. Do not forget the simplicity of the block: create is just called in a familar pattern. Or the name of the test `test_assessment_create`. 

These are all viable context

src/django/test/test_assessment.py
mjvkeulen: (Again the comment above the block could basically be repeated in the asserts)

src/django/test/test_assignment.py
mjvkeulen: Also heavily implicated in the format of this assert tbh

src/django/test/test_rubric.py
mjvkeulen: Different names would not help here?

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,166 @@
...
+
+        <tr
+            class="final-grade-row"
+            :class="{ 'final-grade-seperator': displayFinalGradeSeperator }"
+        >
+            <td>
+                <b>Final Grade: </b>
+                <icon
+                    v-if="edit"
+                    v-b-tooltip="gradeOverrideTooltip"
```
Larspolo, line 38: It would also be nice to have a visual indication that the grade is manually overwritten in the read mode. Some info icon with this tooltip would do.
image

mjvkeulen: Done

src/django/VLE/models/rubric.py
```diff
@@ -0,0 +1,225 @@
...
+            for criterion_data in criteria_data:
+                levels_data, = generic_utils.required_params(criterion_data, 'levels')
+
+                validate_locations(levels_data, 'Level')
+                validate_unique_names(levels_data, 'Level')
+                if old:
+                    existing_level_data_criterion_pks = existing_level_data_criterion_pks.union({
+                        level_data.get('criterion')
+                        for level_data in levels_data
+                        if level_data.get('criterion') > 0
```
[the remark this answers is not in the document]
Larspolo: @mjvkeulen This now doesnt throw a 500 anymore, but it throws a scilent 400 error (no toast):
image

NOTE: between step 1 and step 2 also save the rubric, and start editing it in step 2

src/vue/src/components/rubric/edit/LevelEdit.vue
```diff
@@ -0,0 +1,130 @@
...
+            return Math.min(...this.criterion.levels.map((level) => level.id), 0)
+        },
+    },
+    methods: {
+        /* Adds a level after (to the right of) the current level */
+        addLevel () {
+            const afterLevel = this.level
+
+            this.criterion.levels.splice(afterLevel.location + 1, 0, {
+                name: `Less than ${afterLevel.name}`,
```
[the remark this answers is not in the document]
mjvkeulen: > Clicking the same "+" button twice, creates the error "Level name: Less than Full marks is not unique."

Would otherwise create the error "Name cannot be empty" (just showing that the error argument is not relevant to this discussion)

I am fine with an empty name, I like the less than prefix as it motivates you to work from high to low (similar to how the points are reduced)

src/django/VLE/models/rubric.py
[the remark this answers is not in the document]
mjvkeulen: Good catch, fixed with dedicated test

src/django/VLE/serializers.py
```diff
@@ -1152,3 +1224,161 @@ def get_target(self, jir):
...
+            'levels',
+            'rubric',
+            # Annotated fields
+            'is_in_use',
+        )
+        read_only_fields = ()
+        write_once_fields = ('rubric',)
+
+    prefetch_related = [
+        Prefetch('levels', queryset=VLE.models.Level.objects.order_by('location')),
```
Larspolo, line 1271: [D] I think I would prefer the `order_by` to be based on `-points` so that the highest amount of points is always left. Can get really confusing if that would not be the case
image

src/vue/src/components/rubric/edit/LevelEdit.vue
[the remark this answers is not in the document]
Larspolo: Changed it to NIT, to prevent this further halting the PR.

But to also react to you:
> just showing that the error argument is not relevant to this discussion

True, although I find "name cannot be empty" to be a clearer message to the user on what needs to be done. Especially because (at least in Firefox) it highlights the field that needs to be filled.

> similar to how the points are reduced

src/vue/src/components/assessment/Assessment.vue
engelhamer: I like the current better, actually I think 'overall' is a bit confusing and excessive in some instances already.

src/vue/src/components/assessment/AssessmentGrades.vue
engelhamer: Too long

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,172 @@
...
+            }
+        },
+    },
+}
+</script>
+
+<style lang="sass">
+@import '~sass/partials/rubric.sass'
+
+$final-grade-font-size: 1.5em
```
engelhamer, line 151: Would prefer a heading here instead (for consistency).

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,172 @@
...
+$grade-padding-left: 50px
+
+#final-grade-input
+    max-width: 100px
```
engelhamer, line 155: Why not use the `inline` class?

engelhamer: Works with em and thus suits the input better probably

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -6,17 +6,64 @@
                 End of assignment
             </h2>
         </template>
+
+        <template v-if="journal">
+            <load-wrapper :loading="loading">
+                <assessment
+                    v-if="assessment && !loading"
```
engelhamer, line 13: Loading should be obsolete giving the load wrapper.

src/vue/src/components/profile/NotificationSettings.vue
```diff
@@ -195,6 +195,11 @@ export default {
                     key: 'new_grade_notifications',
                     description: 'Receive an email when you receive a grade.',
                 },
+                {
+                    name: 'Overall journal assessment updates',
+                    key: 'new_overall_journal_assessment_notifications',
+                    description: 'Receive an email when a journal receives an overall assessment.',
```
engelhamer, line 201: In the communication to students, I don't like the term 'overall assessment' as it sounds vague and is not used colloquially. A bit too formal. Do you think we can maybe merge the notification option with entry grades?

src/vue/src/views/Assignment.vue
```diff
@@ -51,6 +51,12 @@
                         <option value="markingNeeded">
                             Sort by marking needed
                         </option>
+                        <option
+                            v-if="assignment.stats.unpublished_overall_assessments"
+                            value="unpublishedOverallAssessments"
+                        >
+                            Sort by assessment
```
engelhamer, line 58: Confusing what the difference with entry is now:

how about
'Sort by marking needed' -> 'Sort by unmarked entries'
'Sort by assessment' -> 'Sort by assessment published'

setup.cfg
```diff
@@ -2,12 +2,12 @@
 max-line-length = 120
 exclude = */migrations/*,*/settings/*,*/tasks/__init__.py,*/test/factory/__init__.py
 max-complexity = 22
-max-function-length = 210
+max-function-length = 230
```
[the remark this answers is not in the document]
Larspolo: Yeah, not a favour of changing this even higher

src/django/VLE/models/rubric.py
```diff
@@ -0,0 +1,225 @@
...
+                criterion=OuterRef('pk'),
+            ).values(
+                'pk',
+            ),
+        )
+
+        return self.annotate(is_in_use=Exists(rubric_submission_qry))
+
+
+# QUESTION RUBRIC: Should we enforce max to min (left to right) on the levels of one criterion? So each sequential
```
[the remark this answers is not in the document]
Larspolo: Agree, also in another comment somewhere else

src/vue/src/components/assessment/AssessmentGrades.vue
Larspolo: The else case is way longer:
`return 'Override the default grade. Any future grade updates will no longer affect the final grade.'`

src/django/test/test_assessment.py
```diff
@@ -0,0 +1,547 @@
...
+        assessment = factory.Assessment(journal=journal, rubric_submission__rubric=self.full_rubric)
+        rubric_submission = assessment.rubric_submission
+        criterion_submission = rubric_submission.criterionsubmission_set.first()
+        assert criterion_submission, 'A Criterion submission should be generated by the factories'
+
+        assessment.is_draft = True
+        assessment.is_published = False
+        assessment.is_archived = False
+        assessment.save()
+        assert teacher.can_edit(assessment), 'Assigned teachers can always edit drafts'
```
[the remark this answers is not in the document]
Larspolo: Or test on the permission specific. And then check if both teachers and TAs have that permission by default

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: No, this is something I brought up a couple of times before. The load wrapper does not guard against its content being admitted into the DOM before loading is finished. So initialization issues arise when relying on the load-wrapper for this task.

src/django/VLE/models/notification.py
```diff
@@ -61,6 +62,14 @@ class Notification(CreateUpdateModel):
                 'button_text': 'View Grade',
             },
         },
+        NEW_OVERALL_JOURNAL_ASSESSMENT: {
+            'name': 'new_overall_journal_assessment_notifications',
+            'content': {
+                'title': 'New overall assessment',
```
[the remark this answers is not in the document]
mjvkeulen: Did you have an alternative in mind?

src/django/VLE/models/rubric.py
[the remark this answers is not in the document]
mjvkeulen: It allows a (stubborn) teacher to work from min to max, or possibly insert a negative number of points (e.g. penalty) in between relevant levels.

This is why I opted to nudge towards max to min, but not to enforce it.

src/django/VLE/models/rubric.py
```diff
@@ -0,0 +1,225 @@
...
+
+    name = models.TextField()
+
+    description = models.TextField(
+        blank=True,
+    )
+
+    points = models.FloatField()
+
+    initial_feedback = models.TextField(
```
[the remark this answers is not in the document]
mjvkeulen: Default feedback for the selected level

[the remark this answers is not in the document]
engelhamer: I found this feature a bit confusing because that's really what the rubric contents are for

src/vue/src/components/journal/JournalEndCard.vue
engelhamer: That's weird since the load wrapper literally contains this exact `v-if` too? Hiw is this possible?

src/vue/src/components/profile/NotificationSettings.vue
mjvkeulen: I thought as a student, would I respond differently if my portfolio has received a grade to an individual element, or if my portfolio has received an overall evaluation determining my final grade?

I figured I would care quite a bit more, sufficiently so that I might even configure my notification settings differently between the two types of notifications.

mjvkeulen: Regarding the exact phrasing, I am open to suggestions of course. 

I figured with '_overall assessment_', the meaning is quite clear at least. It is the overall assessment of your journal?

src/vue/src/views/Assignment.vue
mjvkeulen: I went with 'Sort by unpublished assessments', as I think that is what you wanted and in line with 'Sort by unmarked entries'?

src/django/VLE/models/rubric.py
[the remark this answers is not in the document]
mjvkeulen: E.g. a teacher can configure a feedback snippet that is generally applicable when scoring that specific level of a rubric.

src/vue/src/components/assessment/AssessmentGrades.vue
mjvkeulen: As in, use an HTML header element?

src/django/VLE/serializers.py
mjvkeulen: Let's focus the discussion in the QUESTION thread. 

Just to add on the topic of `order_by '-points'`, we would still need location fields as we allow equal points for two or more levels within a criterion.

src/django/VLE/models/notification.py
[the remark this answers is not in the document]
mjvkeulen: I figured this was in line with how we approach grades, when a grade is updated we also send a "New Grade" notification afaik.

src/django/VLE/models/rubric.py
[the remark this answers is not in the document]
mjvkeulen: I encountered this feature when checking out Brightspace's rubric system and thought it was a nice addition.

Larspolo (changes requested):

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -0,0 +1,333 @@
...
+                    <b class="float-right">
+                        Total:
+                    </b>
+                    <b-button
+                        class="green-button"
+                        variant="link"
+                        @click="addCriterion"
+                    >
+                        <icon name="plus"/>
+                        Add criterion
```
Larspolo, line 123: Addding a simple criteria to an existing rubric gives me the following 500 error:
image

src/vue/src/components/rubric/edit/LevelEdit.vue
```diff
@@ -0,0 +1,124 @@
...
+                v-model="level.name"
+                placeholder="Name"
+                type="text"
+                trim
+                required
+            />
+        </b-form-group>
+
+        <b-form-group>
+            <b-form-textarea
```
Larspolo, line 14: Would like to be able to resize this in the width as well tbh, that would allow me to read it better / edit the text better.

Or maybe we can wider the text area when clicking on it? So that it automatically is better for writing?

Or possibly also reducing the font size (which also better represents the readmode)?

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -0,0 +1,338 @@
...
+    .level-container
+        height: 100%
+        min-width: $rubric-min-cell-width
+        position: relative
+
+        .add-level
+            position: absolute
+            top: 10px
+            right: -13px
+            transform: translateX(50%)
```
mjvkeulen, line 329: @engelhamer the rounded button is an improvement, but I would prefer if this remains aligned in the middle of a level cell

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -0,0 +1,338 @@
...
+            padding: 0px
+            width: 20px
```
mjvkeulen, line 331: @engelhamer there is insufficient room for the button once in active state once clicked:

image

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,159 @@
...
+                class="text-right align-middle"
+            >
+                <icon
+                    v-if="assessment.points !== null"
+                    v-b-tooltip="'Grade is set manually by the assessor: Any future entry or bonus point updates will '
+                        + 'no longer affect the final grade'"
+                    class="fill-red mr-2 shift-up-4"
+                    name="info-circle"
+                />
+                <b>Final grade:</b>
```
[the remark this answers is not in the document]
mjvkeulen: Is it not still displayed for the final grade in read mode?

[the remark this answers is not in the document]
mjvkeulen: Yes, but not in edit mode. Added to edit mode as well.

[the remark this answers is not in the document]
mjvkeulen: I like the first suggestion

src/vue/src/components/rubric/edit/RubricEdit.vue
mjvkeulen: Shit, but we do needs steps on how to reproduce this

src/vue/src/components/journal/JournalDetails.vue
```diff
@@ -44,6 +44,17 @@
                     'mt-2': !assignment.is_group_assignment && $hasPermission('can_view_all_journals'),
                 }"
             />
+
```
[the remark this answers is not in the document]
mjvkeulen: Fixed

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,159 @@
...
+        </b-tr>
+
+        <b-tr class="background-light-grey">
+            <b-td
+                colspan="2"
+                class="text-right align-middle"
+            >
+                <icon
+                    v-if="assessment.points !== null"
+                    v-b-tooltip="'Grade is set manually by the assessor: Any future entry or bonus point updates will '
```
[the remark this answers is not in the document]
mjvkeulen: I am neutral towards this

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -0,0 +1,159 @@
...
+                />
+                <b>Final grade:</b>
+            </b-td>
+            <b-td class="text-nowrap">
+                <b-form-input
+                    v-if="edit"
+                    v-model="assessment.points"
+                    class="inline align-middle"
+                    type="number"
+                    :placeholder="defaultFinalGrade.toString()"
```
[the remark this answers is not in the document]
mjvkeulen: The old click to edit implementation was a bit more clear for this workflow, but might get repetitive when doing a full assignment.

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -0,0 +1,333 @@
...
+            </b-button>
+            <h2 class="theme-h2">
+                {{ (rubric.name) ? rubric.name : 'Rubric' }}
+            </h2>
+        </div>
+
+        <rubric
+            v-if="readMode"
+            :rubric="rubric"
+        />
```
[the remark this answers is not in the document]
mjvkeulen: I added it to the journal assessment, but not sure if I would like it here as well.

src/vue/src/components/rubric/CriterionRow.vue
```diff
@@ -0,0 +1,220 @@
...
+table.criterion-levels
+    padding: 0px
+    border-collapse: separate
+    tr:first-child td
+        min-width: $rubric-min-cell-width
+        border-radius: 10px
+        .level-selected-check
+            display: none
+        &.selected
+            background-color: #d1ecf1
```
mjvkeulen, line 212: Should we extract this or make this one of our theme colors?

src/vue/src/components/rubric/CriterionRow.vue
```diff
@@ -0,0 +1,220 @@
...
+
+table.criterion-levels
+    padding: 0px
+    border-collapse: separate
+    tr:first-child td
+        min-width: $rubric-min-cell-width
+        border-radius: 10px
+        .level-selected-check
+            display: none
+        &.selected
```
mjvkeulen, line 211: How do you guys feel about the additional borders introduced by the level table horizontally?

image
image

image

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -1,70 +1,83 @@
...
+            </b-td>
+        </b-tr>
 
-        <tr
-            class="final-grade-row"
-            :class="{ 'final-grade-seperator': displayFinalGradeSeperator }"
-        >
-            <td>
-                <b>Final Grade: </b>
+        <b-tr class="background-light-grey">
```
mjvkeulen, line 45: This looks a bit funky if there is no content in between

image

engelhamer:
Background-light-blue on the mark for assessment branch

Op zo 20 jun. 2021 17:26 schreef mjvkeulen ***@***.***>:

src/vue/src/components/assessment/AssessmentGrades.vue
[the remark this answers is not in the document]
engelhamer: Canvas does it this way too.

src/vue/src/components/assessment/AssessmentGrades.vue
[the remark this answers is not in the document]
engelhamer: Bootstrap table has support for sticky columns, but I think they only stick to the left hand side

src/vue/src/components/rubric/edit/RubricEdit.vue
engelhamer: It is confusing that when the cells are very large, the button also slides down to maintain in the middle.

src/vue/src/components/rubric/edit/RubricEdit.vue
engelhamer: Shall we try to ignore bootstrap accessibility stuff (the outline) here?

src/vue/src/components/rubric/CriterionRow.vue
engelhamer: I don't really mind, actually I think the sligt padding looks better than if the line would touch the blue shaded area

src/vue/src/components/assessment/AssessmentGrades.vue
engelhamer: We could remove the v-if and also show those rows for 0? Would make it nice and consistent across all uses 🙂

src/vue/src/components/rubric/CriterionRow.vue
engelhamer: Background-light-blue on mark for assessment branch

src/vue/src/components/assessment/AssessmentGrades.vue
Larspolo: I agree with that suggestion

src/vue/src/components/rubric/CriterionRow.vue
Larspolo: I dont mind it either

src/vue/src/components/assessment/AssessmentGrades.vue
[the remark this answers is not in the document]
Larspolo: Good enough for me, if also for @mjvkeulen he can resolve this threat

src/vue/src/components/rubric/edit/RubricEdit.vue
Larspolo: ```{"id":4,"assignment":1,"name":"Many levels","description":"Rubric description","criteria":[{"id":12,"name":"Overflow","description":"Would rock dream final production person. Seek forget Mrs deal. Think bad something treat end cultural.","location":0,"levels":[{"id":37,"name":"Level 31","description":"Indicate stock mind visit base. Technology involve partner. Ahead item what notice.","points":5,"initial_feedback":"Initial feedback snippet","location":0,"criterion":12,"is_in_use":false},{"id":38,"name":"Level 32","description":"Look traditional while level. Onto physical green blood him property. High though speak move.","points":5,"initial_feedback":"Initial feedback snippet","location":1,"criterion":12,"is_in_use":false},{"id":39,"name":"Level 33","description":"Behind money fish story be imagine attention respond. Reason quickly mind strategy very science.","points":5,"initial_feedback":"Initial feedback snippet","location":2,"criterion":12,"is_in_use":false},{"id":40,"name":"Level 34","description":"Art money door affect. Plan school nothing what sport. Catch themselves impact something hard adult blood.","points":5,"initial_feedback":"Initial feedback snippet","location":3,"criterion":12,"is_in_use":false},{"id":41,"name":"Level 35","description":"Fish all watch. Customer bank apply certain live system together. Little you foreign or.","points":5,"initial_feedback":"Initial feedback snippet","location":4,"criterion":12,"is_in_use":true},{"id":42,"name":"Level 36","description":"Summer not social voice energy left window. Enjoy near country sound.","points":5,"initial_feedback":"Initial feedback snippet","location":5,"criterion":12,"is_in_use":false},{"id":43,"name":"Level 37","description":"Group history follow then once reduce. Issue door lose.","points":5,"initial_feedback":"Initial feedback snippet","location":6,"criterion":12,"is_in_use":false}],"rubric":4,"is_in_use":true},{"id":-1,"name":"Criteria 2","description":"","long_description":"","rubric":4,"location":1,"levels":[{"name":"Full marks","points":10,"location":0,"id":-1},{"name":"No marks","points":0,"location":1,"id":-2}]}],"is_in_use":true}```
Maybe the raw data helps

src/vue/src/components/journal/JournalAssessment.vue
```diff
@@ -219,7 +219,7 @@ export default {
                 ? rubricSubmission.criterionsubmission_set.reduce(
                     (acc, criterionSubmission) => acc + parseFloat(criterionSubmission.points ?? 0), 0,
                 )
-                : 0
+                : null
```
engelhamer, line 222: Why do you want this to be `null` instead of 0? Is it displayed anywhere?

engelhamer:
Can give negative assessment:
image

engelhamer:
scroll_rubric
Shall we also limit the max width for the cells? This lower table is extremely wide (don't think content is easy to write in this case).

src/vue/src/views/Assignment.vue
engelhamer: Yes

src/vue/src/components/profile/NotificationSettings.vue
engelhamer: Throughout my whole student career I have never asked peers about their 'overall assessment', but always about their grade. Even in cases where this was actually an 'overall assessment'.

src/django/VLE/models/notification.py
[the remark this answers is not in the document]
engelhamer: It would be 'Assessment updated'. But OK for now.

src/django/VLE/models/assessment.py
```diff
@@ -0,0 +1,118 @@
...
+            **self.UNPUBLISHED,
+        )
+
+    def assessment_history_for_journal(self, journal, func='filter'):
+        """Returns the assessment history of the given journal from oldest to newest"""
+        return getattr(self, func)(
+            journal=journal,
+            is_draft=False,
+        ).order_by(
+            'update_date',
```
engelhamer, line 54: Reverse this, such that the active assessment ends up at the top of the list? This is more convenient than having to scroll down to see the current one.

src/vue/src/components/journal/JournalAssessment.vue
mjvkeulen: To me it makes sense that the value is null if no levels / points have been provided

src/vue/src/components/rubric/edit/RubricEdit.vue
mjvkeulen: The center makes sense, as we are adding another whole level (of the same size as the current row). 

The button on top is as if its related specifically to the name field and not as pleasant on the eyes (not as symmetrical)

src/django/VLE/models/assessment.py
mjvkeulen: I believe the entry grade history is also old to new? 

Makes sense when evaluating a chronological list?

src/vue/src/components/rubric/edit/LevelEdit.vue
mjvkeulen: Dynamically resizing table cell width sucks, I implemented option two: increase text area width if ever focused.

Larspolo:
The tutorial does not (yet) include the rubric part, think that might be really useful actually

src/django/VLE/models/assessment.py
engelhamer: But having to scroll is not nice.

When e.g. viewing news articles the top results are also most recent, though still in chronological order (just reversed).

engelhamer (approved):

src/vue/src/components/rubric/edit/LevelEdit.vue
Larspolo: Much better! Although it now doesnt shrink when going out of the box, is this intended behaviour or not? Both are fine by me, but wanted your opinions on it too

Larspolo: Also the criteria description could use this feature, that still stays the same for me

src/django/VLE/models/assessment.py
Larspolo: Would expect grade history to also be new to old btw

engelhamer:
Didn't notice, but would want it to shrink indeed. Definitely not desirable
that the table stays larger.

Op di 22 jun. 2021 13:59 schreef Lars ***@***.***>:

Larspolo (approved):

PR #1172, opened 2021-06-11
More widely capture information frontend when something goes wrong

mjvkeulen:
Capture all frontend toasted error calls in sentry as info

- Fixes Z#1439


src/vue/src/helpers/toasted.js
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Called as $toasted.global.errorNoLog */
+    Vue.toasted.register(
+        'errorNoLog',
+        (payload) => payload.message,
+        errorOptions,
+    )
+
+    /*
+     * Default $toasted.error will also send a capture to sentry of the level `info`
```
engelhamer, line 57: Doesn't this mean waaay too much is captured? Error toasts happen even on login timeouts right?

engelhamer (commented):
Doesn't

src/vue/src/helpers/toasted.js
mjvkeulen: Yes, so this will capture all errors including these. What we will need to do is mute errors of a certain type, but this will give us complete insight into what errors users are encountering.

engelhamer (approved):

Larspolo (approved):

PR #1175, opened 2021-06-19
Fixed an issue where pending comments would not be published when publishing a grade

Larspolo:
## Description
Fixed an issue where pending comments would not be published when publishing a grade

I suggest watching the files not in diff mode, because somehow that became a real mess
-[] `assert list(publish_grades.call_args.kwargs['grades']) == list(Grade.objects.none())`
## Summary of changes
- Fixes [Z#1988](https://zube.io/ejourn-al/ejournal/c/1988)
- Fixes [Z#1985](https://zube.io/ejourn-al/ejournal/c/1985)

Not [Z#1987](https://zube.io/ejourn-al/ejournal/c/1987)


src/django/test/test_import_utils.py
```diff
@@ -308,7 +309,7 @@ def test__copy_grade_based_on_jir_action(self):
             entry, entry.grade, teacher2, JournalImportRequest.APPROVED_WITH_GRADES_ZEROED)
         assert grade.grade == 0, 'Approved with grades zeroed creates a new grade object set to zero.'
 
-    def test__select_vle_coupling_based_on_jir_action(self):
+    def test_select_vle_coupling_based_on_jir_action(self):
```
mjvkeulen, line 312: Isnt the old name more appropriate seeing the function is defined as `_select_vle_coupling_based_on_jir_action`?

Larspolo: Oww, missed that

mjvkeulen (changes requested):
Great find, can't believe this was an issue!

On the two items addressed

#### Publishing pending comments
- Good test for creating a newly published grade
- Missing test for publishing all assignment grades / all journal grades. Both call `publish_grades`, so just testing that function in this regard would be sufficient. I believe publish all assignment grades already asserts `publish_grades` is called, not sure about publishing all journal grades.

#### Publishing grade should sent notification to the LMS (Z#1988)
- This seems addressed in the code but is missing an added test.

I took the liberty of rewriting `update_entries_to_most_recent_grade` into a single query and adding it to the entry queryset as `update_link_to_most_recent_grade` and added a unit test for it.

src/django/test/test_assignment.py
```diff
@@ -1359,25 +1360,32 @@ def test_publish_assignment_grades(self):
...
             'unpublished grades should now be published'
         assert not Entry.objects.get(pk=grade_other_journal.entry.pk).grade.published, \
-            'grades not in assignment should not be published'
+            'grades not in another assignment should not be published'
 
         # Now check if the query amount doesnt change if there are more grades to be published
         for _ in range(2):
-            factory.Grade(published=False, entry__node__journal=journal)
+            grade = factory.Grade(published=False, entry__node__journal=journal)
+            factory.Comment(entry=grade.entry, published=False)
```
mjvkeulen, line 1388: Q) Comments are created but no sequential check follows?

src/django/test/test_assignment.py
```diff
@@ -1359,25 +1360,32 @@ def test_publish_assignment_grades(self):
...
             mock.patch('VLE.tasks.notifications.generate_new_grade_notifications'),
-            mock.patch('VLE.utils.grading.task_bulk_send_journal_status_to_LMS')
+            mock.patch('VLE.utils.grading.task_bulk_send_journal_status_to_LMS.apply_async') as to_lms
         ):
             grading.publish_grades(
                 grades=Grade.objects.from_assignment(assignment).only_unpublished(),
                 author=author,
                 send_grade_notifications=False,
             )
+            assert to_lms.called, \
```
mjvkeulen, line 1370: Safer to use `assert_called_with` where possible

src/django/test/test_journal.py
```diff
@@ -533,12 +534,17 @@ def test_update_journal(self):
...
+    @mock.patch('VLE.utils.grading.publish_grades')
+    @mock.patch('VLE.models.user.User.check_permission')
+    def test_publish_all_journal_grades(self, check_permission, publish_grades):
+        journal = factory.Journal()
+        api.update(
+            self, 'journals',
+            params={'pk': journal.pk, 'published': True}, user=journal.assignment.author
+        )
+        check_permission.assert_called_with('can_publish_grades', journal.assignment)
+        assert publish_grades.called, \
```
mjvkeulen, line 546: Ditto

mjvkeulen (changes requested):

src/django/test/test_assignment.py
Larspolo: Whoops, fixed now

engelhamer (approved):

mjvkeulen (approved):

PR #1176, opened 2021-06-20
Fixed an issue where the is_teacher flag could get reset on some non-lti user updates

Larspolo:
The added test explains the problem


mjvkeulen (approved):
Clear problem, good catch, and clear test!

engelhamer (approved):

PR #1177, opened 2021-06-20
Fixed an issue where errors where not given to the end user

Larspolo:
## Description
When params are an empty string or None, it should actually raise an error instead of silently continue and hope for the best

Also fixes an issue where creating an assignment without points_possible supplied would cause a 500 error


mjvkeulen (approved):
I am in favor of the change, a required param of which the value is empty (be it `None` or `''`) should be explicitly handled. 

I can't foresee where this might break something, taking a wait-and-see approach to this is fine by me.

Pity we sort of reinvent the wheel of DRF (where you allow blank/null values explicitly via models).

src/django/VLE/utils/generic_utils.py
```diff
@@ -23,8 +23,8 @@ def required_params(post, *keys):
     result = []
     for key in keys:
         try:
-            if post[key] == '':
-                VLE.utils.error_handling.VLEMissingRequiredKey(key)
+            if post[key] == '' or post[key] is None:
```
mjvkeulen, line 26: We do close one avenue with this:

Let's say we want to enforce front end always provides a value for X in their request. X is optional, so None is a valid value.

With this change, this param would always be extracted via `optional_params`. So now we lose the guarantee that frontend provided the param explicitly as `None`?

PR #1179, opened 2021-06-22
Added a message for the user to indicate why the save button is disabled

Larspolo:
## Description
Creating a preset node entry takes 5 seconds of no user interaction before actually saving, this may confuse the user for a bit to why buttons are readonly, added a message for clarity
image

## Summary of changes
- Fixes [Z#2002](https://zube.io/ejourn-al/ejournal/c/2002)


src/vue/src/components/entry/Entry.vue
```diff
@@ -296,7 +296,11 @@ export default {
             window.clearTimeout(this.autoSaveTimeout)
 
             if (this.entryContentHasPendingChanges()) {
-                this.autoSaveMessage = ''
+                if (this.create) {
+                    this.autoSaveMessage = 'Waiting to save...'
```
mjvkeulen, line 300: [NIT] Saving... iso Waiting to save...

Larspolo: I first had saving, but this can stay for a long time, especially when a student has much to type, it will stay with "saving...", creating a feeling we might have a slow platform

engelhamer:
Unsaved changes

Op di 22 jun. 2021 13:04 schreef Lars ***@***.***>:

src/vue/src/components/entry/Entry.vue
engelhamer: Can also just put `Unsaved changes`

Larspolo: Would that be good enough for a user to know that they should wait a bit?

mjvkeulen: Unsaved changes works for me, I think others platforms use that as well

mjvkeulen: `Unsaved changes` (e.g. dirty) which is then updated to `Saving` if the request is sent and pending.

mjvkeulen (approved):

engelhamer (approved):

PR #1181, opened 2021-06-22
Added check migrations command

Larspolo:
## Description
Checks for leaf migrations and names containing `_auto_`

Then we never have the problem of pushing unmerges migrations


mjvkeulen (changes requested):
I like the command, seems like a good addition. Might make more sense (if-minor) to add this to the project checks instead (https://docs.djangoproject.com/en/3.2/topics/checks/).

Larspolo:
Cant be bothered to do the suggested change, closing PR instead

PR #1182, opened 2021-06-22
Fixed an issue in the migration chain which caused a pending error

Larspolo:
## Description
Split migrations that caused a pending error


mjvkeulen (approved):

engelhamer (commented):
Good

PR #1184, opened 2021-06-23
Expanded grading tests

Larspolo:
## Description
Because more tests is more better

Also added better response value to bulk sen to LMS

## Summary of changes
- Fixes Z#2015


src/django/test/test_grade.py
```diff
@@ -0,0 +1,71 @@
+import test.factory as factory
+from test.utils import api
+from unittest import mock
+
+from django.test import TestCase
+
+
+class EntryAPITest(TestCase):
```
mjvkeulen, line 8: Old name

mjvkeulen, line 8: [NIT]: Prob worth creating a 

```
@classmethod
    def setUpTestData(cls):
        cls.assignment = factory.Assignment()
```
And some other practical reusable model instances

Larspolo: I'm starting to not like those setup functions, because they only reduce some repeated lines, but not improve speed and can even reduce it. For example when I now add tests that would not need an assignment, it would wait resources on still creating one
We should look for fixtures instead.

mjvkeulen: I disagree, and I don't understand what you mean with for resources?

For example, you currently call `factory.UnlimitedEntry()` quite a couple of times across different tests here. Each of these is currently generating a journal and assignment, student, teacher, format, templates etc. The setUpTestData would really help.

mjvkeulen: > because they only reduce some repeated lines, but not improve speed and can even reduce it

You are aware `setUpTestData(cls)` is run once for the test module, unlike  `setUp(self)` which is run before each test in the module?

mjvkeulen: > We should look for fixtures instead.

I don't see where these would help more than our current approach

Larspolo: huh, no, why didnt we change that yet then?

mjvkeulen: All new test modules already do so, and some are already refactored as we touch old modules

engelhamer (approved):

mjvkeulen (approved):

PR #1193, opened 2021-06-30
Added preview to assessment history

mjvkeulen:
Peek 2021-06-30 10-00


Larspolo (approved):

src/vue/src/components/assessment/AssessmentHistory.vue
```diff
@@ -90,10 +89,25 @@
...
+                            class="ml-2 edit-icon"
+                            @click.native="restore(data.item)"
+                        />
+                    </div>
+                </template>
+
+                <template #row-details="data">
+                    <assessment
+                        :assessment="data.item"
                     />
```
Larspolo, line 111: image
Readmode of this component when no rubric is attached, is really unclear. Its not introduced in this PR, so will make an issue. Just pointing it out here as well

mjvkeulen: What particularly is unclear? This PR introduces (for the grade history only) a different background color to the previewed assessment

Larspolo: I know that this PR did not introduce it, I also stated that.

But the titles "criteria" and "levels" are not correct titles now. Also empty rows could possible be appended with a "0" in the score section iso an empty box

PR #1197, opened 2021-07-05
Added the ability to view entry field descriptions after posting

mjvkeulen:
Peek 2021-07-09 13-38

Ugly entry highlighting different field configurations:
image

- Fixes Z#2029


src/vue/src/components/entry/EntryFields.vue
```diff
@@ -33,12 +33,30 @@
...
-                :class="{ 'required': field.required }"
+                :class="{
+                    'required': field.required,
+                    'optional': !field.required,
+                }"
                 class="theme-h2 field-heading"
             >
                 {{ field.title }}
+
+                <icon
```
engelhamer, line 44: I don't really like the look of this icon when there is no title. Also in combination with the 'optional' tag there appears a lot of content whereas a teacher intentionally leaves a part untitled.

image

The popover is very small (limited in width) and does not behave well with scrolling and a lot of content (which is sometimes the case in practice).

engelhamer (commented):
I understand the need to view the description while grading, but do not think that a popover is the most accessible and appealing solution. How about:

src/vue/src/components/entry/EntryFields.vue
mjvkeulen: > I don't really like the look of this icon when there is no title. Also in combination with the 'optional' tag there appears a lot of content whereas a teacher intentionally leaves a part untitled.

I agree it does not look pretty when there is a description but no title, however using a different icon will not address this.

I can move the display to an inline toggle, I could also fix the limited width and scrolling issue. 

I originally went for the popover solution as allows for a brief hover without disturbing where a teacher is in the document.  Personally, I feel this fits the workflow of a teacher most. Expanding the description shifts the entry and mixes the description with the answer.

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -33,12 +33,30 @@
             class="mb-2"
         >
             <h2
-                v-if="field.title"
-                :class="{ 'required': field.required }"
+                :class="{
+                    'required': field.required,
```
Larspolo, line 37: If we do decide to also use the required symbol, then this should also be used in the edit mode:
image

src/vue/src/sass/partials/popover.sass
```diff
@@ -0,0 +1,7 @@
+.popover
+    border-radius: 10px
+    border-color: $border-color
+    .popover-header
+        border-radius: 10px 10px 0px 0px
+        background-color: $theme-light-grey
+        border-bottom: 1px solid $border-color
```
Larspolo, line 7: Headers are also bold

src/vue/src/sass/partials/popover.sass
```diff
@@ -0,0 +1,7 @@
+.popover
+    border-radius: 10px
```
Larspolo, line 2: [NIT] Should we make this a variable ;)

src/vue/src/components/entry/EntryFields.vue
```diff
@@ -33,12 +33,30 @@
...
+                    'optional': !field.required,
+                }"
                 class="theme-h2 field-heading"
             >
                 {{ field.title }}
+
+                <icon
+                    v-if="field.description"
+                    :id="`field-${field.id}-description-info`"
```
Larspolo, line 46: This will cause duplicate key errors when switching nodes with same template. Thats why other keys also use the node id, e.g.: `node-${nodeID}-field-${field.id}`. I suggest also using that here (and line 54 then as well)

Larspolo (changes requested):
Looking good!

src/vue/src/components/entry/EntryFields.vue
Larspolo: The option "Truncating the text to fit a line, then showing 'more...' to toggle the full description" also sounds nice though. I see the shifting stuff, but with sizable descriptions, this option will also cause problems, as the description might go out of view

mjvkeulen: Do we have something in place to truncate a sandboxed iframe's content?

src/vue/src/components/entry/EntryFields.vue
Larspolo: @mjvkeulen this is still a thing

engelhamer (approved):

src/vue/src/components/entry/EntryFields.vue
mjvkeulen: @engelhamer opinion?

engelhamer: I would prefer it to be EITHER star OR <optional> in one form. Not both.

Larspolo (approved):

PR #1200, opened 2021-07-05
Publishing assessments temporarily mutes LMS TODO's

Larspolo:
## Summary of changes
- Fixes Z#2015


src/django/test/test_grading_background_task.py
```diff
@@ -0,0 +1,125 @@
...
+SUCCESSFUL_RESPONSE = {
+    'code_mayor': 'success'
+}
+FAILED_RESPONSE = {
+    'code_mayor': 'failed'
+}
+
+
+class GradingBackgroundTaskTest(TestCase):
+
```
mjvkeulen, line 18: [NIT] Eventhough common, we do not follow a class def by hspace

src/django/test/test_grading_background_task.py
```diff
@@ -0,0 +1,125 @@
...
+    def setUp(self) -> None:
+        self.journal = Journal.objects.get(pk=factory.Journal(entries__n=0).pk)
+        self.author = self.journal.authors.first()
+        self.author.grade_url = 'https://ejournal.app'
+        self.author.sourcedid = 'TEST'
```
mjvkeulen, line 23: Would using factory.LtiJournal work?

src/django/test/test_grading_background_task.py
```diff
@@ -0,0 +1,125 @@
...
+            'When author has no grade url, it should say so'
+
+        self.author.sourcedid = None
+        self.author.save()
+        resp = grading.send_author_status_to_LMS(self.journal, self.author)
+        assert 'has no sourcedid' in resp['description'], \
+            'When author has no sourcedid url, it should say so'
+
+    @mock.patch('VLE.lti_grade_passback.GradePassBackRequest.send_post_request', return_value=SUCCESSFUL_RESPONSE)
+    def test_send_grade(self, send_post_request: mock.Mock):
```
mjvkeulen, line 49: [NIT]: The module context makes it clear that `test_send_grade` applies to the context of grading background task.

However, this is not an ideal term to search for / running keep testing for. I prefer not having to look at the module for context, or being able to plug part of the func name in to run _only_ that test (think `test_create` vs `test_entry_create`)

src/django/test/test_grading_background_task.py
```diff
@@ -0,0 +1,125 @@
...
+        send_post_request.assert_called()
+        assert resp['to_teacher'] is not None, \
+            'With an ungraded entry, teachers should receive a TODO'
+
+        send_post_request.reset_mock()
+
+        resp = grading.send_author_status_to_LMS(self.journal, self.author, grade_only=True)
+
+        assert resp['to_teacher'] is None, \
+            'If grade only is true, no TODOs should be temporarily muted for the teacher'
```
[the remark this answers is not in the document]
Larspolo: Added this:
```
# If you want to temporarily mute TODOs for teachers, grade_only can be supplied
# This may be used when publishing an assessment for example
```

src/django/test/test_grading_background_task.py
Larspolo: Didnt know it existed, changed

src/django/test/test_grading_background_task.py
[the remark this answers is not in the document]
engelhamer: Also this assert should read `'If grade only is true, TODOs should be temporarily muted for the teacher'` right?

engelhamer (approved):

mjvkeulen (approved):

PR #1205, opened 2021-07-07
Added a warning message when preset node names get too long

Larspolo:
## Description
One letter above the max of 30:
image
All M's can still cause it to be too long (only displays 23 M's), but I figured that will not happen
image

## Summary of changes
- Fixes Z#1748


src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodeEdit.vue
```diff
@@ -42,6 +42,11 @@
                     v-if="presetNode.template && presetNode.template.allow_custom_title"
                 >
                     The template setting <i>"allow custom title"</i> allows students to override this value.
+                    <br/>
+                </small>
+                <small v-if="presetNode.display_name.length > 30">
```
mjvkeulen, line 47: I believe we need this check for a custom entry title as well

mjvkeulen (changes requested):
I would expect this check to be similar for a custom entry-title

engelhamer (approved):

PR #1209, opened 2021-08-31
Improved the simultaneous editing experience

mjvkeulen:
Introduces an active user display for actions such as editing a group assignment's entry or a journal's overall assessment. If one of the active users is editing the resource, that resource is then "_claimed_" and cannot be edited until the user is finished and the resource is released. Once finished the other active users are notified and the resource is updated.

Peek 2021-08-31 10-13

Adds the Channels library to the backend to support incoming WebSocket connections. To keep the resource cost minimal each resource shares the same WebSocket connection (so 1 WS connection per user, but still allows a user to have a presence for multiple resources).

The backend logic is grouped under a new app (think module) Channels User Presence (**cup**). Which can be run under a webserver specializing in async connection. 
- The same user can have a **Presence** at multiple resources represented as **Room**s. 
-  A presence is identified by a channel name (unique, provided by the channel package)
- Rooms group channels, and are also identified by a unique channel name. These can be used to broadcast messages to groups of users.

This PR also adds Redis to function as a Channel layer, allowing us to talk between different instances of an application without moving through a DB.


src/vue/src/store/modules/usage.js
```diff
@@ -0,0 +1,216 @@
...
+
+const actions = {
+    connect (context) {
+        return new Promise((resolve, reject) => { // eslint-disable-line consistent-return
+            if (context.state.socket.isConnected) {
+                return resolve(context.state.socket.ws)
+            }
+
+            const accessToken = context.rootGetters['user/jwtAccess']
+            const ws = new WebSocket(`${wsUsageUrl}?token=${accessToken}`)
```
Larspolo, line 120: [Q]
image
I notice three websocket creation network requests, are these all needed? I would expect only 1. The others also do not communicate anything.

I also noticed three heartbeats:
image

src/django/cup/views/usage.py
```diff
@@ -0,0 +1,25 @@
+import VLE.utils.generic_utils as utils
+import VLE.utils.responses as response
+from cup.models import Room
+from cup.utils.usage import get_room_channel_name
+from rest_framework import viewsets
+
+
+class UsageView(viewsets.ViewSet):
+    def list(self, request):
```
Larspolo, line 9: Can any user now request all information about any room mixed with type? Would expect here as well some check like `user.can_view(resource)` or sth

src/django/VLE/serializers.py
```diff
@@ -234,6 +235,7 @@ class Meta:
     allow_custom_categories = serializers.BooleanField(source='chain.allow_custom_categories')
     allow_custom_title = serializers.BooleanField(source='chain.allow_custom_title')
     title_description = serializers.CharField(source='chain.title_description')
+    chain_id = serializers.IntegerField(source='chain.pk')
```
Larspolo, line 238: Why this change?

src/django/VLE/views/entry.py
```diff
@@ -18,12 +18,12 @@
 
 
 class EntryView(viewsets.ViewSet):
-    """Entry view.
+    def retrieve(self, request, pk):
+        entry = serialize.EntrySerializer.setup_eager_loading(Entry.objects.filter(pk=pk)).get()
+        request.user.check_can_view(entry.node.journal)
```
Larspolo, line 23: This now bypasses the drafted entries I think. Should also check for that at least (a nice permission system would be good for this too)

src/django/cup/consumers/usage_consumer.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def disconnect(self, close_code):
+        Presence.objects.leave_all(self.channel_name)
+
+    @touch_presence
+    def receive(self, text_data):
+        text_data_json = json.loads(text_data)
+        action_str = text_data_json['action']
+
+        # TODO WS: Can view / edit resource
```
Larspolo, line 42: ^

src/django/cup/consumers/usage_consumer.py
```diff
@@ -0,0 +1,79 @@
...
+            room_channel_name_str = get_room_channel_name(text_data_json)
+
+            Room.objects.remove_if_exists(
+                room_channel_name=room_channel_name_str,
+                user_channel_name=self.channel_name,
+            )
+        elif action_str == self.RELEASE_ALL:
+            Presence.objects.leave_all(self.channel_name)
+        else:
+            raise CUPInvalidAction(f'Unknown action provided: "{action_str}".')
```
Larspolo, line 72: [NIT] Possibly all cases in separate function, especially when further expanding this, it would be nice

src/django/cup/jwt_auth_middleware.py
```diff
@@ -0,0 +1,71 @@
...
+            if (jwt_token_list := parse_qs(scope['query_string'].decode('utf8')).get('token', None)):
+                jwt_token = jwt_token_list[0]
+                jwt_payload = self.get_payload(jwt_token)
+                user_credentials = self.get_user_credentials(jwt_payload)
+                user = await self.get_logged_in_user(user_credentials)
+                scope['user'] = user
+            else:
+                scope['user'] = AnonymousUser()
+        except (InvalidSignatureError, KeyError, ExpiredSignatureError, DecodeError) as e:
+            capture_exception(e)
```
Larspolo, line 42: Should this also get an anonymous user?

src/django/cup/tasks.py
```diff
@@ -0,0 +1,12 @@
+from celery import shared_task
+from cup.models import Room
+
+
+@shared_task(name='cup.tasks.prune_presences')
+def prune_presences():
+    Room.objects.prune_presences()
+
+
+@shared_task(name='cup.tasks.prune_rooms')
```
Larspolo, line 10: [Q] Is there a way of automatically having them enabled? How long would this take? Or would each developer always need to set this up manually?

src/django/cup/utils/constants.py
```diff
@@ -0,0 +1,10 @@
+resource_type_set = {
```
Larspolo, line 1: contants are often IN_ALL_CAPS

Larspolo (changes requested):
Backend reviewed

engelhamer:
Not ready for merging.

PR #1211, opened 2021-09-07
Fixed an issue where the last edited and creation date of entries could be inaccurate

Larspolo:
EJ-789 #close


src/django/VLE/models/entry.py
```diff
@@ -120,7 +121,12 @@ class Entry(CreateUpdateModel):
         null=True,
     )
 
-    last_edited = models.DateTimeField(auto_now_add=True)
+    last_edited = models.DateTimeField(
+        null=True,
+    )
+    first_posted = models.DateTimeField(
```
mjvkeulen, line 127: We use submit for our buttons that perform this action, how about `first_submitted`?

src/django/VLE/models/entry.py
```diff
@@ -120,7 +121,12 @@ class Entry(CreateUpdateModel):
         null=True,
     )
 
-    last_edited = models.DateTimeField(auto_now_add=True)
+    last_edited = models.DateTimeField(
```
mjvkeulen, line 124: [NIT]: I wouldn't mind a docstring or comment explaining the differences between the date fields.

src/django/VLE/models/entry.py
```diff
@@ -273,18 +285,22 @@ def link_categories(self, category_ids):
         ]
         VLE.models.EntryCategoryLink.objects.bulk_create(entry_category_links)
 
-    def should_send_new_entry_notification(self, is_new, was_draft):
+    def should_update_last_edited(self, was_draft):
+        # Entries should update last edited when it is switched from draft to no draft
```
mjvkeulen, line 289: I would still like to be able to determine when a student made the last edit to entries for drafts. So we would still be able to determine the activity for drafted entries.

I don't believe `update_date` suffices, when content is updated the entry is not saved again by definition.

I do agree that last_edit in the eyes of the teacher makes more sense as it's rewritten in this PR.

So does this mean we need another field that is always updated in the manner `last_edited` was before this PR? (After each entry create or update request).

mjvkeulen (changes requested):

engelhamer (approved):

src/vue/src/components/entry/EntryNonStudent.vue
```diff
@@ -76,22 +76,27 @@
...
-                    />
-                    <br/>
+                    <template v-if="node.entry.first_posted">
+                        Posted
+                        <timeago
+                            v-b-tooltip.hover="$root.beautifyDate(entryNode.entry.first_posted)"
+                            :datetime="entryNode.entry.first_posted"
+                        />
+                        by {{ entryNode.entry.author }}
+                        <!-- Last edited is only displayed if its edited later then three minutes after posting -->
```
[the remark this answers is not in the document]
Larspolo: Created issue

src/vue/src/components/entry/EntryNonStudent.vue
```diff
@@ -174,6 +179,7 @@
                         class="fill-red"
                     />
                 </template>
+                <!-- TODO: find out what this is again -->
```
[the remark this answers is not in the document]
Larspolo: Jup, grade creation date, not entry creation, so its all good

src/django/VLE/models/entry.py
mjvkeulen: Tracking when any entry was last edited (as before this PR), doesn't seem like some obscure use case?

engelhamer:
But this shouldn't matter for drafts? It's all the studentsm's business and
only becoems relevant from when the entry is submitted/posted.

Op wo 15 sep. 2021 15:56 schreef mjvkeulen ***@***.***>:

src/django/VLE/models/entry.py
Larspolo: The case for this specific feature was teacher focussed

mjvkeulen (changes requested):
Entry model now has the following date fields::
- creation_date (draft start)
- update_date (awkward right now, should be updated whenever the entry title or content is - - updated (what last_edited is atm)
- first_submitted (I like this addition)
- last_edited (staat nu weer goed)

I would like to maintain the functionality of what last_edited does currently, this would allow us to gain insights into when the user was working on an entry last.

To achieve this I would leave `update_date` alone, as throughout our models this consistently displays when a model's concrete fields were updated last.

So I would suggest the addition of a new field? To accomplish this requires minimal work (3 occurrences where entry.last_edited is updated), and has to be done in this PR as the migration is tied to this change.

I will admit that the name might be slightly confusing, because update_date and last_edited both already exist.

src/django/VLE/migrations/0099_add_first_and_last_submitted.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 3.2.3 on 2021-09-22 10:13
+
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+# Generated by Django 3.2.3 on 2021-09-07 10:49
+
+
+
```
mjvkeulen, line 10: Prob just fix for the linter (do we exclude these?)

src/django/VLE/migrations/0099_add_first_and_last_submitted.py
```diff
@@ -0,0 +1,53 @@
...
+def provide_one_off_default(apps, schema_editor):
+    Entry = apps.get_model('VLE', 'Entry')
+
+    Entry.objects.filter(is_draft=False).update(
```
mjvkeulen, line 14: Q) What do we do with redrafts?

src/django/VLE/models/entry.py
```diff
@@ -120,11 +116,20 @@ class Entry(CreateUpdateModel):
         null=True,
     )
 
+    # Last time the content was edited, does include draft edits, but not grade edits, used for analysis
```
mjvkeulen, line 119: Clear comments

src/django/VLE/models/entry.py
```diff
@@ -273,18 +284,22 @@ def link_categories(self, category_ids):
         ]
         VLE.models.EntryCategoryLink.objects.bulk_create(entry_category_links)
 
-    def should_send_new_entry_notification(self, is_new, was_draft):
+    def should_update_last_submitted(self, was_draft):
+        # Entries should update last edited when it is switched from draft to no draft
```
mjvkeulen, line 288: [NIT] Should this comment not be captured in the function name?

E.g. self.submitting_draft(self, was_draft) and then

if self.submitting_draft(was_draft):
    entry.last_submitted = now

src/django/VLE/models/entry.py
```diff
@@ -273,18 +284,22 @@ def link_categories(self, category_ids):
...
+        return was_draft and not self.is_draft
+
+    def should_set_first_submitted(self, was_draft):
```
mjvkeulen, line 291: [NIT] This would then be `first_submission` inline with the other comment

src/django/test/test_entry.py
```diff
@@ -1362,8 +1362,11 @@ def test_draft(self):
             resp = api.create(self, 'entries', params=create_draft_params, user=self.student)['entry']
             assert not task_journal_status_to_LMS.called, \
                 'Grade updates should not be send to Canvas when a drafted entry is created'
+        entry = Entry.objects.get(pk=resp['id'])
         assert resp['is_draft'], 'Entry response should be draft if is_draft is set to true'
-        assert Entry.objects.get(pk=resp['id']).is_draft, 'Entry should be draft if is_draft is set to true'
+        assert entry.is_draft, 'Entry should be draft if is_draft is set to true'
+        assert not entry.first_submitted, 'Entry should not get a first_submitted when still in draft'
+        assert not entry.last_submitted, 'Entry should not get a last_submitted when still in draft'
```
mjvkeulen, line 1369: Did I miss it somewhere else or do we need to add a line to check whether last_edited is updated when an entry is updated (even when draft), but not when graded?

src/vue/src/components/entry/Entry.vue
```diff
@@ -78,25 +78,27 @@
...
-                        name="history"
-                    />
-                    <br/>
+                    <template v-if="node.entry.first_submitted">
+                        Posted
+                        <timeago
+                            v-b-tooltip.hover="$root.beautifyDate(node.entry.first_submitted)"
+                            :datetime="node.entry.first_submitted"
+                        />
+                        by {{ node.entry.author }}
```
mjvkeulen, line 87: Q) Not introduced in this PR, but is this handled in the serializer or is this wrong?

mjvkeulen (changes requested):

src/django/VLE/migrations/0099_add_first_and_last_submitted.py
Larspolo: We do exclude these, because auto generated stuff caused errors

https://ejournal.atlassian.net/browse/EJ-1105

src/django/VLE/migrations/0099_add_first_and_last_submitted.py
Larspolo: We dont have that distinction currently, hard to do anything with it. But from now on, we keep the first and last submittted for this reason

src/django/VLE/models/entry.py
Larspolo: I find this more clear. Could add a new function that is called inside `should_update_last_submitted`

src/vue/src/components/entry/Entry.vue
Larspolo: `author = serializers.CharField(source='author.full_name', default=VLE.models.User.UNKNOWN_STR)`

What would be wrong?

src/django/test/test_entry.py
Larspolo:
```

    def test_update_entry(self):
        entry = api.create(self, 'entries', params=self.valid_create_params, user=self.student)['entry']

        params = {
            'pk': entry['id'],
            'content': entry['content'].copy()
        }

        updated_entry = api.update(self, 'entries', params=params.copy(), user=self.student)['entry']
        assert entry['last_edited'] != updated_entry['last_edited'], \
            'Last edited should update when entry content changes'
```

Larspolo: Also added one more check for drafted entries

mjvkeulen (approved):

PR #1212, opened 2021-09-07
Added start and end dates in manage LTI view

Larspolo:
## Description
image

# Fixes
- EJ-1019 #close


engelhamer (commented):
Only thing: 2 courses with same name, one with end year 2020 and other with start year 2020 still look the same. But this is not problematic in practice.

engelhamer (approved):

src/django/VLE/models/course.py
```diff
@@ -61,6 +61,16 @@ def add_assignment_lti_id(self, lti_id):
...
 
+    def get_course_dates_string(self):
+        dates_string = ''
+        if self.startdate:
+            dates_string += str(self.startdate.year)
+        if self.startdate and self.enddate:
+            dates_string += ' - '
+        if self.enddate:
+            dates_string += str(self.enddate.year)
+        return dates_string
```
mjvkeulen, line 72: Would utils/generic_utils.js courseWithDatesDisplay work? This would be more inline with how other locations display such dates.

mjvkeulen (changes requested):

src/django/VLE/models/course.py
Larspolo: Not with the same amount of effort. Took the function from there though. This would require the API to change as well, because now only the course names are serialized

mjvkeulen: How about for now also use the same format as that function use (`(<date_start> - <date_end>)`) and open an issue for doing this consistently. 

One way would be to add another serialized field to `course` and `assignment` which replaces `courseWithDatesDisplay` and `assignmentWithDatesDisplay`. Something like `list_name` or `selection_name`.

mjvkeulen (approved):

PR #1218, opened 2021-09-12
Added query and execution time profilers

Larspolo:
Silk doesn't work anymore, looked around quite a bit, and found these were the most convenient.

See https://docs.google.com/document/d/1suTO-3wr37KFfOPH7EpGGOoKWg_Holjnm3GZm3pjfQ8/edit# for more info on how this can be used


src/django/VLE/utils/profiling.py
```diff
@@ -0,0 +1,44 @@
+
+import marshal
+
+try:
+    import cProfile as profile
+except ImportError:
+    import profile
```
mjvkeulen, line 7: What forced this?

Larspolo: Think its not needed for our Python version, but this was also done in `django_cprofile_middleware`

requirements/base.txt
```diff
@@ -20,7 +20,6 @@ text-unidecode==1.3
 python-dateutil==2.8.1 # TODO Unsure of its dependants
 
 certifi==2020.12.5 # TODO figure out specific use cases
-httplib2==0.19.1 # TODO figure out specific use cases
```
engelhamer, line 23: Why can this be removed now? What was it for?

src/django/VLE/settings/local.py
```diff
@@ -28,9 +33,20 @@
 CORS_ORIGIN_ALLOW_ALL = True
 CSP_DEFAULT_SRC = ("'self'", "'unsafe-inline'")
 
-if os.environ.get('SILK', False):
-    INSTALLED_APPS += ['silk']
-    MIDDLEWARE = ['silk.middleware.SilkyMiddleware'] + MIDDLEWARE
+if os.environ.get('PROFILING', False):
+    # Used for the django_query_profiler
```
engelhamer, line 37: Why these?

src/django/VLE/utils/profiling.py
```diff
@@ -0,0 +1,44 @@
+
```
engelhamer, line 1: Disable codecov for this file?

requirements/base.txt
Larspolo: I think this version was never meant to be forced. It is a dependency on two packages, with the profiling requiring a higher version than 0.19.1, so thats why I removed it, it will still be downloaded

src/django/VLE/settings/local.py
Larspolo: Those are used for the CSS and JS to display the fancy profiling bars

engelhamer (approved):

src/django/VLE/settings/local.py
```diff
@@ -28,9 +33,20 @@
...
+    CSP_DEFAULT_SRC += (
+        'maxcdn.bootstrapcdn.com', 'cdnjs.cloudflare.com', 'd3js.org', 'cdn.jsdelivr.net', 'ajax.googleapis.com',
+    )
+    INSTALLED_APPS += ['django_query_profiler']  # noqa: F405
+    MIDDLEWARE = [
+        'django_query_profiler.client.middleware.QueryProfilerMiddleware',
+        'VLE.utils.profiling.ProfilingMiddleware',
+    ] + MIDDLEWARE + [  # noqa: F405
+    ]
```
mjvkeulen, line 46: What does this do? Bracket open/close on separate line and all

mjvkeulen (approved):
Just one question, seems to work fine!

src/django/VLE/settings/local.py
Larspolo: This is left over from me moving the profiliers around. If its after the MIDDLEWARE, then you exclude the performance of the middleware, nice to have

PR #1219, opened 2021-09-13
Added the ability to set a usage limit for a specific template

Larspolo:
image

image


src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,54 @@
...
+                    placeholder="-"
+                    min="0"
+                    step="1"
+                    :formatter="maxUsesFormatter"
+                />
+                <h2 class="theme-h2 field-heading mb-2">
+                    Max uses
+                </h2>
+                <small>
+                    Specify the amount that students can freely use this template. 0 for unlimited usage.
```
mjvkeulen, line 47: amount -> number (we can count the number of times the template will be used).

The number of times that students can make use of this template. Note that this also includes any deadlines that make use of this template. Use 0 to indicate unlimited usage.

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,54 @@
...
+                    id="max-uses-input"
+                    v-model="template.max_uses"
+                    type="number"
+                    class="float-right mb-2 ml-2"
+                    size="2"
+                    placeholder="-"
+                    min="0"
+                    step="1"
+                    :formatter="maxUsesFormatter"
+                />
```
mjvkeulen, line 42: Q) Do we need another indication to represent unlimited use than 0? Might not be very intuitive for our non-technical users. Small text below the input with unlimited when 0 is set would be fine for example.

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,54 @@
...
+                    },
+                ]"
+                class="float-right mb-3 ml-3"
+            />
+            <h2 class="theme-h2 field-heading mb-2">
+                Free use
+            </h2>
+            <small>
+                Allow students to use freely use this template for new journal entries as often as specified.
+                This can even be unlimited.
```
mjvkeulen, line 27: How about:
```
Allow students to freely create entries based on this template, or set an upper usage limit. 
When disabled, this template can only be used for deadlines you add to the timeline.
```

src/django/test/test_entry.py
```diff
@@ -1542,3 +1542,76 @@ def test_should_send_new_entry_notification(self):
...
+        student1 = journal1.authors.first().user
+        student2 = journal2.authors.first().user
+        template1 = journal1.assignment.format.template_set.first()
+        template2 = journal1.assignment.format.template_set.last()
+        valid_create_params = {
+            'journal_id': journal1.pk,
+            'template_id': template1.pk,
+            'content': {}
+        }
+        Field.objects.all().update(required=False)
```
mjvkeulen, line 1558: [NIT]: This would block switching the setup to the test module (as it would affect other state as well). Limiting the query to Field.objects.filter(template__format=journal1.assignment.format)

mjvkeulen (changes requested):

src/django/test/test_entry.py
```diff
@@ -1542,3 +1542,76 @@ def test_should_send_new_entry_notification(self):
...
+        Entry.objects.all().delete()
+        assert template1 != template2, \
+            'This test requires at least two templates to be generated by the default assignment factory'
+
+        factory.UnlimitedEntry(template=template1, node__journal=journal1)
+        template1.chain.max_uses = 2
```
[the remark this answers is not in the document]
Larspolo: `factory.errors.InvalidDeclarationError: Received deep context for unknown fields: {'chain__max_uses': 2} (known=['description', 'location', 'options', 'required', 'template', 'title', 'type'])`

src/django/test/test_entry.py
```diff
@@ -1542,3 +1542,76 @@ def test_should_send_new_entry_notification(self):
...
+            'With template 2 being able to be used unlimited, it should be returned in get_add_node for the timeline'
+
+        # With 2 max uses, and 1 already created, it should only be able to create 1 more entry
+        api.create(self, 'entries', params=valid_create_params, user=student1)
+        resp = api.create(self, 'entries', params=valid_create_params, user=student1, status=400)
+        assert 'You cannot use this template anymore.' == resp['description']
+
+        template_pks = [t['id'] for t in timeline.get_add_node(journal1)['templates']]
+        assert template1.pk not in template_pks, \
+            'Since template 1 is fully used, it should no longer be returned in get_add_node for the timeline'
```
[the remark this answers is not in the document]
Larspolo: image
Good one, think this is way better

src/vue/src/components/template/TemplateEditSettings.vue
Larspolo: I think the text on the left hand side is good enough as non-technical/new users will likely also read more of the text to understand it better

src/django/test/test_entry.py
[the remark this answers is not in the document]
mjvkeulen: Ah my bad, we flattened the chain creation structure by overloading `create`. 

factory.Journal(assignment__format__templates=[{'type': Field.TEXT, 'max_uses': 2}] would work!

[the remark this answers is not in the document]
mjvkeulen: Doesn't work! factory.Template(max_uses=2) does, but format `templates` is not setup correctly.

src/django/VLE/migrations/0097_templatechain_max_uses.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('VLE', '0096_mark_for_assessment'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='templatechain',
+            name='max_uses',
+            field=models.PositiveIntegerField(default=0),
```
mjvkeulen, line 16: Make use of UNLIMITED iso 0

src/django/VLE/models/journal.py
```diff
@@ -442,6 +442,35 @@ def get_sorted_nodes(self):
...
+            )
+            .annotate(
+                template_count=Count('pk')
+            )
+            .filter(
+                max_uses__gt=0,
+                template_count__gte=F('max_uses')
+            )
+            .values('pk')
+        )
```
mjvkeulen, line 458: And here I thought u loved floating commas

src/django/VLE/models/journal.py
```diff
@@ -442,6 +442,35 @@ def get_sorted_nodes(self):
...
+                archived=False,
+                preset_only=False
+            )
+        )
+        if not include_used_up_templates:
+            qry = qry.exclude(chain__in=self.get_used_up_template_chain_pks())
+
+        return qry.order_by(
+            'name'
+        )
```
mjvkeulen, line 472: [D]: get_available_templates(include_used_up....) is inherently confusing (to me). 

How about `get_templates` and `get_unavailable_templates` which calls `get_templates` and applies the exclude?

src/django/VLE/views/entry.py
```diff
@@ -79,7 +79,13 @@ def create(self, request):
             entry_utils.validate_node(node, template)
         # Unlimited entry
         else:
+            if not journal.get_available_templates().filter(pk=template.pk).exists():
```
mjvkeulen, line 82: For example, if journal.get_unavaible_templates().filter(pk=..).exists() would now skip the negation

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -82,6 +93,14 @@ export default {
                 this.$set(template, '_showDetails', template.id === row.item.id && !row.detailsShowing)
             })
         },
+        usedCount (row) {
+            return this.nodes.map(
+                (n) => (n.entry && n.entry.template && n.entry.template.id === row.item.id ? 1 : 0),
+            ).reduce((a, b) => a + b)
+        },
+        isUsedUp (row) {
+            return row.item.max_uses > 0 && this.usedCount(row) >= row.item.max_uses
```
mjvkeulen, line 102: [NIT]
define unlimited on the js side as well (e.g. in data, or in utils/constants)
max_uses !== unlimited && ..

Just so later on this is obvious from the js side as well

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -10,14 +10,22 @@
...
                 <b-button
                     variant="link"
-                    :class="{ 'input-disabled': addRequestInFlight }"
+                    :class="{ 'input-disabled': addRequestInFlight || isUsedUp(row) }"
                     @click="$emit('select-template', row.item.id); addRequestInFlight = true"
                 >
                     {{ row.item.name }}
                 </b-button>
+                <small
+                    v-if="row.item.max_uses > 0"
```
mjvkeulen, line 24: [NIT] row.item.max_uses !== unlimited

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,54 @@
...
+                    :formatter="maxUsesFormatter"
+                />
+                <h2 class="theme-h2 field-heading mb-2">
+                    Max uses
+                </h2>
+                <small>
+                    The number of times students can make use of this template.
+                    Note this also includes any deadlines that make use of this template.
+                    Use 0 to indicate unlimited usage.
+                </small>
```
mjvkeulen, line 49: The read mode of a template is missing an indication that it's configured for limited use. 
image

This is quite a crucial setting, so I would expect an indication before switching to edit mode. Possibly @engelhamer has an idea on how to visualize this?

What comes to my mind is a 0/x on the template menu in the assignment editor.

src/vue/src/components/template/TemplateEditSettings.vue
mjvkeulen: How about an infinite icon below/next to the input when 0 is the value? https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=infinite

mjvkeulen (changes requested):

src/django/VLE/migrations/0097_templatechain_max_uses.py
Larspolo: No, because then if it changes, we wont get a new migration, while we should have a new migration

src/django/VLE/models/journal.py
Larspolo: Feel like this is a NIT? Added it anyway, here was a lot of debugging, so thats also why its so spacious, easily being able to uncomment stuff

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -10,14 +10,22 @@
...
+                    class="ml-2"
+                >
+                    Used: {{ usedCount(row) }} / {{ row.item.max_uses }}
+                </small>
```
mjvkeulen, line 28: This takes up a bit more room than needed:
image

How about moving the <small> block inside the button block and separating the template name with a <br/>?

src/django/VLE/models/journal.py
Larspolo: Not really agreeing with the proposed structure.

Get templates could refer to also the archived or preset only templates or even only the used templates
Get available templates gets all available templates, do you wish to also include the used up templates, but other then that still available? Add that flag

Think its no big issue on both sides

src/django/VLE/views/entry.py
Larspolo: It also adds it inside the function, this example doesnt convince me. Especially because we still need a `get_available_templates` function anyway

src/django/VLE/migrations/0097_templatechain_max_uses.py
engelhamer: If what changes?

src/django/VLE/models/journal.py
engelhamer: Used up is confusibg to me. Would prefer smth like `limit_reached`

src/vue/src/components/template/TemplateEditSettings.vue
Larspolo: image
Looks good like this

Larspolo: Like this look good to me:
image

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: Would prefer a button for infinity, filling in 0 is not intuitive. I believe there is another instance where something similar happens (so: just offer the button on the front end, but keep 0 in the back end).

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: I can look into this 👍

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,54 @@
...
+                    min="0"
+                    step="1"
+                    :formatter="maxUsesFormatter"
+                />
+                <h2 class="theme-h2 field-heading mb-2">
+                    Max uses
+                </h2>
+                <small>
+                    The number of times students can make use of this template.
+                    Note this also includes any deadlines that make use of this template.
```
engelhamer, line 47: I don't like this: teachers would have to count the amount of times they add this template to a deadline manually? And it is not indicated while adding deadlines to the timeline that the number of uses is exceeded? What would intuitively happen if this is the case?

Would prefer the uses in preset deadlines to be independent of this setting.

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: @Larspolo: the 2 is a bit vague in this list in my opinion. Should we make it a table?

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -82,6 +93,14 @@ export default {
                 this.$set(template, '_showDetails', template.id === row.item.id && !row.detailsShowing)
             })
         },
+        usedCount (row) {
+            return this.nodes.map(
+                (n) => (n.entry && n.entry.template && n.entry.template.id === row.item.id ? 1 : 0),
+            ).reduce((a, b) => a + b)
+        },
+        isUsedUp (row) {
```
engelhamer, line 101: Again, limit reached iso used up would sound more intuitive.

src/vue/src/components/template/TemplateSelection.vue
Larspolo: Inside the button makes it also higthlight when hovering, makes it float in the center compared to the template name, and doesnt reduce it by far.

I can also move it next to the button:
image

src/vue/src/components/template/TemplateEditSettings.vue
Larspolo: image
Tried something, but my preference is without anything. Maybe the indication in the template section helps for you too?

src/django/test/test_entry.py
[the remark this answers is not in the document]
Larspolo: Can close this I suppose?

src/django/VLE/migrations/0097_templatechain_max_uses.py
Larspolo: The variable UNLIMITED

src/django/VLE/models/journal.py
Larspolo: Changed the naming schema

src/vue/src/components/template/TemplateEditSettings.vue
Larspolo: The 2 has a hover for more info

src/vue/src/components/template/TemplateEditSettings.vue
engelhamer: How about: leave empty to indicate unlimited usage?

engelhamer: The placeholder could then be infinity

src/django/VLE/models/template.py
```diff
@@ -88,6 +97,7 @@ def create_template_and_fields_from_data(
                 format=format,
                 preset_only=data['preset_only'],
                 allow_custom_categories=data['allow_custom_categories'],
+                max_uses=data.get('max_uses', 0),
```
mjvkeulen, line 100: Default to TemplateChain.UNLIMITED

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,63 @@
...
+                        name="infinity"
+                    />
+                </b-button>
+                <h2 class="theme-h2 field-heading mb-2">
+                    Max uses
+                </h2>
+                <small>
+                    The number of times students can freely use this template.
+                    Note this does not include any deadlines that make use of this template.
+                    Click on the infinite icon (∞) to indicate unlimited usage.
```
mjvkeulen, line 57: to allow unlimited use.

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -129,6 +167,11 @@ export default {
             type: Object,
         },
     },
+    data () {
+        return {
+            UNLIMITED: templateGlobals.UNLIMITED,
```
mjvkeulen, line 172: Haha I hate this too, this is purely for the linter right? (Unused import, as its not detected between the <template> tags)

src/vue/src/components/template/TemplateSelection.vue
```diff
@@ -82,6 +96,15 @@ export default {
                 this.$set(template, '_showDetails', template.id === row.item.id && !row.detailsShowing)
             })
         },
+        usedCount (row) {
+            console.log(this.nodes)
```
mjvkeulen, line 100: Left over (is this not part of our linter?), I am sure it's a warning during the build stage.

src/django/VLE/models/journal.py
mjvkeulen: Resolved for me

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,63 @@
...
+                />
+                <b-button
+                    class="green-button float-right mb-2 ml-2"
+                    :class="{'input-disabled': template.max_uses === UNLIMITED}"
+                    @click="template.max_uses = UNLIMITED"
+                >
+                    <icon
+                        name="infinity"
+                    />
+                </b-button>
```
mjvkeulen, line 50: This can be merged as an [input group](https://bootstrap-vue.org/docs/components/input-group).

This will nicely group the infinity icon with the input field.

src/vue/src/components/template/TemplateEditSettings.vue
```diff
@@ -1,28 +1,63 @@
...
+
+            <template #gated-settings>
+                <b-form-input
+                    id="max-uses-input"
+                    v-model="template.max_uses"
+                    type="number"
+                    class="float-right mb-2 ml-2"
+                    size="2"
+                    placeholder="∞"
+                    min="1"
```
mjvkeulen, line 38: This requires a bit more work to get right: 0 is still valid as input.

mjvkeulen, line 38: A value of 0 should be validated backend and frontend.

0 is effectively preset only, this is confusing.
- do we want to remove this distinction?
- Toggle free use to false when saving a max use with a 0 value?

mjvkeulen (changes requested):
I really like the use of the infinity icon and the template menu max use indication!

I slightly tweaked the template select max use display and added a tooltip on hover when disabled.

image

Feel free to revert e2412aea4106646184663b4774c74d184abc1af5 if you dont like it.

src/vue/src/components/template/TemplateEditSettings.vue
Larspolo: Guess if a teacher really wants it that way

Larspolo: Good if we postpone this to the PR with also the other validation checks?

mjvkeulen: As in just typing 0 will work

mjvkeulen (approved):

PR #1224, opened 2021-09-16
Fixed an issue where description only fields would be displayed as optional

mjvkeulen:


Larspolo (approved):

src/vue/src/components/template/TemplateField.vue
```diff
@@ -193,7 +193,7 @@ export default {
                 d: 'Date',
                 dt: 'Date Time',
                 s: 'Selection',
-                n: 'No Submission (Description Only)',
+                n: 'Description (No Submission)',
```
engelhamer, line 196: Now the mapping is broken. The letter is not first in the full term anymore.

engelhamer (approved):
This would be better solved by [EJ-736](https://ejournal.atlassian.net/browse/EJ-736)

PR #1226, opened 2021-09-17
Added validation rule for assessment model points possible

mjvkeulen:
Should not be possible to set points possible lower than any intermediate progress goal


Larspolo (commented):
Note to self: test this with LTI

src/django/test/test_assessment_model.py
```diff
@@ -31,6 +31,15 @@ def test_update_assessment_model_with_negative_points_possible(self):
         assert 'greater than or equal to 0' in resp['description'], \
             'Assessment model points possible should be gte 0, and the user is clearly informed if this is not so'
 
+    def test_update_assessment_model_with_points_possible_lt_a_progress_goals_target(self):
+        progress_goal = factory.ProgressPresetNode(format=self.assignment.format, target=10)
+        points_possible_lt_progress = deepcopy(self.update_params)
+        points_possible_lt_progress['points_possible'] = progress_goal.target - 1
+
+        resp = api.update(self, 'assessment_models', params=points_possible_lt_progress, user=self.teacher, status=400)
+        assert 'cannot be less than' in resp['description'], \
```
Larspolo, line 40: [NIT] would prefer checking for sth to do with "points possible"

Larspolo (approved):
Works. LTI does not update the points possible, even if it could. Think the LTI 1.3 branch does update it if it can

Frontend validation is a different error message though:
image

engelhamer (approved):

PR #1227, opened 2021-09-17
Fixed an issue where the LMS grade check was not based on the scaling of points possible

Larspolo:
EJ-1082 #close


src/django/VLE/utils/grading.py
```diff
@@ -173,8 +179,8 @@ def send_author_status_to_LMS(journal, author, left_journal=False, grade_only=Fa
     # Send student latest grade. But only send it when there are new entries OR grade changed
     response_student = None
     if journal.published_nodes.filter(entry__vle_coupling=VLE.models.Entry.NEEDS_GRADE_PASSBACK).exists() or \
-       journal.LMS_grade != grade:
-        if journal.LMS_grade != grade:
+       not math.isclose(journal.LMS_grade, normalize_grade(grade, author), abs_tol=0.00001):
+        if not math.isclose(journal.LMS_grade, normalize_grade(grade, author), abs_tol=0.00001):
```
mjvkeulen, line 183: Tolerance can be less strict:

abs_tol=0.00001 would imply 0.49990 != 0.49991 right? 

Not sure how much this matters for the problem we are trying to solve?

mjvkeulen (approved):

src/django/VLE/utils/grading.py
Larspolo: True, on the other hand, we can safely have this precision: https://docs.python.org/3/tutorial/floatingpoint.html

engelhamer (approved):

PR #1228, opened 2021-09-17
Fixed an issue where journal imports could cause errors after source was deleted

Larspolo:


src/django/VLE/serializers.py
```diff
@@ -1124,7 +1124,7 @@ def get_comments(self, entry):
         ).data
 
     def get_jir(self, entry):
-        if entry.jir:
+        if entry.jir and entry.jir.source:
```
mjvkeulen, line 1127: Still requires handling frontend

mjvkeulen: Nvm, I realised this could only occur after the approval of a JIR. Frontend does not work with already processed JIRs

mjvkeulen (approved):

engelhamer (approved):

PR #1231, opened 2021-09-29
Added an analytics dashboard

mjvkeulen:
# Analytics

Analytics introduces a dedicated dashboard to monitor the progress of students within an assignment. The Analytics dashboard can be accessed on the **assignment** page in the _Analytics_ card under the button **all**.

image

At the top of the Analytics dashboard is an overview panel of the assignment:

image

By default this shows some general statistics such as the total number of entries.

By clicking on the **Filters** button, you are presented with a wide range of filters and some statistics which directly relate to the selected filter configuration:

image

This allows for some relatively complex queries. For example: would you be only interested in which students of a specific group made use of a particular category? This is possible by combining the **Journal Filter** for _groups_ with the **Entry Filter** for _categories_.

It is also possible to gather more information about a specific deadline, on the lefthand side in the analytics dashboard you will find the hopefully familiar _timeline_. Simply select the deadline you are interested in. Please note it is only possible to select a single deadline at a time.

image

## Graphs

The analytics panel contains multiple graphs, each providing different insights into the progress of your students.

### Student activity

The student activity graph is a heatmap showing when your students were most active and/or most busy submitting entries.

image

Each block represents a specific day of the week, and the color represents how active students were that day. If you would like to know the exact number of activities that took place, just hover the blocks to see a tooltip that indicates the exact date and number of activities that took place.

heatmap

The Analytics dashboard comes with some additional group assignment functionality, such as a direct overview of the number of students who are not yet part of a journal, and the ability to message this group directly.

### Student activity histogram

Would you prefer a histogram over a heatmap? 

image

It is possible to configure the bin size manually, for example you might be interested whether it makes sense to set a deadline on a Friday or a Sunday? You can set the bin size to _days_ instead of _months_.

submission_on

Similarly, it is also possible to switch to a histogram of percentages or to include the submission status.

> Would you also like to display the entry draft status?
> Enable the **include drafts** settings in the **Global Filters** panel
>
> image

### Grades

It is possible to view a grade histogram of all the _journals_ or _entries_.

image

When looking at the _journal_ grades the assignment goal will be visualized as a green vertical line.
It is possible to switch between the entry or journal view by clicking on the **Entries** or **Journal** tab. Again it is possible to configure the bin size or to work with a percentage of the total number of journals or entries, rather than the absolute quantities. 

> Note: when looking at the grades of entries, any possible overall assessment grades or bonus points are not taken into account as these apply to the journal as a whole.

### Deadline submissions

It is possible to more closely inspect a specific deadline. First, select the deadline in the _timeline_. A new chart will display which shows the distribution of submission statuses.

deadlines_pie

It is possible to directly contact a specific group. For example, you might want to email all the students who have not yet submitted an entry for a specific deadline.

> Tip: If any students have not joined a group journal yet, but they should have already submitted a deadline, these students will also be outlined on the bottom of the graph if you would like to contact this group!


src/vue/src/components/analytics/AnalyticsOverview.vue
```diff
@@ -0,0 +1,221 @@
...
+                    class="mb-2"
+                    :class="{ 'input-disabled': filters.entryDeadline }"
+                    label="name"
+                    trackBy="id"
+                    :options="templates"
+                    :multiple="true"
+                    :searchable="true"
+                    :multiSelectText="filters.entries.templates.map(template => template.name).join(', ')"
+                    :showCount="false"
+                    :placeholder="filters.entryDeadline ? 'Active deadline filter' : 'Filter By Template'"
```
Larspolo, line 116: This is very minimal UI change, and I didn't know why I couldn't click it immediately (took me 3 to 5 seconds)

Could be "Cannot set template filter because deadline filter is active". Or hide it. Or gray out possibly?

image

src/vue/src/components/analytics/JournalUserListModal.vue
```diff
@@ -0,0 +1,214 @@
...
+            ref="table"
+            :per-page="perPage"
+            :current-page="currentPage"
+            responsive
+            :fields="fields"
+            :items="journals"
+            :filter="filter"
+            class="mt-2 mb-0 user-overview"
+            primary-key="id"
+            empty-filtered-text="No groups match the given search criteria."
```
Larspolo, line 35: Not exactly sure where it is, but the "No results for filter" text does not always appear. When typing "dsgf" in journal fullname/username filter, the "Deadline submissions" and "Grades" (journal tab only) do not show this text.

src/vue/src/components/analytics/graphs/DeadlineSubmissionStatusPieChart.vue
```diff
@@ -0,0 +1,219 @@
...
+            :options="combinedOptions"
+            :series="pieDataAndLabels.serie"
+            height="500"
+        />
+
+        <journal-user-list-modal
+            :id="'journal-user-list-modal'"
+            :journals="deadlineSelectedSliceData"
+            :journalIdUsersMap="journalIdUsersMap"
+        />
```
Larspolo, line 20: It is unclear to me that clicking on the piechart would result in this popup. Fine for now I think, but a different button (that can also be pressed with other filters in other graphs) would be preferred. Issue worthy?

Larspolo (commented):
Just some frontend comments to get you through the weekend. Look nice though!

src/vue/src/components/assignment/StatisticsCard.vue
```diff
@@ -54,6 +54,16 @@
...
         </div>
+
+        <b-button
+            v-if="$hasPermission('can_view_analytics')"
+            class="grey-button edit-button float-right mt-2"
+            variant="link"
+            @click="$router.push({name: 'Analytics'})"
+        >
+            <icon name="chart-bar"/>
+            All
```
Larspolo, line 65: Would expect the "All" be in the header of this menu. Sth like this:
image

src/django/test/factory/role.py
```diff
@@ -51,3 +51,5 @@ class TeacherRoleFactory(TaRoleFactory):
     can_view_grade_history = True
     can_post_teacher_entries = True
     can_manage_journal_import_requests = True
+    can_view_analytics = True
+    can_send_bulk_email_notifications = True
```
Larspolo, line 55: According to the mirgration, this should also be in the TARoleFactory. [Q] Do we want that? Both permissions as well?

src/django/test/test_analytics.py
```diff
@@ -0,0 +1,113 @@
...
+        self.deadline_entry.node.preset.save()
+        assert not e['late_submission'], 'Deadline entry submitted before the deadline due date is late'
+
+        # Restore dates
+        self.deadline_entry.last_edited = dt
+        self.deadline_entry.save()
+        self.deadline_entry.node.preset.due_date = deadline_dt
+        self.deadline_entry.node.preset.save()
+
+    def test_get_entry_analytics(self):
```
Larspolo, line 83: Missing assert statements. _Why_ is this being tested?

src/django/test/test_analytics.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self.deadline_entry.last_edited = self.assignment.due_date - relativedelta(days=1)
+        self.deadline_entry.save()
+        e = Entry.objects.filter(pk=self.deadline_entry.pk).annotate_entry_analytics().get()
+        assert not e['late_submission'], 'Deadline entry submitted before the assignment due date are on time'
+
+        deadline_dt = self.deadline_entry.node.preset.due_date
+        self.deadline_entry.node.preset.due_date = self.deadline_entry.last_edited - relativedelta(seconds=1)
+        self.deadline_entry.node.preset.save()
+        assert not e['late_submission'], 'Deadline entry submitted before the deadline due date is late'
```
Larspolo, line 75: *on time

Also, you need to re-get the entry

Also also, if the check above would yield true, this can never be false (right?) (given that this due date is earlier then the assignment due date)

src/django/test/test_assignment.py
```diff
@@ -2078,3 +2078,25 @@ def test_validate_lock_date():
...
+        teacher = assignment.author
+        student_1 = factory.Journal(entries__n=0, assignment=assignment).author
+        student_2 = factory.Journal(entries__n=0, assignment=assignment).author
+
+        params = {
+            'assignment_id': assignment.pk,
+            'students_only': True,
+        }
+        resp = api.get(self, 'assignments/participants', params=params, user=student_1)
+        assert len(resp['users']) == 0, 'Students cannot use the API'
```
Larspolo, line 2093: Would expect it to raise forbidden actually. Or only the APs where the student is in collaboration with (group journal), and at least returning their own data

src/django/test/test_event.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_create_sent_bulk_email_event(self):
+        Event.objects.create_sent_bulk_email_event(user=self.teacher, n_emails=3)
+        assert Event.objects.filter(users=self.teacher, type=Event.Type.SENT_BULK_EMAIL, extra__n_emails=3).exists(), \
+            'Event is succesfully created and extra contains the number of emails sent'
+
+    def test_create_user_login_via_jwt_event(self):
+        api.login(self, self.teacher)
+        assert Event.objects.filter(users=self.teacher, type=Event.Type.LOGIN_JWT).exists(), \
+            'On login a login event is created corresponding to the user that logged in'
```
Larspolo, line 39: Missing the login via LTI check

src/vue/src/components/analytics/graphs/SubmissionOnHistogram.vue
```diff
@@ -0,0 +1,279 @@
...
+            :series="histogramSeries"
+        />
+    </b-card>
+</template>
+
+<script>
+import VueApexCharts from 'vue-apexcharts'
+import analyticsUtils from '@/utils/analytics.js'
+import merge from 'lodash.merge'
+// import moment from 'moment'
```
Larspolo, line 46: left over

src/vue/src/components/analytics/graphs/SubmissionOnHistogram.vue
```diff
@@ -0,0 +1,279 @@
...
+                    onTime: analyticsUtils.scaleAsPercentage(histogramSeriesData.onTime, this.data.length, 10),
+                    late: analyticsUtils.scaleAsPercentage(histogramSeriesData.late, this.data.length, 10),
+                    drafts: analyticsUtils.scaleAsPercentage(histogramSeriesData.drafts, this.data.length, 10),
+                    combined: analyticsUtils.scaleAsPercentage(histogramSeriesData.combined, this.data.length, 10),
+                }
+            }
+
+            return histogramSeriesData
+
+            // return {
```
Larspolo, line 216: left over

src/vue/src/components/analytics/graphs/SubmissionOnHistogram.vue
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            return series
+        },
+    },
+    methods: {
+        getIndex (m) {
+            return Math.floor(m.diff(this.minMoment, this.submittedOnConf.step, true))
+        },
+        // mapValuesToDatePairsArr (arr) {
```
Larspolo, line 268: left over

src/vue/src/components/analytics/graphs/WeekDayHeatmap.vue
```diff
@@ -0,0 +1,194 @@
...
+                },
+                // https://hihayk.github.io/scale/#5/5/75/75/0/0/25/25/216cce/33/108/206/white
+                colors: [themeColors.blue], // this would specificy a distribution among a single color
+                plotOptions: {
+                    heatmap: {
+                        shadeIntensity: 0.5,
+                        radius: 2,
+                        distributed: true,
+                        useFillColorAsStroke: false, // Do not fill the gaps between blocks
+                        // colorScale: {
```
Larspolo, line 63: left over

src/vue/src/components/analytics/graphs/WeekDayHeatmap.vue
```diff
@@ -0,0 +1,194 @@
...
+            return merge({}, this.options, this.heatMapOptions)
+        },
+        // TODO: Move to analytics.vue shared with SubmissionOn
+        moments () {
+            return this.data.map((elem) => moment(new Date(elem.date)))
+        },
+        minMoment () {
+            return moment.min(this.moments).startOf('isoWeek')
+        },
+        // maxMoment () {
```
Larspolo, line 105: left over

src/vue/src/views/Analytics.vue
```diff
@@ -0,0 +1,490 @@
...
+            }
+
+            if (this.studentsWithoutJournal.length) {
+                stats.push({
+                    title: 'Students without journal',
+                    value: this.filteredStudentsWithoutJournal.length,
+                })
+            }
+
+            // QUESTION ANALYTICS: Is the average journal entry grade desired? (Disregarding entry specific filters)
```
Larspolo, line 365: Dont think so

Larspolo (changes requested):

src/vue/src/components/analytics/graphs/GradeHistograms.vue
```diff
@@ -0,0 +1,135 @@
...
+            </b-form-checkbox>
+
+            <h2 class="theme-h2">
+                Grades
+            </h2>
+        </template>
+
+        <b-tabs v-model="activeHistogram">
+            <b-tab
+                title="Entries"
```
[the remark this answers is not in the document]
mjvkeulen: This is intentional, the number of entries on the journal tab specifically applies to the journal filter.

Would you prefer to see a duplication of the entry filter number of the entry tab?

src/vue/src/components/analytics/graphs/GradeHistograms.vue
```diff
@@ -0,0 +1,135 @@
...
+            <b-form-checkbox
+                v-model="gradeHistogramConf.asPercentage"
+                class="float-right mt-1"
+                inline
+            >
+                Percentage
+            </b-form-checkbox>
+
+            <h2 class="theme-h2">
+                Grades
```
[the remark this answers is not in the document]
mjvkeulen: Annoying bug between b-tabs and apexchart. Fixed with hack

src/vue/src/components/analytics/AnalyticsOverview.vue
mjvkeulen: It is greyed out (disabled), compare it to filter by category. I believe this is default disabled styling across the app.

The text is also different, I can change it but `Cannot set template filter because deadline filter is active` is a bit lengthy.

src/vue/src/components/analytics/JournalUserListModal.vue
mjvkeulen: I think you are confused about where this is taking place.

image

src/vue/src/components/analytics/graphs/DeadlineSubmissionStatusPieChart.vue
mjvkeulen: So you raise two seperate issues:

- It is not intuitive that the deadline submission chart is interactive
- The modal to communicate with students would be desired in other locations as well.

What do you suggest?

src/django/VLE/models/entry.py
```diff
@@ -61,6 +62,48 @@ def annotate_name(self):
...
         )
 
+    def annotate_entry_analytics(self):
+        return self.values(
+            'is_draft',
+            'title',
+            'creation_date',
+            'last_edited',
+        ).annotate(
+            id=F('pk'),
```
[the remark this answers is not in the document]
mjvkeulen: I don't quite follow, this is serialized as is and for some reason we use `id` frontend

src/django/VLE/models/event.py
```diff
@@ -0,0 +1,75 @@
...
+
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+
+from .user import User
+
+
+class EventQuerySet(models.QuerySet):
+    def create_sent_bulk_email_event(self, user: User, n_emails: int) -> 'Event':
```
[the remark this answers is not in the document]
mjvkeulen: It is a way to handle definitions later on in the file (to prevent circular missing types)

src/django/VLE/tasks/email.py
```diff
@@ -220,3 +221,43 @@ def send_email_feedback(user_pk, topic, ftype, feedback, user_agent, url, file_c
...
+            reply_to=[f'{sending_user.full_name} <{sending_user.email}>'],
+            to=[user.email],
+        )
+        email.attach_alternative(html_content, 'text/html')
+
+        try:
+            email.send()
+            success_user_pks.append(user.pk)
+        except Exception as e:
+            capture_exception(e)
```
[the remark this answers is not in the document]
mjvkeulen: The default exception will have all this context quite high up in its stacktrace no?

src/django/VLE/utils/responses.py
```diff
@@ -5,13 +5,45 @@
...
+from django.utils.duration import duration_iso_string
+from django.utils.functional import Promise
+from django.utils.timezone import is_aware
+
+
+class EJDateFormatDjangoJSONEncoder(DjangoJSONEncoder):
+    """
+    Override to follow settings datetime and date format
+    """
+    def default(self, o):
```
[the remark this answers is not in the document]
mjvkeulen: The override paths are tested rest is well tested by lib already

src/django/VLE/views/analytics.py
```diff
@@ -0,0 +1,23 @@
...
+from VLE.models import Assignment, Entry, Journal
+
+
+@api_view(['GET'])
+def entry_analytics(request):
+    assignment_id, = utils.required_typed_params(request.query_params, (int, 'assignment_id'))
+
+    assignment = Assignment.objects.get(pk=assignment_id)
+
+    request.user.check_permission('can_view_analytics', assignment)
```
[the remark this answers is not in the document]
mjvkeulen: I would expect this to be embedded in the permission check.

Also what statistics are available for an unpublished assignment? I don't believe we allow unpublish once started.

src/django/VLE/views/analytics.py
```diff
@@ -0,0 +1,23 @@
...
+
+    viewable_journals = Journal.all_objects.filter(
```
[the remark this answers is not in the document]
mjvkeulen: Because all the costly annotations are not needed?

Our structure of doing `allowed_journals` followed by a lot of additional annotations is a bit sucky for performance.

src/django/VLE/views/assignment.py
```diff
@@ -656,6 +658,47 @@ def participants_without_journal(self, request, pk):
 
         return response.success({'participants': participants_without_journal})
 
+    @action(methods=['get'], detail=False)
+    def participants(self, request):
```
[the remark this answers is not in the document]
mjvkeulen: So the url currently is /assignments/participants/ which is not bad tbh.

This is also in line with all the other actions in the assignment view, like `teacher_entries`, `participants_without_journal` etc. We can move this to a more broad issue to standardize the API or smth.

src/django/VLE/views/assignment.py
```diff
@@ -656,6 +658,47 @@ def participants_without_journal(self, request, pk):
...
+
+        if students_only:
+            aps = AssignmentParticipation.objects.filter(
+                assignment=assignment,
+                user__participation__role__can_have_journal=True,
+            )
+        else:
+            aps = AssignmentParticipation.objects.filter(assignment=assignment)
+
+        validated_user_data = (
```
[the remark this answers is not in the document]
mjvkeulen: I expected a list of flat user data, not AP data.

User serializer does not have journal as a field, and groups touch on `for_course` which I wanted to stay away from. 

Your point of keeping output consistent with serializers where possible stands, however. Not sure if I would move this to its own serializer, leave it as is or flatten the data frontend.

src/django/VLE/views/email.py
```diff
@@ -170,9 +170,45 @@ def send_feedback(request):
...
+
+    email_tasks.send_email_on_behalf_of_user_to_users.delay(
+        request.user.pk,
+        list(validated_user_pks),
+        subject,
+        content,
+        context_url=utils.gen_url(assignment=assignment, user=request.user),
+    )
+
+    Event.objects.create_sent_bulk_email_event(user=request.user, n_emails=len(user_ids))
```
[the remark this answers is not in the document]
mjvkeulen: No external dependencies or complexity for something which has simple requirements to satisfy our needs.

src/django/VLE/views/email.py
```diff
@@ -170,9 +170,45 @@ def send_feedback(request):
...
+        (int, 'user_ids'),
+        (str, 'subject'),
+        (str, 'content'),
+    )
+
+    assignment = Assignment.objects.get(pk=assignment_id)
+    request.user.check_permission('can_send_bulk_email_notifications', assignment)
+
+    validated_user_pks = (
+        AssignmentParticipation.objects
```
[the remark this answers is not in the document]
mjvkeulen: Is the combination of the permission check and viewable by not sufficient?

src/django/test/factory/role.py
mjvkeulen: I thought so

src/django/test/test_analytics.py
mjvkeulen: I am not sure I agree, the asserts are sufficiently simple (does the serialized data key match with the expected serialized object value), that asserts messages repeating the code line do not contribute much

src/vue/src/components/analytics/graphs/SubmissionOnHistogram.vue
mjvkeulen: All the commented code can be used to transform the series into a time series instead of a discrete histogram with custom buckets.

Its just that I didn't like the looks of how the library then handles the x-axis labels.

src/vue/src/components/analytics/graphs/GradeHistograms.vue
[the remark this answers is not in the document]
Larspolo: A tooltip would also suffice

src/vue/src/components/analytics/AnalyticsOverview.vue
Larspolo: Yeah, I think the whole app would improve with a better distinction

src/vue/src/components/analytics/JournalUserListModal.vue
Larspolo: Yeah, the comment is not placed on the correct line, the explenation is about the analytics, not the email popup

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: Yeah, we should make it an issue. But I dont like making it worse because elsewhere it is also done

src/django/VLE/views/email.py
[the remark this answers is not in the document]
Larspolo: What is our need? I would for example like to be able to add levels or categories

src/django/VLE/views/analytics.py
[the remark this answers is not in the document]
Larspolo: Its not, but I guess it works for now. Also assigned groups are not checked now

src/django/VLE/views/email.py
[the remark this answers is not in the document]
Larspolo: Now you can send emails to users that cannot see the assignment yet, but you can see them

src/vue/src/components/analytics/JournalUserListModal.vue
Larspolo: image
Screenshot for clarity

src/vue/src/components/analytics/graphs/DeadlineSubmissionStatusPieChart.vue
Larspolo: Dont have a good idea. But as a teacher I see this feature being easily overlooked. Or when they do find out once, then the next time not knowing how to do it

src/django/VLE/tasks/email.py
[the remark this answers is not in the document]
Larspolo: But it may miss some context, but I noticed we dont do it elsewhere either, changed to a NIT

src/django/VLE/utils/responses.py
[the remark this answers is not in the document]
Larspolo: Why does it not use `return super().default(o)` for those cases? Codecov detects only the first statement as visited

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: Would prefer then requesting the AP data, and then doing `ap.user.username`, so the structure is kept

src/django/test/test_analytics.py
Larspolo: Well, not for every assert. But I would expect sth like:
- "When retrieving the entry analytics, all the entries within a single assignment should be returned"
- "Entries should be serialized with all the variables that are needed for analytics"
- "Preset node should be none for unlimited entries"

I also think you are more testing the annotate function. This can also be done separately to make it more clear. And then here only test if the correct entries get annotated

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
mjvkeulen: I dislike requesting AP data, this should be User data imo.

src/django/VLE/views/email.py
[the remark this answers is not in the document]
mjvkeulen: Add it as a choices field once used?

src/django/VLE/views/email.py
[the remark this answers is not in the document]
mjvkeulen: So if you are the legitimate supervisor of that user, you can send that user an email despite them not being able to see the assignment yet (via a manual API call at that)

src/django/test/test_assignment.py
mjvkeulen: An API not returning data for users which are not authorized seems fine by me as well, this also omits the repetition of a check which is already partly done via the query for the data itself.

src/django/VLE/views/assignment.py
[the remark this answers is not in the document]
Larspolo: Could be a users api call then? `/users?assignment_id=x`

src/django/VLE/views/email.py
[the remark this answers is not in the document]
Larspolo: So then we are building our own logging module

src/django/VLE/views/email.py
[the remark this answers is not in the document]
Larspolo: Yup, like we also do with the notificaiton system, only send notifications if the user can see the context

src/vue/src/components/analytics/JournalUserListModal.vue
mjvkeulen: This is quite specific. Only for the deadline submission graph, and only when you filter journals down to 0. E.g. for an entry filter, this is shown:

image

mjvkeulen: Should be addressed now

image

src/django/VLE/utils/responses.py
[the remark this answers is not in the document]
mjvkeulen: Only the blocks for Date and DateTime are changed rest is default. This is something we already do in the serializers but rather weakly compared to this method.

src/django/test/test_assignment.py
Larspolo: Its different from the rest of the API. Why do this now? It may cause confusion

src/django/VLE/views/email.py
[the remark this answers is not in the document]
mjvkeulen: Note the "once used".

Even if you were to introduce a lib like structlog, we would still need to write similar logic to store it within the DB.

src/django/VLE/views/email.py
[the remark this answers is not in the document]
mjvkeulen: This hardly applies to this example, the teacher decided to send this user an email and was verified that he has the right to do so. The rest is minor, and if you want to shield your example prob best to tack it onto the `('can_send_bulk_email_notifications', assignment)` check.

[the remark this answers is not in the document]
mjvkeulen: I would even argue that if in this scenario the user cannot see the resource from some reason, it is better that they DO get an email so they might respond with WTF I can't even see the resource.

src/django/test/test_assignment.py
mjvkeulen: - Allows can_view to function both as a guard on the filter as on the result of the API
- I disagree this differs from the rest of the API, there are many places where the result is reduced or zero depending on your permission.

src/vue/src/components/analytics/JournalUserListModal.vue
Larspolo: Not yet the "Grades" chart in the journal tab

image

src/django/VLE/utils/responses.py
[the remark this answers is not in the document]
Larspolo: Would this not suffice then?
```
    def default(self, o):
        # See "Date Time String Format" in the ECMA-262 specification.
        if isinstance(o, datetime.datetime):
            return o.strftime(settings.ALLOWED_DATETIME_FORMAT)
        elif isinstance(o, datetime.date):
            return o.strftime(settings.ALLOWED_DATE_FORMAT)
        else:
            return super().default(o)
```

src/django/VLE/views/email.py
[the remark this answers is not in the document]
Larspolo: > prob best to tack it onto the ('can_send_bulk_email_notifications', assignment) check.

The `check_permission` function only checks for the permission itself, this is done all over the backend. Thats where the improved permissions PR is coming from

> I would even argue that if in this scenario the user cannot see the resource for some reason, it is better that they DO get an email so they might respond with WTF I can't even see the resource.

Good point that this is useful to point out. Would then even say that immediately informing the teacher is better then informing the student, that then has to inform the teacher. We could return a list of all users that are not able to view the assignment, and then display that to the teacher with a popup "Remove the users from the list | Send anyway"

src/django/test/test_assignment.py
Larspolo: > there are many places where the result is reduced or zero depending on your permission.

And no error message is giving? Which?

src/django/VLE/views/email.py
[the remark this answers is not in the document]
mjvkeulen: I agree that embedding this nicely for the teacher would be better. 

But, leaving this as a responsibility for the student for now would work fine for now and we can make an issue to resolve this more cleanly when it does occur.

src/django/test/test_assignment.py
mjvkeulen: First place I looked: groups -> list -> can_only_view_users_of_own_groups -> group not part of own -> yields empty data. 

Works on the same principle where viewable_by guards the query.

Larspolo: Thats different. There you actually do get all the groups that they can see. If we go by that case, we should expect all the APs from the students' (group) journal as well

engelhamer:
Then toUsers can be omitted

Op di 12 okt. 2021 09:05 schreef mjvkeulen ***@***.***>:

src/vue/src/components/assignment/StatisticsCard.vue
```diff
@@ -1,11 +1,23 @@
...
+                @click="$router.push({name: 'Analytics'})"
+            >
+                <icon name="chart-bar"/>
+                All
+            </b-button>
+
+            <h3
+                class="theme-h3"
+            >
+                Analytics
```
Larspolo, line 17: [NIT] The title changed, but the filename is still the same

src/django/VLE/views/email.py
```diff
@@ -170,9 +170,43 @@ def send_feedback(request):
...
+        request.user.pk,
+        list(validated_user_pks),
+        subject,
+        content,
+        context_url=utils.gen_url(assignment=assignment, user=request.user),
+    )
+
+    Event.objects.create_sent_bulk_email_event(user=request.user, n_emails=len(user_ids))
+
+    return response.success(description='Form succesfully received, we are working on notifying all users!')
```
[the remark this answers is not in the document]
engelhamer:
```
    return response.success(description='Message sent!')
```
'Message sent' sounds much better and I really see no reason to keep this weird phrasing.

src/vue/src/api/email.js
```diff
@@ -0,0 +1,7 @@
+import auth from '@/api/auth.js'
+
+export default {
+    sendEmailOnBehalfOfUserToUsers (data, connArgs = auth.DEFAULT_CONN_ARGS) {
```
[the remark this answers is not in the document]
engelhamer: No response from @mjvkeulen yet.

src/vue/src/components/analytics/StatsDisplay.vue
```diff
@@ -0,0 +1,56 @@
...
+}
+</script>
+
+<style lang="sass">
+@import '~sass/modules/colors.sass'
+
+.stat-display
+    text-align: center
+    &:not(:last-child)
+        border-right: 3px solid $theme-dark-grey
```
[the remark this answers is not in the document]
engelhamer: Not NIT, just consistency of the design? Same as capitalisation of sentences etc.

src/vue/src/components/analytics/graphs/SubmissionOnHistogram.vue
```diff
@@ -0,0 +1,298 @@
...
+                }
+            })
+
+            const onTimeValuesArr = Object.values(onTimeBucketValMap)
+            const lateValuesArr = Object.values(lateBucketValMap)
+            const draftValuesArr = Object.values(draftBucketValMap)
+
+            return {
+                onTime: onTimeValuesArr,
+                late: lateValuesArr,
```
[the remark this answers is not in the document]
engelhamer: Yes. Do exactly the same as in this gif (toggle the checkbox) and then 'late' appears in the legend, even though it is not in the graph. Hovering on 'late' in the legend shows nothing highlighted, but blurs the blue part.

This is a bit confusing in my opinion.

engelhamer (approved):
Fixed some styling bugs I encountered during review. Some minor comments still open but fine to merge.

src/vue/src/components/columns/TimelineTwoColumns.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+    <b-row noGutters>
+        <b-col
+            class="combined-content-timeline-page"
+        >
+            <b-row noGutters>
+                <b-col
+                    md="12"
+                    xl="4"
```
Larspolo, line 9: [NIT] Would rather keep the 3 columns for the timeline. If this is wide enough for the other pages, I dont see why it wouldnt be wide enough here. Also it keeps the consistency between the analytics and other timeline pages

engelhamer: The graphs can take up a lot of horizontal space. In this case there is also no panel in the right hand column, so why not use the available space that would otherwise be left blank?

Larspolo: For consistency

PR #1232, opened 2021-10-02
Improved the UX for editing entries with slow internet

Larspolo:
## Description
When users would take a long time to send their changes to the server, it could lead to data being set back. This wont happen now, and it will automatically send the newer data to the server when a (slow) request has finished
EJ-1047 #close
https://ejournal.atlassian.net/browse/EJ-1047


src/django/VLE/utils/entry_utils.py
```diff
@@ -40,17 +40,7 @@ def check_fields(template, content_dict):
...
 
 
-def validate_node(node, template):
-    if not (node.preset and node.preset.forced_template == template):
-        if (
-            node.preset
-            and node.preset.forced_template.chain == template.chain
-            and template.archived
-        ):
-            raise VLEBadRequest('Your teacher has updated the template for this entry, please reload and try again.')
```
mjvkeulen, line 50: Q) How is this case now resolved? This scenario still holds: the front-end data can be stale after the teacher made an update but the student fetched the nodes before the update.

src/django/VLE/utils/entry_utils.py
```diff
@@ -40,17 +40,7 @@ def check_fields(template, content_dict):
...
-        else:
-            raise VLEBadRequest('Provided template is not used by the deadline, please reload and try again.')
```
mjvkeulen, line 52: Q) How is this case now resolved?

src/django/VLE/views/entry.py
```diff
@@ -60,42 +43,26 @@ def create(self, request):
         if len(journal.needs_lti_link) > 0:
             return response.forbidden(journal.outdated_link_warning_msg)
 
-        # Check if the template is available
-        if not (node_id or assignment.format.template_set.filter(archived=False, preset_only=False,
-                                                                 pk=template.pk).exists()):
-            return response.forbidden('Entry template is not available.')
+        with transaction.atomic():
+            # Only unlimited nodes provide template ids
```
mjvkeulen, line 47: [D] Is this quirky good, or should we simply add an explicit flag (unlimited: True) to the API?

src/django/VLE/factory.py
```diff
@@ -198,20 +198,10 @@ def make_node(journal, entry=None, type=VLE.models.Node.ENTRY, preset=None):
...
-
-    if not template.chain.allow_custom_title:
-        title = None
-
-    if not template.chain.allow_custom_categories or category_ids is None:
-        params['category_ids'] = list(template.categories.values_list('pk', flat=True))
-    elif category_ids:
-        params['category_ids'] = category_ids
-
+def make_entry(template, author, node):
```
mjvkeulen, line 201: Since this now only creates a blank draft entry (does not set title, or custom categories etc), which I think is good and much more befitting of our frontend flow, should we just rename this to make_entry_draft? I initially had to rethink why all the checks could be omitted, this would make it clear at a glance to me.

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -44,6 +44,11 @@ export default {
             type: Object,
         },
     },
+    data () {
+        return {
+            NEW_ENTRY_ID: -1,
```
mjvkeulen, line 49: Q) What does this represent exactly?

mjvkeulen (changes requested):
Good addition, clean flow with a single promise chain. 

Just some questions

src/vue/src/components/timeline/TimelineNodeCircle.vue
mjvkeulen: I think it simply means not yet created?

src/django/VLE/utils/entry_utils.py
Larspolo: We dont pass the template anymore. If the user creates an entry, it will be linked to the template that is then the current template. This should no longer be a problem backend. Although frontend, the template wont be reloaded, will need to fix that

src/django/VLE/utils/entry_utils.py
Larspolo: We dont pass the template anymore. If the user creates an entry, it will be linked to the template that is then the current template. This should no longer be a problem backend. Although frontend, the template wont be reloaded, will need to fix that

src/vue/src/components/timeline/TimelineNodeCircle.vue
Larspolo: Yes

src/vue/src/components/entry/Entry.vue
```diff
@@ -294,142 +298,170 @@ export default {
...
         },
-        autoSave () {
-            this.autoSaveMessage = 'Saving...'
-            this.saveChanges(true)
-                .then(() => {
-                    this.autoSaveMessage = 'Saved as draft'
-                })
-                .catch(() => {
-                    this.autoSaveMessage = 'Failed to save'
+        saveChanges () {
```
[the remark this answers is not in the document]
engelhamer: This field count also includes description-only fields, which may be confusing to the user.

E.g. 
image

engelhamer (approved):
Required field error msg could be improved for descriptions, but otherwise alright.

PR #1233, opened 2021-10-04
Updated application monitoring structure

mjvkeulen:
- Update Sentry and project structure
- Adds a tunnel redirect frontend to be compatible with add blockers like uBlock Origin


src/vue/src/helpers/sentry.js
```diff
@@ -33,9 +34,26 @@ function beforeSend (event, hint) { // eslint-disable-line no-unused-vars
...
+                tracingOrigins: [
+                    'localhost',
+                    'ejournal.app',
+                    window.location.origin,
+                    /^\//,
+                ],
+            }),
+        ],
         beforeSend,
+        // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
```
engelhamer, line 54: What to do with this?

mjvkeulen: Worth to investigate later, its why I left it in

mjvkeulen: Moved to issue

Larspolo (approved):

engelhamer (approved):

PR #1240, opened 2021-10-09
Improved the category preview in the timeline

engelhamer:
category-display

EJ-928 #done

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1240)


Larspolo:
Maartens solution also looks nice

engelhamer:
> I dislike that there is no overflow indication, imo it should be clear to the user when some additional categories are cropped.

There is a fade.

mjvkeulen:
> Maartens solution also looks nice

@Larspolo you need to be decided on this I feel.

- @engelhamer prefers the verbosity of showing the categories in full (max one line when not selected).
- I prefer the more sober look of just showing the category colors. Adding and filtering categories is still verbose. If someone forgot what a color represents, they effectively have to just hover/click on the node to see the categories in full.

> Good enough for me, but I think we miss a core issue now. If a user would like look for a specific category in a quick overview (without using the filter system), this cannot work

The latter has the added benefit that @Larspolo his previous issue is (partically) also resolved. 
We can always show all categories corresponding to an activity when just looking for an overview.

Larspolo:
Agree with Maarten

engelhamer:
Implemented dot preview.

category-display

src/vue/src/components/category/EntryCategories.vue
```diff
@@ -2,7 +2,7 @@
     <div
         v-if="($store.getters['category/assignmentCategories'].length
             && (editable || entry.categories)) || (displayOnly && template.categories.length > 0)"
-        class="p-2 background-light-grey round-border d-inline-block"
+        class="p-2 bordered-content d-inline-block"
```
engelhamer, line 5: This change was leaked from #1270

Larspolo (approved):

mjvkeulen:
I love the look, would you be in favor of expanding the categories on hover as well? @engelhamer

mjvkeulen (approved):

PR #1242, opened 2021-10-14
Similarly phrase the overall journal assessment action button text to the entry grade actions

mjvkeulen:
For entry grading actions we use:
- Save Grade
- Save & Publish Grade

For assessment we now use:
- Save Assessment
- Save & Publish Assessment

At least now it's consistent, but should we not drop "Grade" and "Assessment"? It should be clear from the context what object the action refers to.


engelhamer:
We added 'grade' to the grade box since there is no other indication that the number you enter there is a grade (even though the context implies it).

Assessment has a dedicated section that you have to navigate to. The name is in the button to this section as well as in the modal header. Therefore it is not really required.

engelhamer (approved):
Still fine by me though.

PR #1246, opened 2021-10-22
Added the ability to update existing entries with template changes

Larspolo:
## Description
Simple, yet effective

image

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1246)


src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -147,8 +199,18 @@ export default {
                     .then((template) => {
                         this.templateCreated({ template, fromPresetNode: this.template.fromPresetNode })
                     })
-            } else {
-                this.updateTemplate({ id: this.template.id, data: this.template, aID: this.$route.params.aID })
+            } else if (
+                this.updateExistingEntries
+                    || window.confirm('Are you sure you do not want to update the existing entries?')
```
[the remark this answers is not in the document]
mjvkeulen: I agree a warning is appropriate! 

However:

> There should be no confirmation **when there is no change** that will affect existing entries

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -65,18 +65,72 @@
...
+                    Template is in use by {{ template.used === 1 ? 'one entry' : `${template.used} entries` }}.
+                    <ul>
+                        <li v-if="isTemplateCategoryListDirty(dirtyCheckableTemplate)">
+                            Changes to the default categories will also be changed for the existing entries.
+                        </li>
+                        <li v-if="isTemplateFieldSetDirty(dirtyCheckableTemplate)">
+                            Removed fields are also removed from existing entries.
+                        </li>
+                        <li v-if="isTemplateFieldSetDirty(dirtyCheckableTemplate)">
+                            Field type changes will clear the corresponding entry content.
```
mjvkeulen, line 105: Whereas this is to the point, I don't think this will always be immediately clear. Let's add an example:

Field type changes (e.g. changing text to video) will clear the corresponding entry content.

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -65,18 +65,72 @@
...
+                        </li>
+                    </ul>
+                    If you choose to <i>not</i> update existing entries,
+                    those entries won't recieve any (future) changes to their template.
```
mjvkeulen, line 109:
```
Entries that are <b>not</b> updated are exempt from any (future) template updates.
```

src/vue/src/store/modules/template.js
```diff
@@ -71,9 +71,9 @@ const mutations = {
 }
 
 const actions = {
-    list (context, { aID, force = false, connArgs = auth.DEFAULT_CONN_ARGS }) {
+    list (context, { aID, includeUses = false, force = false, connArgs = auth.DEFAULT_CONN_ARGS }) {
         function fn () {
-            return auth.get('templates', { assignment_id: aID }, connArgs)
+            return auth.get('templates', { assignment_id: aID, include_uses: includeUses }, connArgs)
```
mjvkeulen, line 76: include_uses has to be made part of the cache key (`aID-${includeUses}`) or smth

src/vue/src/router/index.js
```diff
@@ -200,7 +200,7 @@ router.afterEach((to) => {
         store.dispatch('assignment/retrieve', { id: to.params.aID })
         store.dispatch('presetNode/list', { aID: to.params.aID })
         store.dispatch('category/list', { aID: to.params.aID })
-        store.dispatch('template/list', { aID: to.params.aID })
+        store.dispatch('template/list', { aID: to.params.aID, includeUses: true })
```
mjvkeulen, line 203: This does not need the template's uses on each page, right?

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -65,18 +65,72 @@
...
+                            class: 'red-button',
+                        },
+                    ]"
+                    class="float-right ml-3"
+                />
+                <h2 class="theme-h2 field-heading mb-2">
+                    Update existing entries
+                </h2>
+                <small>
+                    Template is in use by {{ template.used === 1 ? 'one entry' : `${template.used} entries` }}.
```
mjvkeulen, line 96: [D] Is it confusing `used` only relates to the current template version?

So if we have an archived version this is off (but still technically accurate of course)?

@engelhamer

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -98,6 +153,8 @@ export default {
         return {
             nameInvalidFeedback: null,
             nameInputState: null,
+            updateExistingEntries: false,
```
mjvkeulen, line 156: [D] I think this can be on by default.

I believe it is more likely template changes should be reflected in the final entries than not, despite our original design philosphy.

src/vue/src/components/template/TemplateEdit.vue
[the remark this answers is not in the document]
Larspolo: Ah missread, but that should be fixed now too :)

src/vue/src/components/template/TemplateEdit.vue
```diff
@@ -65,18 +65,72 @@
...
+                v-if="create"
                 class="green-button float-right"
                 @click="finalizeTemplateChanges"
             >
-                <icon :name="(create) ? 'plus' : 'save'"/>
-                {{ (create) ? 'Add Template' : 'Save' }}
+                <icon name="plus"/>
+                Add Template
+            </b-button>
+            <template v-if="canUpdateExistingEntries">
```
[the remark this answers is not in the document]
Larspolo: Should be fixed now

Open template that is in use:
image
Add a field:
image
See the warning in the footer

src/vue/src/components/template/TemplateEdit.vue
Larspolo: Sure, we could even make use of the suggestions in this repo as well for these kind of changes

src/vue/src/components/template/TemplateEdit.vue
Larspolo: I like this one, changed it

src/vue/src/store/modules/template.js
Larspolo: I simply set it to always true. Else we would store templates twice, no need for that

src/vue/src/router/index.js
Larspolo: I am in favour of retrieving the max uses anyway, as it has minimal effect on the duration of the request compared to having to store all templates twice: one time calling it with uses, and one time calling it without

src/vue/src/components/template/TemplateEdit.vue
Larspolo: "Active use" maybe? But that might also confuse people. I am in favour of not postponing this PR for this discussion

src/vue/src/components/template/TemplateEdit.vue
Larspolo: Sure, then we can also remove the prompt, as it already requires an extra click. I opted for false as default here because of two reasons
- Mimics the current situation
- To force teachers to make an active choice: either click on yes, or receive a window.confirm box

Also reminder that this PR is mostly as an improvement on the state of art, not a final solution

src/django/VLE/views/template.py
```diff
@@ -12,14 +12,20 @@
...
 
         assignment = Assignment.objects.select_related('format').get(pk=assignment_id)
 
         if not request.user.can_view(assignment):
             return response.forbidden('You are not allowed to view this assignment.')
 
-        serializer = TemplateSerializer(
-            TemplateSerializer.setup_eager_loading(
+        if include_uses and request.user.has_permission('can_edit_assignment', assignment):
+            serializer_class = EditTemplateSerializer
```
mjvkeulen, line 23: Should this not be hit by some test?

src/django/VLE/models/template.py
```diff
@@ -266,12 +266,56 @@ def validate_field_set():
...
+        # Update current fields
+        # Should also delete the content in the entry if type changes
+        fields = Field.objects.filter(template=self)
+        for field in fields:
+            new_field_content = [f for f in field_set if f.get('id', -1) == field.pk]
+            if new_field_content:
+                new_field_content = new_field_content[0]
+            else:
+                # Field not found in content, ignoring
+                continue
```
mjvkeulen, line 284: Also seems reasonable to include in a test?

Larspolo: Kinda impossible actually, if the field has no content, then it is removed. Removed fields are already deleted above. Do you wish to remove this line? I think it is nice to keep in case someone moves the field deletion elsewhere

engelhamer:
Since the wysiwyg template editor makes field types immutable, it may be possible to always update the template for existing entries and still keep existing work. I'll check that one out first.

src/django/VLE/models/template.py
mjvkeulen: I would expect a crash or atleast a log in this situation. So I would opt for removing this

src/django/VLE/models/template.py
```diff
@@ -266,12 +266,56 @@ def validate_field_set():
         validate_categories()
         validate_field_set()
 
+    def update_field_set(self, field_set):
+        field_keys = ['type', 'title', 'location', 'required', 'description', 'options']
+        # Delete old fields
+        # Should also delete the content in the entry
```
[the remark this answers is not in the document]
engelhamer: I would propose the following, to avoid deleting content from student entries:

- Set the field in the template to an `archived` state
- Only serialize it if it has content for an entry already, and make it optional in such cases

src/django/VLE/models/template.py
```diff
@@ -266,12 +266,56 @@ def validate_field_set():
...
+        Field.objects.filter(template=self).exclude(pk__in=[field.get('id', -1) for field in field_set]).delete()
+
+        # Update current fields
+        # Should also delete the content in the entry if type changes
+        fields = Field.objects.filter(template=self)
+        for field in fields:
+            new_field_content = [f for f in field_set if f.get('id', -1) == field.pk]
```
engelhamer, line 279: I find this construction a bit vague. What are we looping over? Why does this work?
Please comment.

src/django/VLE/models/template.py
```diff
@@ -266,12 +266,56 @@ def validate_field_set():
...
+            if new_field_content:
+                new_field_content = new_field_content[0]
+            else:
+                # Field not found in content, ignoring
+                continue
+            if field.type != new_field_content['type']:
```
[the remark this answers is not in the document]
engelhamer: As of wysiwyg we do not allow type transitions anymore (🥳 )

src/django/VLE/models/template.py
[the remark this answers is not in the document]
engelhamer: This way we can omit the 'update existing entries' as well, and simply always use this approach for all entries using a template.

The only thing that is still impossible then, is changing a description for an existing field in an entry. But since the entry is already posted anyways this should not matter.

src/django/VLE/utils/template.py
```diff
@@ -143,27 +143,30 @@ def update_template_chain_based_settings(chain, data):
...
     update_template_chain_based_settings(current_template.chain, data)
 
-    if _should_be_archived(current_template, data):
-        updated_template = Template.objects.create_template_and_fields_from_data(
-            data=data,
-            format=current_template.format,
-            archived_template=current_template,
-        )
+    if not update_existing_entries:
+        if _should_be_archived(current_template, data):
```
engelhamer, line 154: Would like to make fields the archived thing, not template. In that case we can get rid of the TemplateChain as well I think.

src/vue/src/components/template/TemplateEdit.vue
engelhamer: If we always apply the changes to all entries (and archive fields iso delete if in use) this is not a problem anymore.

engelhamer (commented):
I see much better possibility for this feature in the context of the wysiwyg editor.

src/django/VLE/utils/template.py
Larspolo: This is/was meant as a quick solution to the problem. I do see reworks like this might even be better, but will also hold back the development of the rest. This gives teachers (finally) the ability to update existing entries without too much development

src/django/VLE/models/template.py
Larspolo: Added comment and restructured in the wysiwyg PR

src/django/VLE/models/template.py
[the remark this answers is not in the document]
Larspolo: Still sounds like an issue not for this PR

engelhamer:
Implemented through #1254

PR #1250, opened 2021-11-01
Improved loading speed of assignment API requests

Larspolo:
Including only the serializations that are needed.

Course overview page was about 5 to 10 seconds, which is now 0.5 to 1 second. Other pages also some improvements, but 
image

Also fixed an issue where badges would not show if it only has unpublished assessments


src/vue/src/store/modules/assignment.js
```diff
@@ -22,10 +23,10 @@ const mutations = {
 const actions = {
     retrieve (context, { id, cID = null, force = false, connArgs = auth.DEFAULT_CONN_ARGS }) {
         function fn () {
-            return auth.get(`assignments/${id}`, { course_id: cID }, connArgs)
-                .then((response) => {
-                    context.commit('SET_ASSIGNMENT', { assignment: response.data.assignment })
-                    return response.data.assignment
+            return assignmentAPI.get(id, cID, true, true, true, connArgs)
```
mjvkeulen, line 26: Should these arguments not be passable from the retrieve itself?

I would also argue that these arguments should default to false here as well.

The idea is that the vuex modules incrementally remove the need for api/* files, so I would prefer if we could do away with the import here.

mjvkeulen (changes requested):
Just wondering, did you also consider moving the deadlines, journals, and/or stats to their own API?

src/django/VLE/views/assignment.py
```diff
@@ -225,9 +225,22 @@ def list(self, request):
             course = settings.EXPLICITLY_WITHOUT_CONTEXT
             courses = request.user.participations.all()
 
+        serialize_stats, serialize_deadline = \
+            utils.optional_typed_params(
+                request.query_params,
+                (bool, 'serialize_stats', False),
+                (bool, 'serialize_deadline', False),
```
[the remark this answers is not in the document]
Larspolo: deadline is singular, only 1 deadline will be serialized. Is consistent with the `get_deadline` and `get_stats` functions

src/vue/src/store/modules/assignment.js
Larspolo: I imagined that the store would need to have consistent data always, else you will have different values inside the stored assignment, depending on the first time it was accessed

Larspolo: Will change it to an auth again. But waiting for a full review before changing it

mjvkeulen: Yes, this touches on the topic of splitting expensive conditional data to its own API. 

The assignment retrieve is called on **each** page which aID in the route.

Larspolo: Removed the API again

mjvkeulen: Its also cached, but this is complicated by the addition of thee conditional arguments in the API.

Larspolo: Thats why its always true, makes it non-conditional (functioning the same as it does rn). While all other requests can be improved that do not yet have the caching system

src/django/VLE/serializers.py
```diff
@@ -649,6 +650,8 @@ def permission_from_context(self, permission, assignment):
         return self.context['user'].has_permission(permission, assignment)
 
     def get_deadline(self, assignment):
+        if not self.context.get('serialize_deadline', False):
```
mjvkeulen, line 653: [D] Should we standardize something for this pattern?

Only include fields if a specific key (and/or value) is present?

For example a mixin which allows for:
```
conditional_fields = [
    'deadline',
    'journals',
    'stats',
]
```

Which are then only serialized provided the correct context is set?

src/django/VLE/views/assignment.py
```diff
@@ -346,14 +361,24 @@ def retrieve(self, request, pk=None):
...
             course = None
 
+        serialize_journals, serialize_stats, serialize_deadline = \
+            utils.optional_typed_params(
+                request.query_params,
+                (bool, 'serialize_journals', False),
+                (bool, 'serialize_stats', False),
+                (bool, 'serialize_deadline', False),
+            )
+        print(serialize_journals, serialize_stats, serialize_deadline)
```
mjvkeulen, line 371: Left over

src/django/VLE/serializers.py
Larspolo: Agree, and did it https://github.com/eJourn-al/eJournal/pull/1250/commits/466178e8e9f34bee3dd15d05d92713bd4af97fe8?w=1

engelhamer (approved):

mjvkeulen (approved):

PR #1251, opened 2021-11-03
Reduced the width of the login screen

Larspolo:
BEFORE:
image

AFTER:
image


src/vue/src/components/columns/SmallContent.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+    <b-row noGutters>
+        <b-col
+            cols="8"
+            offset="2"
+            lg="4"
+            offsetLg="4"
+            class="main-content"
```
mjvkeulen, line 8: So we created another relative column layout (not used elsewhere right?), this does not solve the issue entirely: a mega widescreen will still show a too wide login screen.

Should we switch to using a max-width instead? E.g. 1000px. 

Tagging @engelhamer to decide on this.

Larspolo: > So we created another relative column layout (not used elsewhere right?)

Yes. Not against

>  this does not solve the issue entirely: a mega widescreen will still show a too wide login screen.

It already caps

mjvkeulen: Sorry, missed the main content class.

Layout doesn't seem to play nice with all viewports.

Too small on medium size screens (1000-1350]:
image

Should be full width on small screens < 576
image

Larspolo: Added more steps

mjvkeulen: Seems to fix the complaints with those dimensions, just this device is a bit awkward:

image

Now its more a NIT

mjvkeulen (approved):

src/vue/src/components/columns/SmallContent.vue
```diff
@@ -0,0 +1,17 @@
+<template>
```
mjvkeulen, line 1: [NIT] Component name does not fit the column pattern, small-content-single-column perhaps?

Larspolo: `WideContent` is also single column

engelhamer (approved):

PR #1252, opened 2021-11-03
Added the ability to publish a saved comment

Larspolo:
image


mjvkeulen (changes requested):
1. Create a comment using Send & Publish After Grade
2. Edit comment
3. Options are now Send, and Send & Publish

This is confusing, which options should I use to just edit my comment without sending it directly (as was intended).

The use of `Send` in this scenario is just confusing. When editing a nonpublished comment I would expect to see `Save`.

I propose to rename `Send & Publish After Grade` to `Save & Publish After Grade` so we can use `Save` when editing a non-published comment.

Larspolo:
The comment is valid, and with it I discovered some [more stuff](https://ejournal.atlassian.net/browse/EJ-1163). Changed the name, also renames "Send & Publish" to just "Publish". Also changed the icon to a save icon

image

src/vue/src/components/entry/CommentCard.vue
```diff
@@ -210,7 +233,7 @@ export default {
             commentAPI.update(this.comment.id, {
                 text: this.comment.text,
                 files: this.comment.files.map((c) => c.id),
-                published: this.comment.published,
+                published: this.comment.published || this.publishOnUpdate === 'true',
```
mjvkeulen, line 236: Is this some limitation of the dropdown button? Why can't we work with `true` as a boolean instead of string?

src/vue/src/components/entry/CommentCard.vue
```diff
@@ -114,22 +114,44 @@
...
+                    :up="true"
+                    :selectedOption="publishOnUpdate"
+                    :options="{
+                        false: {
+                            text: 'Save',
+                            icon: 'save',
+                            class: 'green-button',
+                        },
+                        true: {
+                            text: 'publish',
```
mjvkeulen, line 143: [NIT] Capitalized anyway but..

mjvkeulen (approved):

src/vue/src/components/entry/CommentCard.vue
Larspolo: limitation of the dropdown button. It is based on the object key

src/vue/src/components/entry/CommentCard.vue
Larspolo: Automatically done

engelhamer (approved):

PR #1253, opened 2021-11-05
Added a display for the error messages received from grade passback

Larspolo:
image

image

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1253)


src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -40,6 +40,12 @@
                             >
                                 <slot class="float-right"/>
                                 {{ journal.name }}
+                                <icon
+                                    v-if="journal.LMS_message"
+                                    v-b-tooltip="`The grade could not be send to the LMS. ${journal.LMS_message}`"
+                                    name="exclamation-triangle"
+                                    class="ml-1 text-orange"
+                                />
```
mjvkeulen, line 48: I would move the config of the icon used for this purpose to a component

src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -40,6 +40,12 @@
                             >
                                 <slot class="float-right"/>
                                 {{ journal.name }}
+                                <icon
+                                    v-if="journal.LMS_message"
+                                    v-b-tooltip="`The grade could not be send to the LMS. ${journal.LMS_message}`"
```
mjvkeulen, line 45: When a teacher reads this, what action should they take?

This is currently missing.

src/django/VLE/utils/grading.py
```diff
@@ -208,14 +208,26 @@ def send_author_status_to_LMS(journal, author, left_journal=False, grade_only=Fa
                 author, grade, result_data=result_data, send_score=False, submitted_at=submitted_at)
             response_teacher = grade_request.send_post_request()
 
-    successful = (response_teacher is None or response_teacher['code_mayor'] == 'success') and \
-        (response_student is None or response_student['code_mayor'] == 'success')
     result = {'to_teacher': response_teacher, 'to_student': response_student}
 
+    successful = True
+    if response_teacher is not None and response_teacher['code_mayor'] != 'success':
```
mjvkeulen, line 214: We basically define what constitutes an unsuccessful LMS response in this check: move this to a function?

src/django/VLE/utils/grading.py
```diff
@@ -208,14 +208,26 @@ def send_author_status_to_LMS(journal, author, left_journal=False, grade_only=Fa
                 author, grade, result_data=result_data, send_score=False, submitted_at=submitted_at)
             response_teacher = grade_request.send_post_request()
 
-    successful = (response_teacher is None or response_teacher['code_mayor'] == 'success') and \
-        (response_student is None or response_student['code_mayor'] == 'success')
     result = {'to_teacher': response_teacher, 'to_student': response_student}
 
+    successful = True
```
mjvkeulen, line 213: A bit like defensive programming with permissions, should we not start with False (unsuccessful) and only flag successful to true if the response code exactly matched a successful response?

src/vue/src/components/assignment/JournalCard.vue
mjvkeulen: I also believe showing the actual error msg is not likely to be helpful? Or is this something they then google (technical already imo)?

mjvkeulen (changes requested):

src/vue/src/components/assignment/JournalCard.vue
Larspolo: It has different classes

src/vue/src/components/assignment/JournalCard.vue
Larspolo: > When a teacher reads this, what action should they take?

Passing this back to you, what action should they take? Any suggestions?

> I also believe showing the actual error msg is not likely to be helpful? Or is this something they then google (technical already imo)?

I think it is. The error message indicate if the user left the course, or if the assignment already finished

src/django/VLE/utils/grading.py
Larspolo: This is not for this PR, but could be. We do a lot of awful stuff. Maybe something to improve in the LTI1.3 PR

src/django/VLE/utils/grading.py
Larspolo: This is nicer to combine with the LMS_message response, else we would have to do this twice. NIT?

src/vue/src/components/assignment/JournalCard.vue
mjvkeulen: Just some positioning classes, works fine if this is set on a component instead.

src/vue/src/components/assignment/JournalCard.vue
mjvkeulen: Can we isolate what errors would be due to a more temporary condition (e.g. connectivity)?

If this is too much work, just provide generic advise:

- Try to publish the journal grades and see if the problem persists
- Are you sure the student is still part of the LMS course?
- Contact support if the problem persists and the steps above do not apply.

We can format this nicely, could even be a modal for more space.

src/django/VLE/utils/grading.py
mjvkeulen: Seems like a simple fix, may as well since we touch the line here?

src/django/VLE/utils/grading.py
mjvkeulen: How do you perceive a comment about an approach as a NIT?

src/vue/src/components/assignment/JournalCard.vue
Larspolo: > Can we isolate what errors would be due to a more temporary condition

Not sure on the full list of errors that are thrown

src/django/VLE/utils/grading.py
Larspolo: Because it works as is, and "This is nicer to combine with the LMS_message response, else we would have to do this twice".

src/vue/src/components/assignment/JournalCard.vue
Larspolo: image

src/django/VLE/utils/grading.py
Larspolo: You may add it if you want

src/vue/src/components/assignment/JournalCard.vue
```diff
@@ -40,6 +40,10 @@
                             >
                                 <slot class="float-right"/>
                                 {{ journal.name }}
+                                <l-m-s-message
```
engelhamer, line 43: Would prefer import as `LmsMessage` down below to avoid this weird kebab case style.

engelhamer: Also: maybe move this to next to the grade (above progress bar) instead of next to the student name? Since this error is about the grade not being passed back.

src/vue/src/components/journal/LMSMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+        <b-modal
+            :ref="`grade-passback-error-${journal.id}`"
+            size="lg"
+            noEnforceFocus
+            hide-footer
+        >
+            <template #modal-title>
+                Grade passback error
+            </template>
+            When sending the grade of this journal to the LMS, we received to following error:
```
engelhamer, line 22:
```suggestion
            When sending the grade of this journal to the LMS, we encountered the following error:
```

src/vue/src/components/journal/LMSMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+            <b-alert
+                :show="true"
+                class=" alert-dark"
+            >
+                {{ journal.LMS_message }}
+            </b-alert>
+            To solve this problem, try the following:
```
engelhamer, line 29:
```suggestion
            To resolve this issue, try the following:
```

src/vue/src/components/journal/LMSMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+            <ul>
+                <li>Check if the student is still in the course</li>
```
engelhamer, line 31: `;` at the end of every but the last line?

src/vue/src/components/journal/LMSMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+                <li>Check if the course is not concluded</li>
+                <li>Check if the enrollment of the student is not concluded</li>
```
engelhamer, line 33: The standard spelling of `enrolment` in British English only has one L.

src/vue/src/components/journal/LMSMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+                <li>Try to grade again and see if the problem persists (it may take a minute to update)</li>
+                <li><a
+                    href="mailto:support@ejournal.app"
+                    class="text-blue"
+                >Contact support</a> if the problem persists</li>
```
engelhamer, line 38:
```suggestion
                >Contact support</a> if the issue persists.</li>
```

src/vue/src/components/journal/LMSMessage.vue
Larspolo: Good thing you say. We use "color" in the category edit

image

src/vue/src/components/assignment/JournalCard.vue
Larspolo: I think next to the name is more prominent, which this should be. The badges that are about grading are also not on the progress bar

engelhamer: But why do you put it in the assessment card on the journal page then? I think this should be consistent.

Badges can also be moved, no problem for me.

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <span class="d-inline">
+        <span
+            @click.prevent.stop="showModal(`grade-passback-error-${journal.id}`)"
+        >
+            <icon
+                v-b-tooltip="`The grade could not be send to the LMS. Click to view more.`"
```
engelhamer, line 7:
```suggestion
                v-b-tooltip="`The grade could not be sent to the LMS. Click to learn more.`"
```

src/django/VLE/models/journal.py
```diff
@@ -392,6 +392,10 @@ class Journal(CreateUpdateModel):
     LMS_grade = models.FloatField(
         default=0,
     )
+    LMS_message = models.TextField(
+        null=True,
```
mjvkeulen, line 396: For text fields we are better of with null=False, blank=True, and default='' in situations where we do not use `null` differently than `''`.

src/vue/src/components/journal/JournalAssessmentCard.vue
```diff
@@ -1,6 +1,11 @@
 <template>
     <b-card noBody>
         <template #header>
+            <lms-message
```
mjvkeulen, line 4: Would you mind adding one of these to the preset DB?

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <span class="d-inline">
```
mjvkeulen, line 2: This component should only be visible to users that can grade

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+        <span
+            @click.prevent.stop="showModal(`grade-passback-error-${journal.id}`)"
+        >
+            <icon
+                v-b-tooltip="`The grade could not be sent to the LMS. Click to learn more.`"
+                name="exclamation-triangle"
+                class="ml-1 text-orange"
```
mjvkeulen, line 9: image

Add `shift-up-3` to fix the icon alignment issue.

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+            :ref="`grade-passback-error-${journal.id}`"
+            size="lg"
+            noEnforceFocus
+            hide-footer
+        >
+            <template #modal-title>
+                Grade passback error
+            </template>
+            When sending the grade of this journal to the LMS, we encountered the following error:
+            <b-alert
```
mjvkeulen, line 23: I would expect similar spacing before as after the alert:

image

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+                {{ journal.LMS_message }}
+            </b-alert>
+            To resolve this issue, try the following:
+            <ul>
+                <li>Check if the student is still in the course;</li>
+                <li>Check if the course is not concluded;</li>
+                <li>Check if the enrollment of the student is not concluded;</li>
+                <li>Try to grade again and see if the problem persists (it may take a minute to update);</li>
+                <li><a
+                    href="mailto:support@ejournal.app"
```
mjvkeulen, line 36: I would expect this to open our support modal, which should be doable with a simple root emit.

Nice to have would be some prefilled content and the help tab preselected.

mjvkeulen: [NIT] I would also formulate this as a button (`variant="link"`) at the bottom of the modal (so separate from the list of steps the user can do on their own, as this is really (hopefully) the last resort).

src/vue/src/components/journal/LmsMessage.vue
```diff
@@ -0,0 +1,61 @@
...
+            noEnforceFocus
+            hide-footer
+        >
+            <template #modal-title>
+                Grade passback error
+            </template>
+            When sending the grade of this journal to the LMS, we encountered the following error:
+            <b-alert
+                :show="true"
+                class=" alert-dark"
```
mjvkeulen, line 25: Should we phrase an error not at least as a warning? (So atleast background yellow)

mjvkeulen (changes requested):

src/vue/src/components/assignment/JournalCard.vue
Larspolo: Moved it to the journal details card then

src/django/VLE/models/journal.py
Larspolo: `None` is no message, `''` is more of a mistake that should not happen imo

src/vue/src/components/journal/JournalAssessmentCard.vue
Larspolo: Added in logboek and for user `student`

src/vue/src/components/journal/LmsMessage.vue
Larspolo: Should it? Maybe useful for the student to see as well, and I dont mind others seeing it too

src/vue/src/components/journal/LmsMessage.vue
Larspolo: Like this?
image

This is not the default bootstrap css then

src/vue/src/components/journal/LmsMessage.vue
Larspolo: See https://github.com/eJourn-al/eJournal/pull/1253?w=1#discussion_r752150731

src/vue/src/components/journal/LmsMessage.vue
Larspolo: image

src/vue/src/components/journal/JournalAssessmentCard.vue
Larspolo: https://github.com/eJourn-al/eJournal-commands/pull/12

src/vue/src/components/journal/LmsMessage.vue
Larspolo: > Nice to have would be some prefilled content and the help tab preselected.

Don't know how to do this nicely. Furthermore, not auto-filling content also increases the barrier to contact support ;)

src/vue/src/components/journal/LmsMessage.vue
mjvkeulen: Much of the advice does not apply to non-teaching staff right?

mjvkeulen (approved):
I still disagree with:

- field=models.TextField(default=None, null=**True**)

On the fence about showing the message to students as well

src/vue/src/components/assignment/JournalCard.vue
engelhamer: So then for consistency, also show it next to the grade (which this whole thing is about, not the name of the student) in the journal card?

src/vue/src/components/journal/LmsMessage.vue
engelhamer: I think teacher-only is fine. The errors originate from their actions (grading) after all.

engelhamer (approved):

PR #1255, opened 2021-11-06
Fixed a file upload field issue

Larspolo:
## Description
1. login as teacher
2. create new template
2. add field "file"
3. do not change anything about the upload restrictions
4. login as student
5. create entry with newly created template
6. upload a pdf in the file spot
7. see the following error:
image

Presumably since https://github.com/eJourn-al/eJournal/pull/1223/files?w=1

Fixed backend in the PR, as the * is nice to have as a valid field option


src/django/VLE/validators.py
```diff
@@ -124,7 +124,7 @@ def validate_entry_content(content, field): # NOQA C901
         if not os.path.isfile(fc.file.path):
             raise ValidationError('Entry references non existing file.')
 
-        if field.options:
+        if field.options and field.options != '*':
```
mjvkeulen, line 127: Magic variable which would be better in the field model

mjvkeulen (approved):

engelhamer (approved):

PR #1256, opened 2021-11-07
Improved instrumentation to detect long loading times

Larspolo:
image

Might be useful for the timeline bug. Also would also be pretty nice to fix any of these messages


src/vue/src/components/loading/LoadWrapper.vue
```diff
@@ -41,6 +41,14 @@ export default {
                 }, 500)
             }
         }, 1000)
+        setTimeout(() => {
+            if (this.loading) {
+                this.$store.commit('sentry/CAPTURE_SCOPED_MESSAGE', {
```
mjvkeulen, line 46: Do you have an example Sentry event @Larspolo, I would like to check if we need to enrich the context further.

Larspolo: Sentry is down atm, but the request was send

mjvkeulen: Hm weird, is up again

Larspolo: https://sentry.ejournal.app/organizations/ejournal/issues/877/?query=LoadWrapper&statsPeriod=14d

engelhamer (approved):

src/vue/src/components/loading/LoadWrapper.vue
mjvkeulen: We need relevant context, such as the value of `assignment`  and `storeAssignment`, to actually fix the issue?

mjvkeulen: We need a way to split the generated Sentry issues per page (or component)

Larspolo: https://sentry.ejournal.app/organizations/ejournal/issues/983/?project=2&query=&statsPeriod=14d

mjvkeulen (approved):

PR #1258, opened 2021-11-16
Improved the template copy and import experience

mjvkeulen:
EJ-1161 #done

template_copy

import_template

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1258)


src/vue/src/components/template/TemplateMenuItem.vue
```diff
@@ -25,10 +25,17 @@
...
         <icon
             name="trash"
-            class="trash-icon ml-2 float-right"
+            class="trash-icon ml-1 float-right"
             @click.native.stop="$emit('delete-template', template)"
         />
 
+        <icon
+            name="copy"
+            class="edit-icon ml-1 float-right"
```
Larspolo, line 34: [D] Not part of this PR, but edit icon blue / orange on hover? @engelhamer

src/vue/src/components/template/TemplateMenuItem.vue
```diff
@@ -85,5 +93,22 @@ export default {
             }
         },
     },
+    methods: {
+        ...mapActions({
+            templateCreate: 'template/create',
+        }),
+        copyTemplate () {
+            const copy = JSON.parse(JSON.stringify(this.template))
+            copy.name = `Copy of ${copy.name}`
```
Larspolo, line 102: [NIT] Would prefer a `${copy.name} (x)` system to copy a single template twice. Now you get "Please provide a unique template name."

src/vue/src/components/template/TemplateImport.vue
```diff
@@ -197,6 +210,13 @@ export default {
                     this.$toasted.error('Something went wrong while loading templates to import.')
                 })
         },
+        setNewTemplateName (template) {
+            let newName = template.name
+            while (this.assignmentTemplates.find((elem) => elem.name === newName)) { // eslint-disable-line no-loop-func
+                newName = `Copy of ${newName}`
```
Larspolo, line 216: Great change!

Copy of ${name} may cause confusion, as it is not necessarily a copy. Would suggest having the default name as `${template.name} from ${template.assignment.name}` or `Imported ${template.name}`

src/vue/src/components/template/TemplateMenuItem.vue
```diff
@@ -25,10 +25,17 @@
...
+            class="trash-icon ml-1 float-right"
             @click.native.stop="$emit('delete-template', template)"
         />
 
+        <icon
+            name="copy"
+            class="edit-icon ml-1 float-right"
+            :class="{ 'input-disabled': copyRequestInFlight }"
+            @click.native.stop="copyTemplate"
+        />
```
Larspolo, line 37: image
image

When the right side is too long, it breaks to another row. Check these changes to see how to fix it:

https://github.com/eJourn-al/eJournal/pull/1219/files?w=1#diff-8299bbbde931bf818955d8e2173cfb7a9873c77aa0bc03b432b8785254983763L247
(styling in src/vue/src/views/AssignmentEditor.vue)

Larspolo (changes requested):

src/django/VLE/models/template.py
```diff
@@ -104,8 +104,7 @@ def create_template_and_fields_from_data(
                 chain=archived_template.chain if archived_template else False,
             )
 
-            if not template_import:
```
Larspolo, line 107: [Q] How come this is changed and no tests are changed / added?

mjvkeulen: Validation of whether the categories are part of the assignment already takes place (and is tested). 

This was just an arbitrary (imo) constraint

src/vue/src/components/template/TemplateMenuItem.vue
mjvkeulen: Since it adds a template to assignment I would expect green if we want something particular

Larspolo: This I would suggest making use of the `add-icon` class

src/vue/src/components/template/TemplateMenuItem.vue
mjvkeulen: For 993 width (lowest val still `md`)

image

src/vue/src/components/template/TemplateImport.vue
mjvkeulen: It's true that the import is not an exact copy, as the categories are not copied over. However, I am not fond of the proposed names.

I went with `Imported` as to me it's the best out of the options.

Larspolo (approved):

src/vue/src/components/template/TemplateMenuItem.vue
engelhamer: I think green indeed.

src/vue/src/components/template/TemplateMenuItem.vue
```diff
@@ -85,5 +95,23 @@ export default {
             }
         },
     },
+    methods: {
+        ...mapActions({
+            templateCreate: 'template/create',
+        }),
+
+        copyTemplate () {
```
engelhamer, line 103: I would like the copied template to be selected automatically. If someone copies a template from within the same assignment they are 100% sure going to do _something_ with it, else the original template would have sufficed.

engelhamer (changes requested):

engelhamer (approved):

PR #1260, opened 2021-11-16
Fixed an issue where colors for to-do indicators were inconsistent

mjvkeulen:
EJ-1029 #done


Larspolo (approved):
See https://github.com/eJourn-al/eJournal/pull/1261 for another improvement for this PR

src/vue/src/components/assets/NumberBadge.vue
```diff
@@ -49,53 +49,53 @@ export default {
...
-                'background-purple',
-                'background-green',
-                'background-red',
-                'background-orange',
-            ],
-            unpublishedAssessmentTooltip: 'Overall assessment needs to be published',
+            types: {
+                needsMarking: {
+                    getTooltip: (n) => `${n} ${(n > 1) ? 'entries' : 'entry'} need${(n === 1) ? 's' : ''} marking`,
+                    class: 'background-blue',
```
engelhamer, line 55: Would like this to be a different color since now both needs marking and average points are blue in the statistics card, even though they are not the same.

engelhamer (approved):

src/vue/src/components/assets/NumberBadge.vue
mjvkeulen: Of the pair, I would favor changing the color of average points (the journal todo badges occur more often). 

Any color preference?

engelhamer: Pink or green

mjvkeulen: Went with green as pink is taken by another stat

PR #1265, opened 2021-11-24
Removed unused code

engelhamer:
## Description
Removed unused props, data and methods from all components.

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1265)

PR #1266, opened 2021-11-24
Merged student and teacher views for journal page and fixed various related issues

engelhamer:
## Description
This makes it easier to browse journals that one can view for an assignment (e.g. in a peer feedback setting) and removes the need to duplicate code and text.

## Summary of changes
- EJ-773 #done
- EJ-842 #done
- EJ-795 #done
- EJ-615 #done
- EJ-752 #done
- EJ-645 #done
- EJ-702 #done

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)


engelhamer:
First review: #1265

src/vue/src/components/entry/CommentCard.vue
```diff
@@ -231,23 +243,29 @@ export default {
...
                     this.saveRequestInFlight = false
                 })
         },
         resetComment () {
-            this.comment = JSON.parse(JSON.stringify(this.passedComment))
+            // Create a local copy for edits.
+            this.editedComment = JSON.parse(JSON.stringify(this.comment))
             this.editing = false
         },
+        safeToLeave () {
```
engelhamer, line 265: Prevent unssaved / not yet posted comments from getting lost.

Larspolo:
___
*[src/vue/src/components/assets/ImageCropper.vue, line 82 at r1](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MpMXmd252FuInWRg0Bv:-MpMXmd303GXtIj1FpM-:b-ahle5t) ([raw file](https://github.com/ejourn-al/ejournal/blob/66150542d39f885aab2c622d7e2148b6b4eb85ee/src/vue/src/components/assets/ImageCropper.vue#L82)):*
> ```Vue
>             this.$emit('newPicture', this.croppa.generateDataUrl('image/png'))
>         },
>         /* NOTE: Called from parent. */
> ```

Good to know!

Larspolo:
___
*[src/vue/src/components/entry/Entry.vue, line 129 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHbMgLDF-rFfZ0SB9E:-MrHbMgLDF-rFfZ0SB9F:bpo5tym) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L129)):*
> ```Vue
>                     pill
>                 >
>                     LOCKED
> ```

Preset entry node without an entry causes it to claim it is locked

Larspolo:
___
*[src/vue/src/components/entry/Entry.vue, line 1 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHbtRu3-vff0k46oTl:-MrHbtRu3-vff0k46oTm:bn6smvw) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L1)):*
> ```Vue
> <template>
> ```

Not sure where, but this issue is put in the console when opening an empty preset entry node

Larspolo (changes requested):
> Reviewed 9 of 44 files at r1, 21 of 35 files at r2, 16 of 16 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 13 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/components/entry/CommentCard.vue, line 265 at r1](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MpLkwxUGdmhza3h8L7h-r1-265:-MrHfKd8BQ8qtx3_gASm:b-jc1721) ([raw file](https://github.com/ejourn-al/ejournal/blob/66150542d39f885aab2c622d7e2148b6b4eb85ee/src/vue/src/components/entry/CommentCard.vue#L265)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Prevent unsaved / not yet posted comments from getting lost.
</blockquote></details>

I like it, with also the use of the store iso references
___
*[src/vue/src/components/entry/Entry.vue, line 129 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHbMgLDF-rFfZ0SB9E:-MrHduCY50HGF6-Eo82i:bld4o5y) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L129)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Preset entry node without an entry causes it to claim it is locked


</blockquote></details>

I fixed this issue in the NSR as well, maybe have a look there. I think I also fixed some other console issues
___
*[src/vue/src/components/entry/Entry.vue, line 182 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHhY_Z7gH55bwSL3D2:-MrHhY_Z7gH55bwSL3D3:b-7eudbl) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L182)):*
> ```Vue
>                     :edit="edit"
>                     :template="currentNode.entry.template"
>                     class="align-top mr-2 mt-2"
> ```

I cannot create entries in presetentrynodes atm, but once this is fixed, also check if you can add categories after making some initial changes. Because now changing the categories is locked till other changes are made (fine by me, but this was possible before)
___
*[src/vue/src/components/entry/Entry.vue, line 213 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHcpAC-MHya23yBIbG:-MrHcpAC-MHya23yBIbH:bvej9qe) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L213)):*
> ```Vue
>         </b-card>
> 
>         <comments
> ```

"Save & publish grade" causes a console error:

___
*[src/vue/src/components/entry/Entry.vue, line 404 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHe5HRALUxrGANSV9Y:-MrHe5HRALUxrGANSV9Z:b-wpkk8e) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L404)):*
> ```Vue
> 
>             // If the node is new, first create it, and then run this again
>             if (this.currentNode.entry.id === this.NEW_ENTRY_ID) {
> ```

Creating entry inside preset entry node causes error:


Requires reload before it works again
___
*[src/vue/src/components/entry/EntryGrade.vue, line 169 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHiEZn2myggvjn-0-b:-MrHiEZn2myggvjn-0-c:b-c56hlz) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/components/entry/EntryGrade.vue#L169)):*
> ```Vue
>         /* Checks if a pending grade exists */
>         safeToLeave () {
>             if (this.node && this.node.entry) {
> ```

Now you use node again, where in Entry.vue you use currentNode. When do we use what?
___
*[src/vue/src/views/Journal.vue, line 2 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHcEBeC85FIhxAaIGW:-MrHcEBfCa_w8lAyTyqg:b-hlxvwx) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/views/Journal.vue#L2)):*
> ```Vue
> <template>
>     <timeline-three-columns>
> ```

Whole journal reloads when giving a single grade, was this the case before as well? Seems more intrusive than it needs to be
___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrHjJckAGxMssQRFfg6:b695108) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
> ```Vue
> <template>
>     <timeline-three-columns>
> ```

Navigating to group journal as student doesnt work, here is the console error I get (works on develop):

___
*[src/vue/src/views/Journal.vue, line 159 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjh364IX6Br3K6Szr:-MrHjh364IX6Br3K6Szs:b-pq9u8f) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L159)):*
> ```Vue
>                         >
>                             <icon name="file-import"/>
>                             Import Journal
> ```

This was moved to the assignment footer. Think some merges changed this back. Would expect it in the footer again
___
*[src/vue/src/views/Journal.vue, line 300 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkYf47pZlmU2WbFx1:-MrHkYf47pZlmU2WbFx2:b7ejau8) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L300)):*
> ```Vue
>                 const nextIndex = (curIndex + 1) % this.filteredJournals.length
> 
>                 return this.filteredJournals[nextIndex]
> ```

TODO PFB: make next & prev journal skip over your own journal?
___
*[src/vue/src/views/Journal.vue, line 427 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkpsd1qsPpiF72fS6:-MrHkpsd1qsPpiF72fS7:b2ifidv) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L427)):*
> ```Vue
>             }
> 
>             // Regular findIndex returns -1
> ```

How is this return value used? In the example below I see -1 also false both as unhandles results
```
                const curIndex = this.findIndex(this.filteredJournals, 'id', this.jID)
                const nextIndex = (curIndex + 1) % this.filteredJournals.length

                return this.filteredJournals[nextIndex]
```

Larspolo:
___
*[src/vue/src/views/Journal.vue, line 159 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjh364IX6Br3K6Szr:-MrSp0YaAcVJljou2xpl:bublr6r) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L159)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

This was moved to the assignment footer. Think some merges changed this back. Would expect it in the footer again
</blockquote></details>

STill led to some issues, fixed now



_Code quote:_
```Vue
                            Import Journal
```

Larspolo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 3 of 230 files reviewed, 7 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 182 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHhY_Z7gH55bwSL3D2:-MrSly-v8I-BXjOZfTJz:beldi63) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L182)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

I cannot create entries in presetentrynodes atm, but once this is fixed, also check if you can add categories after making some initial changes. Because now changing the categories is locked till other changes are made (fine by me, but this was possible before)
</blockquote></details>

What I thought was indeed true, you cannot edit categories before submitting some content

But it does change seemlessly, looks good enough for me


___
*[src/vue/src/views/Journal.vue, line 2 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHcEBeC85FIhxAaIGW:-MrSmXlP5lQDnyWPPc3f:btnmku9) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Whole journal reloads when giving a single grade, was this the case before as well? Seems more intrusive than it needs to be
</blockquote></details>

Still an issue @engelhamer 
___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrSmWGHBnyr9XOBQgj_:bfhd29) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Navigating to group journal as student doesnt work, here is the console error I get (works on develop):


</blockquote></details>

Still a problem @engelhamer

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 3 of 230 files reviewed, 7 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 1 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHbtRu3-vff0k46oTl:-MrM3S_-1Q_4ugfa34Lh:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L1)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Not sure where, but this issue is put in the console when opening an empty preset entry node


</blockquote></details>

Done.
___
*[src/vue/src/components/entry/Entry.vue, line 129 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHbMgLDF-rFfZ0SB9E:-MrM3VMU5t4eOwE6SnS0:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L129)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

I fixed this issue in the NSR as well, maybe have a look there. I think I also fixed some other console issues
</blockquote></details>

Done.
___
*[src/vue/src/components/entry/Entry.vue, line 404 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHe5HRALUxrGANSV9Y:-MrM7oeJ8e60ZRw_yG5M:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L404)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Creating entry inside preset entry node causes error:


Requires reload before it works again
</blockquote></details>

Done.
___
*[src/vue/src/components/entry/EntryGrade.vue, line 169 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHiEZn2myggvjn-0-b:-MrM7qzt82DlVTRTvzzf:bzdkcyw) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/components/entry/EntryGrade.vue#L169)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Now you use node again, where in Entry.vue you use currentNode. When do we use what?
</blockquote></details>

Node when it is passed as a prop, currentNode when we use the store.
___
*[src/vue/src/views/Journal.vue, line 2 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHcEBeC85FIhxAaIGW:-MrIeb-i-F-PVQXXk1qk:bpqxy1r) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Whole journal reloads when giving a single grade, was this the case before as well? Seems more intrusive than it needs to be
</blockquote></details>

Yes this was already the case. Nodes do not reload, but journal does. I think this was the easy way to ensure the grade is always in sync. Made this more seamless in the vuex update.
___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrM82AHAuJsZytOsRUj:bnu4kf3) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Navigating to group journal as student doesnt work, here is the console error I get (works on develop):


</blockquote></details>

Fixed on vuex branch. How shall we merge them?
___
*[src/vue/src/components/assets/ImageCropper.vue, line 82 at r1](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MpMXmd252FuInWRg0Bv:-MpMXwtTC87-ygkmSoC7:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/66150542d39f885aab2c622d7e2148b6b4eb85ee/src/vue/src/components/assets/ImageCropper.vue#L82)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Good to know!
</blockquote></details>

Done.

engelhamer:
___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrTWGfJ1bzFbnvhECdI:bxjw12m) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Fixed on vuex branch. How shall we merge them?
</blockquote></details>

The comment above was not yet published. Whoops.

engelhamer:
___
*[src/vue/src/views/Journal.vue, line 159 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjh364IX6Br3K6Szr:-MrTWMHC-MyHd0J7YS0a:b-uvvi2p) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L159)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

STill led to some issues, fixed now
</blockquote></details>

I think this was also fixed in vuex merge.

engelhamer:
___
*[src/vue/src/views/Journal.vue, line 427 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkpsd1qsPpiF72fS6:-MrTWSqn2qxmA0BENSy0:bs51cqj) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L427)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

How is this return value used? In the example below I see -1 also false both as unhandles results
```
                const curIndex = this.findIndex(this.filteredJournals, 'id', this.jID)
                const nextIndex = (curIndex + 1) % this.filteredJournals.length

                return this.filteredJournals[nextIndex]
```
</blockquote></details>

Don't understand the second sentence.

Larspolo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 3 of 230 files reviewed, 5 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrWP-mZ0GPigJXQI2t7:b74iasp) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

The comment above was not yet published. Whoops.
</blockquote></details>

Make an issue and merge this? Or quicklly solve it here too, as it wasnt an issue before
___
*[src/vue/src/views/Journal.vue, line 427 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkpsd1qsPpiF72fS6:-MrWP9ia9u6VKZEwjPRV:b-hlwcwm) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L427)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Don't understand the second sentence.
</blockquote></details>

Why make use of `this.findIndex` iso `this.filteredJournals.findIndex`? The second returns -1 if it cannot be found.

engelhamer:
___
*[src/vue/src/components/entry/Entry.vue, line 182 at r3](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHhY_Z7gH55bwSL3D2:-MrXmsSF61Jbf9sQWR9J:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/afa02cb6722ffe4e2fda71fccbc9ed514b9a0c31/src/vue/src/components/entry/Entry.vue#L182)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

What I thought was indeed true, you cannot edit categories before submitting some content

But it does change seemlessly, looks good enough for me



</blockquote></details>

Done.

src/vue/src/views/Journal.vue
```diff
@@ -1,46 +1,413 @@
...
-        :aID="aID"
-        :jID="jID"
-    />
+    <timeline-three-columns>
+        <template #left>
+            <bread-crumb v-if="$root.lgMax">
+                <img
+                    v-if="journal && (assignment.is_group_assignment || !isJournalAuthor)"
+                    slot="left"
+                    :src="journal.image"
```
engelhamer, line 8: Test

Larspolo (approved):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 3 of 230 files reviewed, 2 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

mjvkeulen (changes requested):
> Reviewed 10 of 128 files at r5, 1 of 194 files at r6, 1 of 4 files at r7, 5 of 217 files at r8, 212 of 213 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 11 unresolved discussions (waiting on @engelhamer)

___
*[config, line 1 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWR4jtEGk_7txERMj2:-MrWR4jtEGk_7txERMj3:b-nyx4x4) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/config#L1)):*
> ```
> Subproject commit 14153fca7d280c17ee11cc02a3d5ebc51f494146
> ```

This PR does not impact config afaik?
___
*[src/vue/src/components/entry/CommentCard.vue, line 192 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWRUNM3vWcSCFmhZb7:-MrWRUNM3vWcSCFmhZb8:b-c8xvxl) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/CommentCard.vue#L192)):*
> ```Vue
>             saveRequestInFlight: false,
>             uploadingFiles: 0,
>             editedComment: {},
> ```

What does the name `editedComment` convey? Local / working copy of the comment?
___
*[src/vue/src/components/entry/CommentCard.vue, line 266 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWRjs-EAEPIJdY9Bc4:-MrWRjs-EAEPIJdY9Bc5:b-iwz99e) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/CommentCard.vue#L266)):*
> ```Vue
>         },
>         safeToLeave () {
>             return JSON.stringify(this.comment.text) === JSON.stringify(this.editedComment.text)
> ```

How come we need a stringify for this comparison?
___
*[src/vue/src/components/entry/Entry.vue, line 2 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWS3leCNcMCKF8aOfs:-MrWS3leCNcMCKF8aOft:b-rrdu3s) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L2)):*
> ```Vue
> <template>
>     <div v-if="currentNode.entry !== null">
> ```

What does the explicit null check accomplish?
___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MrWScTRBqIROyvn6PNO:b-34pm5u) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
> ```Vue
>     computed: {
>         ...mapGetters({
>             currentNode: 'timeline/currentNode',
> ```

Does an entry ever deal with something else than the `current` node?
___
*[src/vue/src/views/Journal.vue, line 325 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrX0D8W9pcDTWiKY6wk:-MrX0D8W9pcDTWiKY6wl:bsywipb) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/views/Journal.vue#L325)):*
> ```Vue
>         }),
>         loadJournal (gradeUpdated) {
>             const initialCalls = []
> ```

Remove the promise.all construct since only one call is made?
___
*[src/vue/src/views/Journal.vue, line 324 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYhn_S7Kx2bEr6oeCv:-MrYhn_S7Kx2bEr6oeCw:b-r4ecyt) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L324)):*
> ```Vue
>             setCurrentNode: 'timeline/SET_CURRENT_NODE',
>         }),
>         loadJournal (gradeUpdated) {
> ```

Saving (and or publishing) an entry grade reloads the page: page goes white all components are reloaded. To me this breaks the UX.
___
*[src/vue/src/views/Journal.vue, line 389 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYiBQC9CPG1D5wIQFS:-MrYiBQC9CPG1D5wIQFT:b-iwei2i) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L389)):*
> ```Vue
>                 })
>             }
>         },
> ```

The borders seem different on these badges?


___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MrYlVbx5jtc16AyQWy4:b-93km9s) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
> ```JavaScript
>     timelineInstance: (state) => state.timelineInstance,
>     currentNode: (state) => state.currentNode,
>     checkNodeNavigationGuards: (state) => () => state.nodeNavigationGuards.every((guard) => guard()),
> ```

This cannot be reactive, better to simply use a function in the module as to not set the wrong expectations (getters return state).  



_Code quote:_
```JavaScript
checkNodeNavigationGuards
```

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 229 of 230 files reviewed, 10 unresolved discussions (waiting on @Larspolo and @mjvkeulen)

___
*[src/vue/src/components/entry/CommentCard.vue, line 192 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWRUNM3vWcSCFmhZb7:-MraQspAEBLNR2r20hZP:b9ee78p) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/CommentCard.vue#L192)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

What does the name `editedComment` convey? Local / working copy of the comment?
</blockquote></details>

A comment that may have been edited compared to the original which is retrieved from the database.
___
*[src/vue/src/components/entry/CommentCard.vue, line 266 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWRjs-EAEPIJdY9Bc4:-MraV7eD5kGrW3dgx1nQ:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/CommentCard.vue#L266)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

How come we need a stringify for this comparison?
</blockquote></details>

Done.
___
*[src/vue/src/components/entry/Entry.vue, line 2 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWS3leCNcMCKF8aOfs:-MraV4pB-kTB1Yl5t2pT:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L2)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

What does the explicit null check accomplish?
</blockquote></details>

Done.
___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MraSPFVCthYRhuypJkv:b-3ntwny) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Does an entry ever deal with something else than the `current` node?
</blockquote></details>

No, but I like to have the distinction clear between a node passed as a prop (which is required in some other components) and thhe currently active node which is retrieved from the store.
___
*[src/vue/src/views/Journal.vue, line 2 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHjJckAGxMssQRFfg5:-MrXspAUBfOuAUoCnKzZ:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L2)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Make an issue and merge this? Or quicklly solve it here too, as it wasnt an issue before
</blockquote></details>

Done.
___
*[src/vue/src/views/Journal.vue, line 300 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkYf47pZlmU2WbFx1:-MrXnFw32fNcaZgKg0GF:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L300)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

TODO PFB: make next & prev journal skip over your own journal?
</blockquote></details>

Done.
___
*[src/vue/src/views/Journal.vue, line 427 at r4](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrHkpsd1qsPpiF72fS6:-MrXuodC9F6HKM2S6UP1:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/8b81de95286595580dfed68fe7ba87d4de365eb1/src/vue/src/views/Journal.vue#L427)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Why make use of `this.findIndex` iso `this.filteredJournals.findIndex`? The second returns -1 if it cannot be found.
</blockquote></details>

Done.
___
*[src/vue/src/views/Journal.vue, line 325 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrX0D8W9pcDTWiKY6wk:-MraSwcJ9oAnoCvTNvW6:b-5erjyz) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/views/Journal.vue#L325)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Remove the promise.all construct since only one call is made?
</blockquote></details>

An additional call may be added right below in the conditional statement.
___
*[src/vue/src/views/Journal.vue, line 324 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYhn_S7Kx2bEr6oeCv:-MraSaLWCPCZr85vE7XL:b-m7bljd) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L324)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Saving (and or publishing) an entry grade reloads the page: page goes white all components are reloaded. To me this breaks the UX.
</blockquote></details>

This change is not introduced in this pull request. However, the vuex rework makes this transition more seamless.
___
*[src/vue/src/views/Journal.vue, line 389 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYiBQC9CPG1D5wIQFS:-MraTEEw9-47EC107nM3:b-of15b6) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L389)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

The borders seem different on these badges?


</blockquote></details>

I don't know why this comment is attached to a closing curly bracket, but for me the assignment instructions look like this 

___
*[config, line 1 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWR4jtEGk_7txERMj2:-MraQopi97l_jK5dO5PT:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/config#L1)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

This PR does not impact config afaik?
</blockquote></details>

Done.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MraUWk69oQHMrQKOLoG:b-gejzlz) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

This cannot be reactive, better to simply use a function in the module as to not set the wrong expectations (getters return state).  
</blockquote></details>

What do you mean with it not being reactive? This getter returns a function that encompassess all checks that have been registered as navigation guards, and these are checked before route leave. It works fine afaik.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 227 of 230 files reviewed, 10 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MraWT-x9Es_twgjQig2:bj972r3) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

No, but I like to have the distinction clear between a node passed as a prop (which is required in some other components) and thhe currently active node which is retrieved from the store.
</blockquote></details>

In components where there is no distinction (no other nodes exist other the "one"), current puts me in the wrong mindset. I am immediately start looking for another node or a reason the node would change.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 227 of 230 files reviewed, 9 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/views/Journal.vue, line 324 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYhn_S7Kx2bEr6oeCv:-MraYAag5XIbC1W6MiYA:b-w06p0f) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L324)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

This change is not introduced in this pull request. However, the vuex rework makes this transition more seamless.
</blockquote></details>

The change is introduced here afaik:

Live:


Changed:


And change is recorded locally, the interruption will be much more significant when live and the load times are longer

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 227 of 230 files reviewed, 9 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/views/Journal.vue, line 389 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYiBQC9CPG1D5wIQFS:-MraZ0I-9pUhMafodpuz:b-24gllo) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L389)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

I don't know why this comment is attached to a closing curly bracket, but for me the assignment instructions look like this 


</blockquote></details>

How strange, something to keep an eye on.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MraZA7bEWp4FRoO50zZ:b-9m603g) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

What do you mean with it not being reactive? This getter returns a function that encompassess all checks that have been registered as navigation guards, and these are checked before route leave. It works fine afaik.
</blockquote></details>

Afaik you can't access this getter as state (which is what vuex getters are expected to be), e.g. this.checkNodeNavigationGuards would not work as this getter is not reactive. You have to call it as a function which is unexpected?

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 227 of 230 files reviewed, 5 unresolved discussions (waiting on @Larspolo and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Mra_OJR6Hbw9D9e8NRB:b-6cmmkc) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

In components where there is no distinction (no other nodes exist other the "one"), current puts me in the wrong mindset. I am immediately start looking for another node or a reason the node would change.
</blockquote></details>

Then the same getter from the store has different names depending on where it is used. I don't like that.
___
*[src/vue/src/views/Journal.vue, line 324 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYhn_S7Kx2bEr6oeCv:-Mraaa06-khw79o9q5bB:b-3kqaus) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/views/Journal.vue#L324)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

The change is introduced here afaik:

Live:


Changed:


And change is recorded locally, the interruption will be much more significant when live and the load times are longer

</blockquote></details>

Ah indeed the visual change was introduced here by accident. The journal and nodes were already fully reloaded though.

I fixed it again here, but still it is much better in vuex.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-Mra_yD0BdyLr6JUX8Xw:bxc9i7k) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Afaik you can't access this getter as state (which is what vuex getters are expected to be), e.g. this.checkNodeNavigationGuards would not work as this getter is not reactive. You have to call it as a function which is unexpected?
</blockquote></details>

The title has check in it which is an active form. What do you suggest then?

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 226 of 230 files reviewed, 5 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Mrabe99EdnFUt85xwYM:b-w1p98f) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Then the same getter from the store has different names depending on where it is used. I don't like that.
</blockquote></details>

I would prefer that the local state conveys how the variable is used. 

It's not like we named `assignment` in the assignment module `currentAssignment`, current is only relevant in the timeline module or another component where other nodes exist as well

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 226 of 230 files reviewed, 4 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MracNhA3qx2u-H4xdZw:biagjjl) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

The title has check in it which is an active form. What do you suggest then?
</blockquote></details>

I would just use a function, same stuff just defined outside of the getters object but still in the module.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 226 of 230 files reviewed, 4 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[config, line 1 at r12](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-Mracz4H9_GKgIHEo1DI:-Mracz4H9_GKgIHEo1DJ:bgmwu8e) ([raw file](https://github.com/ejourn-al/ejournal/blob/a6778c520d05b2c095301635801221868604e56b/config#L1)):*
> ```
> Subproject commit 5790b332d95d78d522aab812577ae9ec438e5723
> ```

Harder to tell from reviewable, but this still changes the submodule pointer?

Larspolo (approved):
> Reviewed 3 of 3 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 3 unresolved discussions (waiting on @engelhamer)

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 3 unresolved discussions (waiting on @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Mray9KrDO9fBnh9GGXB:b-i5axps) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I would prefer that the local state conveys how the variable is used. 

It's not like we named `assignment` in the assignment module `currentAssignment`, current is only relevant in the timeline module or another component where other nodes exist as well
</blockquote></details>

Yes, nad that is true: there can be other journals as well. The currently active journal is named as such. I like that this is explicit.
___
*[config, line 1 at r12](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-Mracz4H9_GKgIHEo1DI:-MrayQWK9Rxz1aVy7-0K:bhtftj7) ([raw file](https://github.com/ejourn-al/ejournal/blob/a6778c520d05b2c095301635801221868604e56b/config#L1)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Harder to tell from reviewable, but this still changes the submodule pointer?
</blockquote></details>

On Github I see no change. This is develop of the config repo.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MrayVj64ztA9Jr9i263:b-lfvl1r) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I would just use a function, same stuff just defined outside of the getters object but still in the module.
</blockquote></details>

So you mean an action?

mjvkeulen (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 2 unresolved discussions (waiting on @engelhamer)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MrbIP58AUfL5KV2vNeR:b-wkctq6) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Yes, nad that is true: there can be other journals as well. The currently active journal is named as such. I like that this is explicit.
</blockquote></details>

I don't understand, an Entry is always part of a single node right?
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MrbJCDoB4URqOEOgGyN:b-arffdh) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

So you mean an action?
</blockquote></details>

You could use actions, but those would be async (which would be unexpected here as well). I would just use a normal function, defined in the store module

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 2 unresolved discussions (waiting on @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MrcaU0f8QL4zcwXkGah:b-uu76iq) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I don't understand, an Entry is always part of a single node right?
</blockquote></details>

Yes, that is true. But I chose to use currentJournal and currentNode for the store values to signify that the values relate to the currently selected journal/node from there, and not passed as props.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-Mrcar0sBAHp7IPDZL7X:b7w2c1k) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

You could use actions, but those would be async (which would be unexpected here as well). I would just use a normal function, defined in the store module
</blockquote></details>

Don't see why that would be much better than the current situation?

Do you refer to a method that is neither a getter, mutation or action but does reside in the vuex store? Is that common? Do we (or others) do this elsewhere?

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 2 unresolved discussions (waiting on @engelhamer)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Mrfum547E8G3ne4HEzk:b-dkoaeo) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Yes, that is true. But I chose to use currentJournal and currentNode for the store values to signify that the values relate to the currently selected journal/node from there, and not passed as props.
</blockquote></details>

I am not sure I follow, we don't use `currentAssignment` to distinguish that the assignment comes from the store module and not directly from an API either?
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-Mrfv6db2kOi-224b6vS:bswyehv) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Don't see why that would be much better than the current situation?

Do you refer to a method that is neither a getter, mutation or action but does reside in the vuex store? Is that common? Do we (or others) do this elsewhere?
</blockquote></details>

Getters: yield reactive state of a store module.
Mutations: Mutate state synchronously
Actions: Async function

The pattern here does not fit any of these afaik, you could use an action (it's just that this call does not have to be async). 

We have some functions defined in store modules which are called elsewhere (e.g. propagateTemplateCategoryUpdate), these are just exported alongside the getters in the store modules

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 225 of 230 files reviewed, 2 unresolved discussions (waiting on @Larspolo and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MrhKFyoE-Rl275ukENZ:bg99b8j) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I am not sure I follow, we don't use `currentAssignment` to distinguish that the assignment comes from the store module and not directly from an API either?
</blockquote></details>

Don't think that is necessary, as there is never really a context where multiple assignments are used in the same action / sequence. Within one journal there's a button to navigate to the next. From the currently active node in the timeline you can navigate straight to another.

Viewing another assignment is not the same type of action.

Don't see value in this debate about a naming scheme.
___
*[src/vue/src/store/modules/timeline.js, line 21 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrYlVbx5jtc16AyQWy3:-MrlSaYa02iUnMp3U0fV:bwmcpup) ([raw file](https://github.com/ejourn-al/ejournal/blob/f465841af4ec8513478705c3d38fe04d7d2bd16f/src/vue/src/store/modules/timeline.js#L21)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Getters: yield reactive state of a store module.
Mutations: Mutate state synchronously
Actions: Async function

The pattern here does not fit any of these afaik, you could use an action (it's just that this call does not have to be async). 

We have some functions defined in store modules which are called elsewhere (e.g. propagateTemplateCategoryUpdate), these are just exported alongside the getters in the store modules


</blockquote></details>

Moved it to outside of the getters.

mjvkeulen (commented):
> Reviewed 2 of 2 files at r14, 1 of 3 files at r15, 4 of 4 files at r17, 2 of 2 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 233 of 235 files reviewed, 3 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/store/modules/timeline.js, line 74 at r18](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrqXMTcAt3FZbSwx8Jh:-MrqXMTdEvRHpPWsv86m:b-2ssgvl) ([raw file](https://github.com/ejourn-al/ejournal/blob/82829f2453b1a054216df33e123831f72a71ad3b/src/vue/src/store/modules/timeline.js#L74)):*
> ```JavaScript
> function checkNodeNavigationGuards () {
>     if (state.nodeNavigationGuards.every((guard) => guard())) {
>         state.nodeNavigationGuards = []
> ```

Checking the guards has the side effect of clearing the guards as well?

Also when store state is modified, we should adhere to using mutations to do so (if possible).

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 233 of 235 files reviewed, 3 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MrqYZmA24PTSUi7qQyy:b-ma8fnu) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Don't think that is necessary, as there is never really a context where multiple assignments are used in the same action / sequence. Within one journal there's a button to navigate to the next. From the currently active node in the timeline you can navigate straight to another.

Viewing another assignment is not the same type of action.

Don't see value in this debate about a naming scheme.
</blockquote></details>

I feel like we agree, similar to the assignment context, `Entry.vue` (and most other components) have no interaction with another node / journal. This is great, as a component encapsulates state and makes reasoning about logic easier. Therefore renaming `node` to `currentNode` in these scenario's just puts one on the wrong track (looking for another node).

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 233 of 235 files reviewed, 2 unresolved discussions (waiting on @Larspolo and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Ms-WMLI-cvVQ_gzMp6-:b17ngzx) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I feel like we agree, similar to the assignment context, `Entry.vue` (and most other components) have no interaction with another node / journal. This is great, as a component encapsulates state and makes reasoning about logic easier. Therefore renaming `node` to `currentNode` in these scenario's just puts one on the wrong track (looking for another node).
</blockquote></details>

Lars: "indifferent, leave as it is now to save hassle."
___
*[src/vue/src/store/modules/timeline.js, line 74 at r18](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrqXMTcAt3FZbSwx8Jh:-Ms1VsVU1SDntBL2tN0A:b-xhdth6) ([raw file](https://github.com/ejourn-al/ejournal/blob/82829f2453b1a054216df33e123831f72a71ad3b/src/vue/src/store/modules/timeline.js#L74)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Checking the guards has the side effect of clearing the guards as well?

Also when store state is modified, we should adhere to using mutations to do so (if possible).
</blockquote></details>

This was required because navigating to another node also checks the guards, and a re-evaluation was triggered immediately after confirming lost changes when going to the previous/next journal. The user would then see the same confirmation dialogue twice, but the last one had no impact on routing (the redirect had already started).

Indeed side effects are not preferable. Fixed this by only changing the current node upon journal load if there already is a current node.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: 233 of 235 files reviewed, 1 unresolved discussion (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-Ms4bS3F6Sbc3hi0KOhG:b-obntf5) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Lars: "indifferent, leave as it is now to save hassle."
</blockquote></details>

I disagree, especially since this is a change that was introduced here. Also, renames are not exactly time-consuming.

mjvkeulen (commented):
> Reviewed 2 of 2 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 1 unresolved discussion (waiting on @engelhamer)

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 1 unresolved discussion (waiting on @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MsEuHHn94igNsvpy2bo:b-8xfpap) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I disagree, especially since this is a change that was introduced here. Also, renames are not exactly time-consuming.
</blockquote></details>

'assignmentEditor/activeComponent'
'assignmentEditor/activeComponentOptions'
'assignmentEditor/selectedCategory'
'assignmentEditor/selectedTemplate'
'assignmentEditor/selectedRubric'
'assignmentEditor/selectedPresetNode'

All I have to say.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 1 unresolved discussion (waiting on @engelhamer)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MsKDbMC49k9d4GUgty9:b-sl5rkj) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

'assignmentEditor/activeComponent'
'assignmentEditor/activeComponentOptions'
'assignmentEditor/selectedCategory'
'assignmentEditor/selectedTemplate'
'assignmentEditor/selectedRubric'
'assignmentEditor/selectedPresetNode'

All I have to say.

</blockquote></details>

I don't understand what you are trying to convey? 

The example you give plays into my argument: assignmentEditor is a module that contains a lot of state and makes use of naming to distinguish between stuff like the selected template, a template draft or all the templates available to the assignment.

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: all files reviewed, 1 unresolved discussion (waiting on @mjvkeulen)

___
*[src/vue/src/components/entry/Entry.vue, line 301 at r9](https://reviewable.io/reviews/ejourn-al/ejournal/1266#-MrWScTRBqIROyvn6PNN:-MsorGqe599VBew9gErW:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/2d65182700590e79bccc42f8fd3105d65b59aa5e/src/vue/src/components/entry/Entry.vue#L301)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I don't understand what you are trying to convey? 

The example you give plays into my argument: assignmentEditor is a module that contains a lot of state and makes use of naming to distinguish between stuff like the selected template, a template draft or all the templates available to the assignment.
</blockquote></details>

Done.

mjvkeulen (approved):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1266)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @engelhamer)

PR #1271, opened 2021-11-26
Added the ability for students to preview the assessment model

mjvkeulen:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1271)


src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -1,44 +1,16 @@
...
-                        scale="0.8"
-                    />
-                    Sum of all points obtained through journal entries.
-                </span>
-            </b-td>
-            <b-td>
-                {{ entryPoints }}
-            </b-td>
-        </b-tr>
+        <rubric-automatic-criterion-display
```
engelhamer, line 3: I prefer to keep such a small (and infrequently used) element within this file. Having to navigate to so many layers that are at the same level in the file tree is annoying.

src/django/VLE/migrations/0103_rubric_visibility.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('VLE', '0102_journal_lms_message'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rubric',
+            name='visibility',
+            field=models.CharField(choices=[('v', 'Visible to students'), ('hu', 'Hidden from students until used in an assessment')], default='v', max_length=2),
```
engelhamer, line 16: Why `hu` and not just `h`?

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,73 @@
...
-        <template v-if="journal">
-            <load-wrapper
-                name="Journal assessment end-card"
-                :loading="loading"
-            >
+        <load-wrapper
+            :loading="loading"
+            name="Journal assessment end-card"
+        >
+            <template v-if="!loading">
```
engelhamer, line 14: Since `loading` is already passed to the `load-wrapper`, any content in this slot is conditional to this variable already. This layer can be removed?

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,73 @@
...
+                        </template>
+                    </rubric>
+                </template>
+            </template>
+        </load-wrapper>
 
-        <deadline-date-display :subject="assignment"/>
+        <deadline-date-display
+            :subject="assignment"
+            class="mr-2"
```
engelhamer, line 57:
```suggestion
            class="mr-2 mt-2"
```

The gap between the rubric and this line is too small.

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,73 @@
...
-                    v-if="assessment && !loading"
+                    v-if="journal && assessment"
                     :edit="false"
                     :assessment="assessment"
                     :journal="journal"
                 />
-            </load-wrapper>
-        </template>
+                <template v-else>
+                    <theme-select
```
engelhamer, line 22: Missing some indication which is the default rubric (or: that there is no default). Now students can click around but are unaware about what rubric is active.

src/vue/src/components/journal/JournalStartCard.vue
```diff
@@ -16,6 +18,33 @@
         >
             No description
         </i>
+
+        <hr/>
```
engelhamer, line 22: Can be omitted? We do not use `hr` elements for card footers anymore (see e.g. progress node).

src/django/VLE/migrations/0103_rubric_visibility.py
mjvkeulen: **H**idden **U**ntill, `h` can be used for just plain hidden at some point.

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Loading does not prevent its content from being added to the dom (unless changed recently)

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Default is preselected

src/vue/src/components/journal/JournalStartCard.vue
mjvkeulen: Hr creates a clean break between the description and additional actions

engelhamer: But it is not consistent with the other cards in the timeline, where there is no such distinction. Also in the panel on the right (progress bar and buttons) there is none.

src/vue/src/components/journal/JournalEndCard.vue
engelhamer: Yes but that is not stated anywhere. Also
> (or: that there is no default).

src/vue/src/components/journal/JournalEndCard.vue
engelhamer: image

How does that happen?

src/django/VLE/migrations/0103_rubric_visibility.py
engelhamer: Which would be hidden, even if it is used?

engelhamer: So that would be a secret rubric?

mjvkeulen: A rubric which is used to formulate a grade, but the rubric is never shared with the students, just the grade.

src/vue/src/components/journal/JournalStartCard.vue
mjvkeulen: None of these other examples have user-generated RT followed by actions

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Added (default) to the rubric select label

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: Glad this is the case, removed the extra template. 

Quite a few places have this duplicate guard in place, was this changed after the load wrapper's introduction?

src/vue/src/components/journal/JournalStartCard.vue
engelhamer: Progress goal has user-generated content with the deadline info underneath, no line
image

Then because there are actions here, the deadline info ends up underneath a line
image

Would prefer to just omit it
image

Or maybe in the footer?
image

Just because the line is not used much and can lead to inconsistency.

src/vue/src/components/journal/JournalEndCard.vue
engelhamer: Yes, this has been the case for 2 years 😄 

Created #1273 to fix these occurrences, there were not that many.

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -143,6 +143,19 @@
 
             <b-form-group
                 class="mt-2"
+                label="Visibility"
+            >
+                <theme-select
+                    v-model="rubric.visibility"
+                    :options="['v', 'hu']"
+                    :customLabel="rubricVisibilityLabel"
```
Larspolo, line 151: Feels like this should be a toggle setting, not a dropdown

src/vue/src/components/rubric/edit/RubricEdit.vue
```diff
@@ -143,6 +143,19 @@
...
+                    :multiple="false"
+                    :searchable="false"
+                    placeholder="Select An Assignment"
+                />
+            </b-form-group>
+
+            <b-form-group
```
Larspolo, line 158: When is this description used? Feels like sth we all overlooked and is never actually used

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,72 @@
...
+                :assessment="assessment"
+                :journal="journal"
+            />
+            <template v-else>
+                <theme-select
+                    v-if="rubrics.length > 1"
+                    v-model="selectedRubric"
+                    label="name"
+                    trackBy="id"
+                    :customLabel="(rubric) => rubric === defaultRubric ? `${rubric.name} (default)` : rubric.name"
```
Larspolo, line 26: Where can I see this? I dont see it in the assignment editor in de "assessment model" node it is not shown in the dropdown. Also for students I don't see it, but Im not sure why

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,72 @@
         <template slot="header">
             <slot name="edit-button"/>
             <h2 class="theme-h2">
-                End of assignment
+                {{ title }}
```
Larspolo, line 6: This not said "assessment model", but in the edit mode it says "assessment". Feels like it should be consistent between the two
image
image

src/django/VLE/views/rubric.py
```diff
@@ -32,18 +32,24 @@ def list(self, request):
...
+            )
+        else:
+            query = RubricSerializer.setup_eager_loading(
+                assignment.rubrics.filter(visibility=Rubric.Visibility.VISIBLE),
+            )
 
         serializer = RubricSerializer(
             query,
-            context={'user': request.user, 'annotated_is_in_use': 'True'},
+            context={'user': request.user, 'annotated_is_in_use': can_grade},
```
Larspolo, line 52: [NIT] Could be moved to a `conditional_field`

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -1,44 +1,16 @@
...
+            :label="'Entry points'"
+            :description="'Sum of all points obtained through journal entries.'"
```
Larspolo, line 5:
```suggestion
            label="Entry points"
            description="Sum of all points obtained through journal entries."
```

src/vue/src/components/assessment/AssessmentGrades.vue
```diff
@@ -1,44 +1,16 @@
...
-                    Bonus points given to this journal.
-                </span>
-            </b-td>
-            <b-td>
-                {{ bonusPoints }}
-            </b-td>
-        </b-tr>
+        <rubric-automatic-criterion-display
+            :label="'Bonus points'"
+            :description="'Bonus points given to this journal.'"
```
Larspolo, line 11:
```suggestion
            label="Bonus points"
            description="Bonus points given to this journal."
```

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,72 @@
...
 
-        <deadline-date-display :subject="assignment"/>
+                <rubric
+                    :rubric="selectedRubric ? selectedRubric : {criteria: []}"
+                    class="mt-2 mb-1"
+                >
+                    <template slot="tbody:post-criteria">
+                        <rubric-automatic-criterion-display
+                            :label="'Entry points'"
+                            :description="'Sum of all points obtained through journal entries.'"
```
Larspolo, line 40:
```suggestion
                            label="Entry points"
                            description="Sum of all points obtained through journal entries."
```

src/vue/src/components/journal/JournalEndCard.vue
```diff
@@ -3,32 +3,72 @@
...
+                            :value="journal ? journal.entry_points : '-'"
+                        />
+
+                        <rubric-automatic-criterion-display
+                            :label="'Bonus points'"
+                            :description="'Bonus points given to this journal.'"
```
Larspolo, line 46:
```suggestion
                            label="Bonus points"
                            description="Bonus points given to this journal."
```

src/vue/src/components/journal/JournalStartCard.vue
Larspolo: I dont mind any of the proposals

Larspolo (changes requested):

src/vue/src/components/rubric/edit/RubricEdit.vue
mjvkeulen: I don't mind either way, but expect this to become more than 2 options

src/vue/src/components/rubric/edit/RubricEdit.vue
mjvkeulen: Description is used within the assignment editor. So only for teachers managing the rubrics (atm)

src/vue/src/components/journal/JournalEndCard.vue
mjvkeulen: image

It should be shown in the assessment model node, provide a default assessment rubric is set?

src/vue/src/components/journal/JournalStartCard.vue
mjvkeulen: Moved to footer

src/vue/src/components/rubric/edit/RubricEdit.vue
Larspolo: Then its better, but as of now its only 2. When would you expect to add more? Changed it to a NIT

src/vue/src/components/rubric/edit/RubricEdit.vue
Larspolo: Not this PR, but feels like this needs a description then, I would expect it to be shown somewhere else as well

src/vue/src/components/journal/JournalEndCard.vue
Larspolo: I swear I did, thats why I was so confused of it not working. Checking it out now, it seems all fine

src/django/VLE/views/rubric.py
Larspolo, line 52: [Q] Previously then was a string? Think that was just wrong, right? Seems like 'True' was truthy, which made it work

Larspolo (approved):

engelhamer (approved):

PR #1273, opened 2021-11-28
Removed redundant conditionals for load wrappers

engelhamer:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1273)


src/vue/src/components/assignment/AssignmentDetails.vue
```diff
@@ -3,108 +3,132 @@
         name="Assignment details"
         :loading="assignmentDetails === null"
     >
-        <template v-if="assignmentDetails !== null">
```
engelhamer, line 6: I don't know why this diff is so large, but the only thing that happened here was the removal of the `template` wrapper and subsequent indentation change for its content.

mjvkeulen (approved):

Larspolo (approved):

PR #1279, opened 2021-12-03
Further improved the assignment serialization

Larspolo:
for courses without `can_view_journals_of_own_groups` specifically flagged. Its a bit worse off if it is, because it does extra queries to check that as well
image

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1279)


src/django/VLE/models/assignment_participation.py
```diff
@@ -44,13 +44,25 @@ def viewable_by(
...
+            VLE.models.Role.objects
+            .filter(
+                course__in=self.values('assignment__courses'),
+                pk__in=supervisor.participation_set.values('role')
+            )
+            .filter(
+                can_view_all_journals=False,
+                can_view_journals_of_own_groups=True,
+            ).exists()
+        )
```
mjvkeulen, line 57: Extract to func on Role queryset

engelhamer (approved):

src/django/VLE/models/assignment_participation.py
Larspolo: Did it

mjvkeulen (approved):

PR #1285, opened 2021-12-06
Fixed an issue where load times were reported incorrectly

mjvkeulen:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1285)


src/vue/src/components/loading/LoadWrapper.vue
```diff
@@ -28,6 +28,7 @@ export default {
     data () {
         return {
             keepSpinning: false,
+            componentDestroyed: false,
```
engelhamer, line 31: Maybe better to store the timeojt object?

Larspolo: and delete that object on destroy, I agree that that would be better. Still approving as this works too

Larspolo (approved):

engelhamer (approved):

PR #1288, opened 2021-12-07
Fixed an issue where TA's would be redirected to homepage

Larspolo:
be TA
have can_view_unpublished_assignment
dont have can_edit_assignment
navigate to unpublished assignment -> redirect to homepage

https://sentry.ejournal.app/organizations/ejournal/issues/44

NOTE: the toast can never occur if the assignments are serialized properly (which they are as of now)

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1288)


mjvkeulen (approved):

src/vue/src/main.js
```diff
@@ -175,11 +175,21 @@ new Vue({
...
-                        route.name = 'AssignmentEditor'
-                    }
-                } else { // Teacher published route
+                if (assignment.is_published) {
+                    // Published assignments, always go to assignment
+                    route.name = 'Assignment'
+                } else if (!this.$hasPermission('can_view_unpublished_assignment', 'assignment', assignment.id)) {
+                    // Unpublished assignments require the `can_view_unpublished_assignment` permission, else throw
+                    // error and redirect to homepage (redirect is done below)
+                    this.$toasted.error('You are not yet allowed to view this assignment.')
```
engelhamer, line 184:
```suggestion
                    this.$toasted.error('You are not allowed to view this assignment.')
```
Could be that this is never even going to be the case.

src/vue/src/main.js
```diff
@@ -175,11 +175,21 @@ new Vue({
...
+                } else if (assignment.is_group_assignment) {
+                    // Group assignments should go to assignment page, even when published
```
engelhamer, line 186: This block would never even execute right? It is an `else` to the `is_published` state..?

Larspolo: Unpublished group assignments? Those should be forwarded to the assignment page iso the editor

engelhamer: So then the comment should say unpublished?

engelhamer (approved):

src/vue/src/main.js
Larspolo: Updated the comment, also better explained the "else"

PR #1290, opened 2021-12-07
Expanded notification digest tests to other weekdays

Larspolo:
Coverage would still go down if the last PR was merged on a monday

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1290)


mjvkeulen (approved):

src/django/test/test_notification.py
```diff
@@ -331,6 +329,20 @@ def test_digest(self):
         assert len(mail.outbox) != before_len_mailbox, \
             'After verification, notification should be send'
 
+    @override_settings(CELERY_TASK_ALWAYS_EAGER=True, CELERY_TASK_EAGER_PROPAGATES=True)
+    def test_digest_monday(self):
+        with time_machine.travel(datetime.date(2021, 9, 6)):
```
engelhamer, line 334: Why specifically this date?

engelhamer (approved):

src/django/test/test_notification.py
Larspolo: Its a Monday. But any Monday would have worked

PR #1292, opened 2021-12-09
Added the ability to change the error message send to Sentry

Larspolo:
From error toasts

Also added it for entry missing required field
image


engelhamer (approved):

src/vue/src/api/auth.js
```diff
@@ -59,7 +60,9 @@ function toastError (error, connArgs) {
         /* The Django throttle module uses detail as description. */
```
mjvkeulen, line 60: `if (connArgs.customErrorToast) {` (few lines above). customErrorToast should also support sending a different message than we use for its Sentry signature?

src/vue/src/api/auth.js
```diff
@@ -29,6 +29,7 @@ const DEFAULT_CONN_ARGS = {
     customSuccessToast: false,
     responseSuccessToast: false,
     customErrorToast: false,
+    convertErrorMessageToSentry: (message) => message,
```
mjvkeulen, line 32: `formatSentrySignature`?

src/vue/src/helpers/toasted.js
```diff
@@ -15,7 +15,9 @@ function captureAndSend (msg, options, customCtx) {
 
     store.commit('sentry/CAPTURE_SCOPED_MESSAGE', {
         level: defaultCtx.level,
-        msg,
+        msg: options.convertErrorMessageToSentry
```
mjvkeulen, line 18: The `options` object should remain reserved for the Vue toasted API. Use or extend `customCtx` if you want to pass something to Sentry. See last line of `initToasted`.

mjvkeulen (changes requested):
I feel the approach is wrong here, this is a Sentry specific problem: how issues should be grouped. 

Sentry offers a lot of tooling for this out of the box as well: https://docs.sentry.io/product/data-management-settings/event-grouping/

To me this seems like a good fit for a [fingerprint rule](https://docs.sentry.io/product/data-management-settings/event-grouping/fingerprint-rules/): group the issues if the message fits a regex. This can be configured in our Sentry instance from the UI.

PR #1295, opened 2021-12-11
Fixed an issue where migration would result in only free use templates

Larspolo:
Templates should keep their old preset only status, not all be defaulted to "free use"

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1295)


src/django/VLE/migrations/0104_moved_preset_only_to_template_chain.py
```diff
@@ -3,20 +3,31 @@
 from django.db import migrations, models
 
 
+def fix_permission_default_for_existing_roles(apps, schema_editor):
```
engelhamer, line 6: For existing **roles**?

src/django/test/migrations/test_0104_moved_preset_only_to_template_chain.py
```diff
@@ -0,0 +1,48 @@
+from django_test_migrations.contrib.unittest_case import MigratorTestCase
+
+
+class TestMigration0065(MigratorTestCase):
```
engelhamer, line 4: Migration number does not match.

Larspolo: You caught me copy pasting :o fixed the names

mjvkeulen (approved):

engelhamer (approved):

PR #1297, opened 2021-12-20
Moved journals to Vuex store and use Vuex assignment more frequently

engelhamer:
## Summary of changes
- EJ-1177 #done
- Made it easier to loop through arbitrary collections of journals that are available for an assignment (for peer feedback)
- Made the transitions between journals and the assignment page smoother

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)


src/vue/src/store/modules/preferences.js
```diff
@@ -20,13 +20,6 @@ const getters = {
     // Search filters.
     todoSortBy: (state) => state.todo.sortBy,
     todoFilterOwnGroups: (state) => state.todo.filterOwnGroups,
-    journalSortAscending: (state) => state.journal.sortAscending,
-    journalFilterSubmitted: (state) => state.journal.filterSubmitted,
-    journalGroupFilter: (state) => state.journal.groupFilter,
-    journalSelfSetGroupFilter: (state) => state.journal.selfSetGroupFilter,
-    journalSearchValue: (state) => state.journal.searchValue,
-    journalSortBy: (state) => state.journal.sortBy,
-    journalAID: (state) => state.journal.aID,
```
mjvkeulen, line 29: Why do you prefer moving preferences to specific modules?

src/vue/src/store/modules/assignment.js
```diff
@@ -28,37 +28,28 @@ const mutations = {
...
-                    serialize_deadline: true,
-                }, connArgs)
-                .then((response) => {
-                    context.commit('SET_ASSIGNMENT', { assignment: response.data.assignment })
-                    return response.data.assignment
-                })
-        }
-
-        return context.dispatch('fromCache', { cache: 'retrieveCache', cacheKey: id, fn: fn.bind(null), force })
+    retrieve (context, { cID, aID, connArgs = auth.DEFAULT_CONN_ARGS }) {
```
mjvkeulen, line 31: What motivates the move away from the cache here?

src/vue/vue.config.js
```diff
@@ -87,6 +87,7 @@ module.exports = {
             })
             : [],
         ),
+        devtool: 'source-map',
```
mjvkeulen, line 90: What does this do?

src/vue/src/store/modules/assignment.js
engelhamer: Reactivity, as well as it seemed to me that the cache mixin maintained a redundant state.

src/vue/src/store/modules/preferences.js
engelhamer: These are not general preferences but search filters for journals. I moved them to the relevant module.

src/vue/vue.config.js
engelhamer: https://vuejs.org/v2/cookbook/debugging-in-vscode.html#Displaying-Source-Code-in-the-Browser

engelhamer: Did not work yet though, should still check again.

src/vue/src/store/modules/assignment.js
mjvkeulen: What issues did you encounter? The cache is reactive, provided you use the getters.

What state is redundant? I feel at worst we have a shared reference.

src/django/test/test_assignment.py
```diff
@@ -371,50 +367,15 @@ def test_get_assignment_with_given_course(self):
...
-        result = AssignmentSerializer(
-            assignment,
-            context={
-                'user': course2.author, 'course': course2,
-            },
-            include={
-                'journals': True
-            }
-        ).data
-        assert len(result['journals']) == 1, 'Course2 is supplied, only journals from that course should appear (1)'
```
mjvkeulen, line 383: Tagging: are these tests moved or no longer needed. For which courses are all journals serialized on the assignment page?

src/vue/src/Store.vue
```diff
@@ -1,68 +1,10 @@
 <script>
 export default {
-    debug: false,
     state: {
         cachedMap: {},
-        filteredJournals: [],
     },
     setCachedMap (cachedMap) {
         this.state.cachedMap = cachedMap
     },
```
mjvkeulen, line 8: How close are we to just removing this file?

Larspolo (changes requested):
> Reviewed 37 of 127 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 37 of 136 files reviewed, 9 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalEndCard.vue, line 84 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmIgcEeXhdRuOpdqk:-MrYmIgcEeXhdRuOpdql:b-jz8ia4) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalEndCard.vue#L84)):*
> ```Vue
>     },
>     props: {
>         journal: {
> ```

Could journal also be retrieved from the cache? `currentJournal: 'journals/currentJournal',`
___
*[src/vue/src/components/journal/JournalStartCard.vue, line 88 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmZnC6IFu1EU1__sE:-MrYmZnC6IFu1EU1__sF:b-enu0b6) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalStartCard.vue#L88)):*
> ```Vue
>             assessment: 'assessment/journalActiveAssessment',
>             endNode: 'timeline/endNode',
>             assignment: 'assignment/assignment',
> ```

Now its both a prob and a getter. How does that work? I think the prop can be removed right? Also the journal prop might be able to be moved to the cache getter
___
*[src/vue/src/components/journal/UserMissingLtiLinkWarning.vue, line 52 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmns52PoPEAON4G7q:-MrYmns52PoPEAON4G7r:bj0maz8) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/UserMissingLtiLinkWarning.vue#L52)):*
> ```Vue
>     computed: {
>         ...mapGetters({
>             currentJournal: 'journals/currentJournal',
> ```

I feel like the assignment prop can be moved to a getter. Maybe Im missing something though
___
*[src/vue/src/router/index.js, line 208 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYnlzT-UIMXFlvDNdC:-MrYnlzT-UIMXFlvDNdD:b-yt0nr7) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/router/index.js#L208)):*
> ```JavaScript
> 
>     if ('aID' in to.params) {
>         store.dispatch('assignment/retrieve', { cID: to.params.cID, aID: to.params.aID })
> ```

@mjvkeulen maybe knows it better, but I feel like the cache key would also need to contain the cID if you include (and use) this as well.
___
*[src/vue/src/store/constants/mutation-types.js, line 34 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYnHnfDARgW5jnjKPx:-MrYnHnfDARgW5jnjKPy:b-xuexqu) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/constants/mutation-types.js#L34)):*
> ```JavaScript
> export const SET_SEARCH_VALUE = 'SET_SEARCH_VALUE'
> export const SET_SORT_BY = 'SET_SORT_BY'
> export const SET_ACTIVE_ASSIGNMENT = 'SET_ACTIVE_ASSIGNMENT'
> ```

[NIT] Why are these sets seperated from the SET_COURSE_* and SET_JIR_*?
___
*[src/vue/src/store/modules/assignment.js, line 31 at r1](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcu-r1-31:-MrYnTPICEDyZcXsE9T1:bpc9bwc) ([raw file](https://github.com/ejourn-al/ejournal/blob/abd0e7be00b7a097d8aab6a332b63fec8322deaa/src/vue/src/store/modules/assignment.js#L31)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

What issues did you encounter? The cache is reactive, provided you use the getters.

What state is redundant? I feel at worst we have a shared reference.
</blockquote></details>

This change might have been connected to the fix in this PR: https://github.com/eJourn-al/eJournal/pull/1296/files
___
*[src/vue/src/store/modules/journals.js, line 149 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYoejmAGJdzlT9mGKn:-MrYoejmAGJdzlT9mGKo:bouerke) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L149)):*
> ```JavaScript
>         state.filters.sortBy = sortBy
>     },
>     [types.SET_ACTIVE_ASSIGNMENT] (state, aID) {
> ```

Feels like something that should not be in the journal module, but in the assignment module instead. And then you can access it from the other modules.



_Code quote:_
```JavaScript
SET_ACTIVE_ASSIGNMENT
```
___
*[src/vue/src/views/Journal.vue, line 100 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYweuA9DFJQaEuRkDc:-MrYweuA9DFJQaEuRkDd:bxd6ozt) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/views/Journal.vue#L100)):*
> ```Vue
>                         class="mb-3"
>                         @jir-processed="loadCurrentJournal(
>                             $store.getters['preferences/journalImportRequestButtonSetting'] in ['AIG', 'AEG', 'AWGZ'])"
> ```

[NIT] would like these values to be actual names, or even better to make this a function on its own: `shouldUpdateCurrentJournal()`

Also, this can lead to a minor issue:

1. click on accept with grades
2. before api call is done, switch to "reject jir" (or any option not in this list
3. when api call finishes, it will  now not reload the journal

Would be better to check the value that was send with the request to prevent this issue

src/vue/src/components/journal/JournalAssessmentCard.vue
```diff
@@ -151,18 +148,22 @@ export default {
     },
     computed: {
         ...mapGetters({
+            currentJournal: 'journals/currentJournal',
```
mjvkeulen, line 151: I would be fine with simply using `journal` for the local state, `currentJournal` makes one expect another journal within the component context. As far as I can tell other journals are only relevant within the journal module itself.

src/vue/src/components/entry/Entry.vue
```diff
@@ -226,11 +226,12 @@
...
         </template>
         <b-card-body
-            class="small background-light-orange"
+            v-if="currentNode.is_locked"
+            class="small background-light-red"
             show
         >
             <b-badge
-                class="background-orange mr-2"
+                class="background-red mr-2"
```
mjvkeulen, line 234: Change is fine by, but are there more of these little changes in this PR? (Was just not expecting these)

src/vue/src/store/constants/mutation-types.js
```diff
@@ -23,19 +23,22 @@ export const PUSH_REFRESH_SUBSCRIBER = 'PUSH_REFRESH_SUBSCRIBER'
...
+export const SET_MULTIPLE_JOURNALS = 'SET_MULTIPLE_JOURNALS'
+export const SET_SINGLE_JOURNAL = 'SET_SINGLE_JOURNAL'
+export const SET_SORT_ASCENDING = 'SET_SORT_ASCENDING'
+export const SET_FILTER_SUBMITTED = 'SET_FILTER_SUBMITTED'
+export const SET_GROUP_FILTER = 'SET_GROUP_FILTER'
+export const SET_MANUAL_GROUP_FILTER = 'SET_MANUAL_GROUP_FILTER'
+export const SET_SEARCH_VALUE = 'SET_SEARCH_VALUE'
+export const SET_SORT_BY = 'SET_SORT_BY'
+export const SET_ACTIVE_ASSIGNMENT = 'SET_ACTIVE_ASSIGNMENT'
+export const DELETE_JOURNAL = 'DELETE_JOURNAL'
```
mjvkeulen, line 35: So we adopted this [pattern](https://vuex.vuejs.org/guide/mutations.html#using-constants-for-mutation-types) when we started with vuex.
I have stopped using the pattern recently as it adds overhead and with our setup we don't get any benefits afaik. What do you think?

Larspolo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 37 of 136 files reviewed, 11 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalAssessmentCard.vue, line 151 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrZ0FGY474uRuITdmuL:-MrZ0FGY474uRuITdmuM:bgjz728) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalAssessmentCard.vue#L151)):*
> ```Vue
>     computed: {
>         ...mapGetters({
>             currentJournal: 'journals/currentJournal',
> ```



After clicking on "submit for assessment" the journal reload doesn't seem to work, same goes for adding bonus points
___
*[src/vue/src/views/Journal.vue, line 1 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrZ-gqn-EtXl8ToOcFl:-MrZ-gqn-EtXl8ToOcFm:b-jluqoc) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/views/Journal.vue#L1)):*
> ```Vue
> <template>
> ```


Many console errors when entering the page, works fine still, but would like to not see those anymore

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,201 @@
+import * as preferenceOptions from '../constants/preference-types.js'
+import * as types from '../constants/mutation-types.js'
+import Vue from 'vue'
+import auth from '@/api/auth.js'
+
+const getters = {
+    journals: (state) => state.journals[state.aID],
```
mjvkeulen, line 7: How do we handle course context? (`for_course` etc)

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,201 @@
...
+    currentJournal: (state, _, rootState) => {
+        if (state.aID in state.journals) {
+            return state.journals[state.aID].find(
+                (journal) => journal.id === Number.parseInt(rootState.route.params.jID, 10),
+            )
+        }
+        return undefined
```
mjvkeulen, line 14: We should settle on whether we return null or undefined in the store getters when an object is not present. I guess undefined makes the most sense, but let's make it consistent (e.g with the assignment module).

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,201 @@
...
+    },
+    [types.SET_SORT_BY] (state, sortBy) {
+        if (!preferenceOptions.JOURNAL_SORT_OPTIONS.has(sortBy)) {
+            throw new Error('Invalid journal sorting option.')
+        }
+        state.filters.sortBy = sortBy
+    },
+    [types.SET_ACTIVE_ASSIGNMENT] (state, aID) {
+        // aID might be a string or integer depening on how it is loaded
+        if (aID != state.aID) { // eslint-disable-line eqeqeq
```
mjvkeulen, line 151: Should we not explicitly cast for the comparison instead?

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,201 @@
...
+    },
+}
+
+const actions = {
+    list (context, { cID, aID, connArgs = auth.DEFAULT_CONN_ARGS }) {
+        auth.get('journals', { course_id: cID, assignment_id: aID }, connArgs)
+            .then((response) => {
+                context.commit('SET_MULTIPLE_JOURNALS', { aID, journals: response.data.journals })
+            })
+    },
```
mjvkeulen, line 169: Could benefit from the cache construction

src/vue/src/store/modules/preferences.js
```diff
@@ -159,13 +119,6 @@ const mutations = {
         state.saved = {}
         state.todo.sortBy = 'date'
         state.todo.filterOwnGroups = true
-        state.journal.aID = null
-        state.journal.sortAscending = true
-        state.journal.filterSubmitted = false
-        state.journal.groupFilter = null
-        state.journal.selfSetGroupFilter = false
-        state.journal.searchValue = ''
-        state.journal.sortBy = 'markingNeeded'
```
mjvkeulen, line 168: By moving these preferences they no longer persist between visits. See store/index.js `createPersistedState`.

mjvkeulen, line 168: When addressing this do not forget to add a reset function to the module journal module (or restore this module), and to call this in the user/logout action as well.

src/vue/src/views/Analytics.vue
```diff
@@ -186,12 +185,11 @@ export default {
 
         /* Journals */
         filteredJournals () {
-            return this.allJournals.filter((journal) => (
```
mjvkeulen, line 189: The `allJournals` name is on purpose (**all**EntryAnalyticsData, **all**Groups), there is quite a lot of state in this component, and the **all** prefix helps with distinguishing between the filtered and non-filtered sets.

src/vue/src/views/Analytics.vue
```diff
@@ -428,19 +426,18 @@ export default {
...
-        initialCalls.push(this.getAssignmentUsers({ studentsOnly: true, aID: this.aID }))
-        initialCalls.push(this.getAssignment({ id: this.aID })) // Cached and stored in the store
+        initialCalls.push(this.loadAllJournals({ cID: this.cID, aID: this.aID }))
+        initialCalls.push(this.getAssignmentUsers({ studentsOnly: true, cID: this.cID, aID: this.aID }))
 
         Promise.all(initialCalls).then((results) => {
             this.convertDateStrsToDateObjs(results[0])
             this.allEntryAnalyticsData = results[0]
             this.allGroups = results[1].sort((a, b) => b.name < a.name)
-            this.allJournals = results[2].journals
```
mjvkeulen, line 443: You still want the load spinner to run on this page until all journals all fetched

mjvkeulen (changes requested):

src/vue/src/store/modules/journals.js
Larspolo: I think we should remove course context

src/vue/src/store/modules/journals.js
Larspolo: This is old copy pasted code. Originally from the middle-click on journal issue. No clue anymore

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 37 of 136 files reviewed, 19 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/router/index.js, line 208 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYnlzT-UIMXFlvDNdC:-MraQBhb2zxWz3aTVZUM:b-piipvo) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/router/index.js#L208)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

@mjvkeulen maybe knows it better, but I feel like the cache key would also need to contain the cID if you include (and use) this as well.
</blockquote></details>

If parameters affect the result in the cache (likely) best to include them in the key

src/django/test/test_assignment.py
engelhamer: They are no longer included in the assignment get, but instead use the journals list api. This is tested in another file.

src/vue/src/Store.vue
engelhamer: https://ejournal.atlassian.net/jira/software/projects/EJ/boards/1?selectedIssue=EJ-610&text=breadcru
https://ejournal.atlassian.net/jira/software/projects/EJ/boards/1?selectedIssue=EJ-846&text=store

src/vue/src/components/journal/JournalAssessmentCard.vue
engelhamer: There also exists `nextJournal` and `prevJournal` on the journal page. For this reason it is called `currentJournal` on that page. It would be confusing to have different names for the same store object (i.e. call it `journal` here, even though it is the same object.

src/vue/src/components/entry/Entry.vue
engelhamer: This was just a simple fix to match the timeline color. Orange signifies a passed due date (entry still submittable) whereas red signifies a passed lock date (no submission possible). Now this is consistent.

src/vue/src/store/constants/mutation-types.js
engelhamer: So you would want to remove this file? Fine by me, just make an issue for it. I assumed we wanted this still.

src/vue/src/store/modules/journals.js
engelhamer: Indeed that's what I would vouch for too. Simply check what the user can see through any course, instead of the course they just happen to be visiting.

I added the `cID` where required in the API calls, but the lookup only checks for assignment id's to make things quicker. Worst case: the page loads with journals that belong to the other course (but same assignment) then update shortly after.

src/vue/src/store/modules/journals.js
engelhamer: Undefined yes.

src/vue/src/store/modules/journals.js
engelhamer: Imo not pereferable because the journal grades etc. could be outdated then.

engelhamer: It is now still 'cached' but updated in the background every time to ensure the values are accurate.

mjvkeulen (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 36 of 136 files reviewed, 19 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/store/modules/journals.js, line 169 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-169:-MrabDaxDXK4HxLdXDaP:b-7kyxl6) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L169)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

It is now still cached but updated in the background every time to ensure the values are accurate.
</blockquote></details>

I feel we did spend some time discussing we could get the best of both worlds?

src/vue/src/views/Analytics.vue
engelhamer: But the store is now also simply `journals` and `filteredJournals`. Looks fine to me.

src/vue/src/views/Analytics.vue
engelhamer: They are likely already there because the user navigates here via the assignment page. But I added in an additional check for the load spinner.

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 35 of 136 files reviewed, 19 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalAssessmentCard.vue, line 151 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrZ0FGY474uRuITdmuL:-MraeQ-r3gPeBWzh7a-I:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalAssessmentCard.vue#L151)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>



After clicking on "submit for assessment" the journal reload doesn't seem to work, same goes for adding bonus points
</blockquote></details>

Done.
___
*[src/vue/src/components/journal/JournalEndCard.vue, line 84 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmIgcEeXhdRuOpdqk:-MragbWT8hANd_Ide-1E:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalEndCard.vue#L84)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Could journal also be retrieved from the cache? `currentJournal: 'journals/currentJournal',`
</blockquote></details>

Done.
___
*[src/vue/src/components/journal/JournalStartCard.vue, line 88 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmZnC6IFu1EU1__sE:-Mragd6nAty8AYzY5qHK:b-sxmvbk) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/JournalStartCard.vue#L88)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Now its both a prob and a getter. How does that work? I think the prop can be removed right? Also the journal prop might be able to be moved to the cache getter
</blockquote></details>

Yes, this was a merge error.
___
*[src/vue/src/components/journal/UserMissingLtiLinkWarning.vue, line 52 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYmns52PoPEAON4G7q:-MragvBS2xh9UXXsJmer:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/components/journal/UserMissingLtiLinkWarning.vue#L52)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

I feel like the assignment prop can be moved to a getter. Maybe Im missing something though
</blockquote></details>

Done.
___
*[src/vue/src/router/index.js, line 208 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYnlzT-UIMXFlvDNdC:-Mrah0V_56OTaQE2q3jt:b-8onvk5) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/router/index.js#L208)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

If parameters affect the result in the cache (likely) best to include them in the key
</blockquote></details>

It is required for the API, but not necessary / preferable as we already discussed on github. We want to move away from the course context.
___
*[src/vue/src/store/constants/mutation-types.js, line 34 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYnHnfDARgW5jnjKPx:-Mrah8uTBGqtmO0gjCtV:bjmgr9i) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/constants/mutation-types.js#L34)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

[NIT] Why are these sets seperated from the SET_COURSE_* and SET_JIR_*?
</blockquote></details>

Because they are all for the journal module and the others are preferences.
___
*[src/vue/src/store/modules/assignment.js, line 31 at r1](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcu-r1-31:-Mrahoub8DfTGhE7fre4:b-6nz8hr) ([raw file](https://github.com/ejourn-al/ejournal/blob/abd0e7be00b7a097d8aab6a332b63fec8322deaa/src/vue/src/store/modules/assignment.js#L31)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

This change might have been connected to the fix in this PR: https://github.com/eJourn-al/eJournal/pull/1296/files
</blockquote></details>

Yes. And also because I wanted data to always be up to date.
___
*[src/vue/src/store/modules/journals.js, line 149 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYoejmAGJdzlT9mGKn:-MrajBNS7WiCuCTp2Qck:b-ckfa8i) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L149)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Feels like something that should not be in the journal module, but in the assignment module instead. And then you can access it from the other modules.
</blockquote></details>

This is only for the applicable filters though.
___
*[src/vue/src/store/modules/journals.js, line 151 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-151:-Mrai0_GA73Ntk5pz91w:b6uoyry) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L151)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

This is old copy pasted code. Originally from the middle-click on journal issue. No clue why I did it like this anymore
</blockquote></details>

Copy paste indeed. See no reason to change it specifically.
___
*[src/vue/src/store/modules/journals.js, line 169 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-169:-Mrai6Fn1Io90Xs-E4ji:b4rp6o9) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L169)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I feel we did spend some time discussing we could get the best of both worlds?
</blockquote></details>

So you want me to re-introduce the cache here and always use force? That seems like a waste of time to me.
___
*[src/vue/src/views/Journal.vue, line 1 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrZ-gqn-EtXl8ToOcFl:-MraiDcUB5vHE3lwCz7T:b-fmd05a) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/views/Journal.vue#L1)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>


Many console errors when entering the page, works fine still, but would like to not see those anymore
</blockquote></details>

Merge mistakes fixed now.
___
*[src/vue/src/views/Journal.vue, line 100 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrYweuA9DFJQaEuRkDc:-Mraj9Wk45lZSzSYmRfC:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/views/Journal.vue#L100)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

[NIT] would like these values to be actual names, or even better to make this a function on its own: `shouldUpdateCurrentJournal()`

Also, this can lead to a minor issue:

1. click on accept with grades
2. before api call is done, switch to "reject jir" (or any option not in this list
3. when api call finishes, it will  now not reload the journal

Would be better to check the value that was send with the request to prevent this issue
</blockquote></details>

Done.

src/vue/src/store/modules/preferences.js
engelhamer: Added journals to the persisted state as well.

This does mean also that the filters sync when refreshing different tabs. This may not be completely expected behaviour, but was already the case I guess?

Larspolo (changes requested):
> Reviewed 9 of 46 files at r3, 88 of 127 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 3 of 3 files at r9, 12 of 12 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 12 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalDetails.vue, line 1 at r11](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Mrb7pfM0NDPDxJxK578:-Mrb7pfM0NDPDxJxK579:b-k2a1pq) ([raw file](https://github.com/ejourn-al/ejournal/blob/133d419542e31154e531e1efafad328f83102340/src/vue/src/components/journal/JournalDetails.vue#L1)):*
> ```Vue
> <template>
> ```

Group journal still have some unwanted behaviour:
- console errors when visiting journal
- grades / todo's being removed when going to the overview

___
*[src/vue/src/views/Journal.vue, line 1 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Mrb6aRy1oVo3AWYZx-Q:-Mrb6aRy1oVo3AWYZx-R:b51e6yg) ([raw file](https://github.com/ejourn-al/ejournal/blob/2b0b7e52b4b5738561bd51dd2cd40dd7bf32340e/src/vue/src/views/Journal.vue#L1)):*
> ```Vue
> <template>
> ```

Dont know where exactly, but I managed to get a "this.assignment is undefined" error by going to the assignment, ofcourse reproducing is difficult, probably has to do with async getting the assignment

Larspolo (changes requested):
> Reviewed 4 of 4 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 12 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalDetails.vue, line 1 at r11](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Mrb7pfM0NDPDxJxK578:-Mrd7_0Q2DjmEcZiRtXE:b-kf1lei) ([raw file](https://github.com/ejourn-al/ejournal/blob/133d419542e31154e531e1efafad328f83102340/src/vue/src/components/journal/JournalDetails.vue#L1)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Group journal still have some unwanted behaviour:
- console errors when visiting journal
- grades / todo's being removed when going to the overview


</blockquote></details>

This issue is now fixed, but the statistics are still off:



1. Average points seems to be a sum iso average. 
2. needs grading is not appearing, while there is an entry that needs grading
3. it doesnt show any boxes, simply emptyness

Larspolo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 12 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

src/vue/src/components/journal/JournalDetails.vue
```diff
@@ -76,23 +77,21 @@ import EditJournalModal from '@/components/journal/EditJournalModal.vue'
...
-            required: true,
-        },
-        journal: {
-            required: true,
-        },
-    },
     computed: {
+        ...mapGetters({
+            assignment: 'assignment/assignment',
+            currentJournal: 'journals/currentJournal',
```
mjvkeulen, line 91: Tagging this pending the discussion on the other PR 

- `current` prefix in a component that handles no other instances of the object (here journal)

src/vue/src/store/constants/mutation-types.js
mjvkeulen: Did you have a preference?

src/vue/src/store/modules/assignment.js
mjvkeulen: Why was this resolved?

src/vue/src/store/modules/journals.js
mjvkeulen: I am in favor as well!

src/vue/src/store/modules/preferences.js
mjvkeulen: That would already be the case indeed

src/vue/src/views/Analytics.vue
```diff
@@ -148,12 +146,16 @@ export default {
...
+            journals: 'journals/journals',
             assignment: 'assignment/assignment',
             categories: 'category/assignmentCategories',
             entryDeadlines: 'presetNode/assignmentPresetNodes',
             templates: 'template/assignmentTemplates',
             currentNode: 'timeline/currentNode',
         }),
+        loading () {
+            return this.journals && !this.loadingInitialCalls
+        },
```
mjvkeulen, line 158: This is an example of a construct which is not needed if we use the frontend cache

src/vue/src/views/Assignment.vue
```diff
@@ -12,10 +12,15 @@
             journal for this assignment.
         </b-alert>
 
-        <div slot="main-content-column">
+        <load-wrapper
+            slot="main-content-column"
+            name="Journal list in assignment overview"
+            :loading="!(journals && assignment)"
+        >
+            <!-- Somehow the load wrapper does not block the evaluation of these conditional loads -->
```
mjvkeulen, line 20: Haha this is something I brought up, which then followed removing all kinds (supposedly) unnecessary wrappers. I wonder what is causing this

src/vue/src/store/constants/mutation-types.js
engelhamer: Would indeed be faster to add new things. I'm fine with removal :)

src/vue/src/views/Assignment.vue
engelhamer: Yeah, weird. I tested the removal of those wrappers and it seemed fine.

src/vue/src/views/Analytics.vue
engelhamer: Renamed it as an early Christmas present!

src/vue/src/views/Assignment.vue
engelhamer: Yeah. Weird. Somehow the content is evaluated even though there is a `v-if` surrounding the slot.

src/vue/src/store/modules/assignment.js
engelhamer: Put the cache back.

src/vue/src/store/modules/journals.js
engelhamer: There is only one journals call active at a time, and it is only fired on the journal or assignment page view root. No reason to have the cache mixin.

src/vue/src/views/Analytics.vue
engelhamer: Used the store loading promise here too

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 126 of 144 files reviewed, 15 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/journal/JournalDetails.vue, line 1 at r11](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Mrb7pfM0NDPDxJxK578:-MrhpKln7BLnFaRQnAjF:byy24oq) ([raw file](https://github.com/ejourn-al/ejournal/blob/133d419542e31154e531e1efafad328f83102340/src/vue/src/components/journal/JournalDetails.vue#L1)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

This issue is now fixed, but the statistics are still off:



1. Average points seems to be a sum iso average. 
2. needs grading is not appearing, while there is an entry that needs grading
3. it doesnt show any boxes, simply emptyness
</blockquote></details>

Done. Also added tests for this.
___
*[src/vue/src/components/journal/JournalDetails.vue, line 91 at r12](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcl-r12-91:-Mrkhr2oDHrAQaFfLNG6:bqy5ohw) ([raw file](https://github.com/ejourn-al/ejournal/blob/ac62390f1606d61fd720e58d3623782a0e8e5e7d/src/vue/src/components/journal/JournalDetails.vue#L91)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Tagging this pending the discussion on the other PR 

- `current` prefix in a component that handles no other instances of the object (here journal)
</blockquote></details>

You know my opinion!
___
*[src/vue/src/store/modules/journals.js, line 7 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-7:-MrhK4Ev6C-HcwsDrdIZ:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L7)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

I am in favor as well!
</blockquote></details>

Done.
___
*[src/vue/src/views/Analytics.vue, line 158 at r12](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcz-r12-158:-MrhRqNS5p1TIFQK0bOt:b2q9lpq) ([raw file](https://github.com/ejourn-al/ejournal/blob/ac62390f1606d61fd720e58d3623782a0e8e5e7d/src/vue/src/views/Analytics.vue#L158)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

This is an example of a construct which is not needed if we use the frontend cache
</blockquote></details>

Only because this entire page is loaded at once. I prefer showing parts that are already loaded sooner, but didn't change it on this page. 
___
*[src/vue/src/views/Journal.vue, line 1 at r10](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Mrb6aRy1oVo3AWYZx-Q:-Mrkhwn08bCTCnbrPnSu:bt2vr1b) ([raw file](https://github.com/ejourn-al/ejournal/blob/2b0b7e52b4b5738561bd51dd2cd40dd7bf32340e/src/vue/src/views/Journal.vue#L1)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Dont know where exactly, but I managed to get a "this.assignment is undefined" error by going to the assignment, ofcourse reproducing is difficult, probably has to do with async getting the assignment



</blockquote></details>

Fixed.

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,209 @@
...
+        console.log(state.filters.searchValue)
+    },
+    [types.SET_SORT_BY] (state, sortBy) {
+        if (!preferenceOptions.JOURNAL_SORT_OPTIONS.has(sortBy)) {
+            throw new Error('Invalid journal sorting option.')
+        }
+        Vue.set(state.filters, 'sortBy', sortBy)
+    },
+    [types.SET_ACTIVE_ASSIGNMENT] (state, aID) {
+        // aID might be a string or integer depening on how it is loaded
```
mjvkeulen, line 158:
```suggestion
        // aID might be a string or integer depending on how it is loaded
```

src/vue/src/store/modules/journals.js
mjvkeulen: For example, the call between `Assignment` and `Analytics` could generally be shared? I feel if the pattern is worth adhering to for the other modules it is worth it here as well. 

Worst case implementation wise it means adding one `force=true` to `Journal.vue`.

src/vue/src/views/Journal.vue
```diff
@@ -287,67 +270,59 @@ export default {
...
+            .then(() => {
+                // Requires assignment to be loaded
+                this.loadCurrentJournal(false)
             })
-                .then((assignment) => {
-                    this.assignment = assignment
 
-                    this.loadJournal(false)
-                })
+        // TODO PFB: Retrieve journals for review based on query param
```
mjvkeulen, line 283: Tagging

src/vue/src/views/Profile.vue
```diff
@@ -81,7 +81,6 @@ export default {
     methods: {
         logOut () {
             this.$store.dispatch('user/logout')
-            this.$router.push({ name: 'Login' })
```
mjvkeulen, line 84: Did the interaction trigger sentry errors? Reloading the page unauthed?

requirements/base.txt
```diff
@@ -27,3 +27,5 @@ sentry-sdk==0.12.1 # Logging
 django-anymail==8.2 # Email client
 chardet==4.0.0 # Encoding estimation for spreadsheets
 setuptools==56.2.0
+
+StrEnum==0.4.6 # Utils
```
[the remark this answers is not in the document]
engelhamer: Hm must be a merge error, I did not manually add this either.

src/vue/src/views/Profile.vue
engelhamer: This redirect is no longer required as it automatically happens upon reloading the window after clearing the local storage.

src/vue/src/views/Journal.vue
engelhamer: Peer feedback will not be introduced in this PR, so this TODO remains for the next PR.

src/vue/src/store/modules/journals.js
engelhamer: Can be resolved imo

src/vue/src/store/modules/journals.js
```diff
@@ -0,0 +1,209 @@
...
+            state.filters.groupFilter = []
+            state.filters.manualGroupFilter = false
+            state.filters.searchValue = ''
+            state.filters.sortBy = 'markingNeeded'
+        }
+    },
+}
+
+const actions = {
+    list (context, { cID, aID, connArgs = auth.DEFAULT_CONN_ARGS }) {
```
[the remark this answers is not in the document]
engelhamer: There are still some API functions left that are used elsewhere (e.g., join or leave a journal). But the unused list, get and delete functions are now removed.

src/vue/src/store/modules/journals.js
engelhamer: It is already shared, the analytics page can immediately use the same journals as on the assignment page. I do not like the pattern but already added it back in to the assignment module. Would appreciate if we stop this back and forth discussion with the same arguments and leave this module as is.

src/vue/src/components/assignment/AnalyticsMenu.vue
```diff
@@ -27,7 +27,10 @@
                 keyPrefix="average-points-statistics"
             />
         </div>
-        <div class="mt-1">
+        <div
+            v-if="stats.needsMarking"
```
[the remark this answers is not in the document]
engelhamer: It was inconsistent: we did show that 0 entry grades required publishing, but did **not** show that 0 overall assessments required publishing. In my opinion both are not required, so I made it consistent.

src/vue/src/views/Assignment.vue
engelhamer: Fine to make an issue, probably a good idea to restructure the load wrapper such that the load conditional should be specified in the slot:

```
<load-wrapper>
    <template v-if="!loading">
        ...
    </template>
</load-wrapper>
```

Then show the spinner in absence of slot content.

engelhamer (commented):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: 114 of 145 files reviewed, 22 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/entry/EntryGrade.vue, line 229 at r15](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrY-TM3DNjvmcJ-Q7JA-r15-229:-Ms-XJX01YZzfCyrqS0Z:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/44576c98032e9af730eb2dee3686beb1beea1cc6/src/vue/src/components/entry/EntryGrade.vue#L229)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Tagging
</blockquote></details>

Done.
___
*[src/vue/src/store/modules/journals.js, line 158 at r15](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r15-158:-Ms-XMm1DhUMWewBHTIC:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/44576c98032e9af730eb2dee3686beb1beea1cc6/src/vue/src/store/modules/journals.js#L158)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

```suggestion
        // aID might be a string or integer depending on how it is loaded
```
</blockquote></details>

Done.
___
*[src/vue/src/store/modules/user.js, line 148 at r15](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcy-r15-148:-Ms-XNfc68brp9bEKFni:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/44576c98032e9af730eb2dee3686beb1beea1cc6/src/vue/src/store/modules/user.js#L148)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Much better than maintaining reset functions for each page. Reload should mostly be cached by the browser anyway.
</blockquote></details>

Done.

Larspolo (changes requested):
> Reviewed 14 of 18 files at r14, 6 of 10 files at r15, 3 of 3 files at r16, 1 of 1 files at r17, 1 of 1 files at r18, 4 of 4 files at r19, 2 of 2 files at r20, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 21 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/store/modules/user.js, line 148 at r20](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Ms4pHyCCdDIkSjY212O:-Ms4pHyCCdDIkSjY212P:b-vytwl2) ([raw file](https://github.com/ejourn-al/ejournal/blob/b544d00c26fe4878c09b09f55e08a8888a270026/src/vue/src/store/modules/user.js#L148)):*
> ```JavaScript
>     logout () {
>         window.localStorage.clear()
>         window.location.reload()
> ```

Was this tested? Doesnt seem to work on FF

Larspolo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 22 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/components/assignment/AnalyticsMenu.vue, line 22 at r20](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Ms4pvU71XfY1sxQIbKT:-Ms4pvU8EfRFd2pDPms_:b-yr5388) ([raw file](https://github.com/ejourn-al/ejournal/blob/b544d00c26fe4878c09b09f55e08a8888a270026/src/vue/src/components/assignment/AnalyticsMenu.vue#L22)):*
> ```Vue
> 
>         <div>
>             Average points
> ```

Only this is now visible before any stats are loaded in, seems a bit weird

mjvkeulen (commented):
:lgtm:

> Reviewed 2 of 2 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-:-MsUFs_v9SVvQj-ZaWn6:bnfp4nl)* status: all files reviewed, 7 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

___
*[src/vue/src/store/modules/journals.js, line 7 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-7:-MsUBaRv59beG5gZFWSn:b-tiysh0) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L7)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Did you create an issue or did you actually remove course context?
___
*[src/vue/src/store/modules/journals.js, line 169 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-169:-MsUE9-VENuS6ar3bgDM:b-dp4dq7) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L169)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

It is already shared, the analytics page can immediately use the same journals as on the assignment page. I do not like the pattern but already added it back in to the assignment module. Would appreciate if we stop this back and forth discussion with the same arguments and leave this module as is.
</blockquote></details>

Try and see it from my perspective: we already spend a meeting on this with the three of us where the conclusion was that the cache construction offers the best of both worlds. I don't see why we are having this discussion again.
___
*[src/vue/src/views/Analytics.vue, line 158 at r12](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcz-r12-158:-MsUEfO0BhYxlvp8Lq6O:b-yfuf1r) ([raw file](https://github.com/ejourn-al/ejournal/blob/ac62390f1606d61fd720e58d3623782a0e8e5e7d/src/vue/src/views/Analytics.vue#L158)):*
<details><summary><i>Previously, engelhamer (Engel Hamer) wrote…</i></summary><blockquote>

Only because this entire page is loaded at once. I prefer showing parts that are already loaded sooner, but didn't change it on this page. 
</blockquote></details>

I totally agree with showing parts of a view earlier! However, that is not related to using the cache.

mjvkeulen (commented):
> Reviewed 1 of 1 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 4 unresolved discussions (waiting on @engelhamer)

engelhamer:
___
*[src/vue/src/store/modules/journals.js, line 7 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-7:-MsosUDABxrr5LCz7iKH:b-jih4pb) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L7)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Did you create an issue or did you actually remove course context?
</blockquote></details>

There are multiple issues related to this:

https://ejournal.atlassian.net/browse/EJ-769
https://ejournal.atlassian.net/browse/EJ-792

engelhamer (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-:-MsUnMomAS0lg47O6puK:bnfp4nl)* status: 142 of 146 files reviewed, 4 unresolved discussions (waiting on @engelhamer, @Larspolo, and @mjvkeulen)

___
*[src/vue/src/components/assignment/AnalyticsMenu.vue, line 22 at r20](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-Ms4pvU71XfY1sxQIbKT:-MsozmOKE3-jtr73T8sS:b-896fix) ([raw file](https://github.com/ejourn-al/ejournal/blob/b544d00c26fe4878c09b09f55e08a8888a270026/src/vue/src/components/assignment/AnalyticsMenu.vue#L22)):*
<details><summary><i>Previously, Larspolo (Lars) wrote…</i></summary><blockquote>

Only this is now visible before any stats are loaded in, seems a bit weird


</blockquote></details>

Done.
___
*[src/vue/src/store/modules/journals.js, line 169 at r6](https://reviewable.io/reviews/ejourn-al/ejournal/1297#-MrK81PKkDnZm9r3JAcw-r6-169:-MsouiczCWOfJf0c4Upp:b-c95z9w) ([raw file](https://github.com/ejourn-al/ejournal/blob/7ead82130ebccec9d74acf1186c3a642c20b9b04/src/vue/src/store/modules/journals.js#L169)):*
<details><summary><i>Previously, mjvkeulen (Maarten van Keulen) wrote…</i></summary><blockquote>

Try and see it from my perspective: we already spend a meeting on this with the three of us where the conclusion was that the cache construction offers the best of both worlds. I don't see why we are having this discussion again.
</blockquote></details>

Added the cache construction and disabled force for the analytics page.

mjvkeulen (approved):

Larspolo (commented):
> Reviewed 4 of 4 files at r23, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1297)* status: all files reviewed, 3 unresolved discussions (waiting on @engelhamer and @mjvkeulen)

Larspolo (approved):

mjvkeulen (changes requested):
Token timeout still needs to be fixed first

mjvkeulen (approved):
Fixed last blocking issue in 
812ab690cedec32b5a2fe8343b1548dea776d4ea
and 
214ac1aab30d95c52e4a07d3be87c2da00db424a

mjvkeulen (approved):

PR #1308, opened 2022-01-28
Improved support for audio playback

Larspolo:
## Description
image

## Summary of changes
- Fixed #EJ-965

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1308)


src/vue/src/main.js
```diff
@@ -120,6 +120,7 @@ new Vue({
             // No: apng (IE), webp (IE, Safari), tiff (all)
             img: 'bmp, gif, ico, cur, jpg, jpeg, jfif, pjpeg, pjp, png, svg',
             pdf: 'pdf',
+            audio: 'mp3, wav, flac, ogg, oga, webma, m4a',
```
mjvkeulen, line 123: Does this match https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Audio_codecs#common_codecs?

Larspolo: It matches caniuse.com

mjvkeulen: Mapped onto which browser (and versions?)

src/vue/src/components/assets/file_handling/FileDisplay.vue
```diff
@@ -17,9 +17,16 @@
     >
         <slot/>
     </pdf-display>
+    <audio-file-display
```
mjvkeulen, line 20: Nice so we can now also attach audio feedback to comments?

src/vue/src/main.js
Larspolo: image
With some partial support just added because of the limited use of those browser versions

src/vue/src/components/assets/file_handling/FileDisplay.vue
Larspolo: Yes, but not record inside eJournal

mjvkeulen (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1308)* status: all files reviewed, 1 unresolved discussion (waiting on @engelhamer and @mjvkeulen)

engelhamer (approved):
Ensured that the audio component matches our style more by rounding the borders (notif

image

image

PR #1309, opened 2022-01-31
Fixed migration that caused errors on test database

Larspolo:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1309)


src/django/VLE/migrations/0104_moved_preset_only_to_template_chain.py
```diff
@@ -6,7 +6,7 @@
 def transfer_preset_only_to_template_chain(apps, schema_editor):
     TemplateChain = apps.get_model('VLE', 'TemplateChain')
     Template = apps.get_model('VLE', 'Template')
-    for template_chain in TemplateChain.objects.all():
+    for template_chain in TemplateChain.objects.filter(template__isnull=False):
```
mjvkeulen, line 9: How about we remove isolated TemplateChain instances instead?

Larspolo: This worked, dont want to spend more time on it, you may improve it if you want

mjvkeulen: So what happens with isolated TemplateChain instances in the future?

mjvkeulen: These should not exist -> just remove them

Larspolo: Think that would also not solve the issue for future instances, so I created this issue https://ejournal.atlassian.net/browse/EJ-1435

engelhamer (approved):

mjvkeulen (approved):

PR #1320, opened 2022-02-10
Speedup group journal load times

mjvkeulen:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1320)


engelhamer:
Assignment is also in persisted state. Reactivity is prefered.

engelhamer:
I think the current situation is ultimately most beneficial for users:

- Page reload is quicker, since the stored assignment persists
- No outdated information remains, as it is always updated behind the scenes

mjvkeulen:
The assignment call is to compute heavy to just refresh on each page. Fetching the assignment each prev/next action, or each PFB activity step is simply not required, but greatly impacts other users.

Force loading the assignment should be a page-specific decision, not some global action.

Larspolo:
Possibly an important detail for this discussion: the assignment retrieve does not do anything with the stats that are retrieved. We should possibly change this discussion to if (and when) we want to get all the journals with their stats

engelhamer (approved):

engelhamer (changes requested):
Assignment due date, name, points possible etc. changes by some other user -> Current user needs to log out and back in

mjvkeulen:
> Current user needs to log out and back in

Agreed, I have removed the assignment module from the persisted state.

PR #1321, opened 2022-02-10
Added a setting which allows teachers to stop sending grades to the LMS

Larspolo:
image


src/django/VLE/migrations/0111_assignment_require_lti_link.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('VLE', '0110_alter_feedbackform_time_spent'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='assignment',
+            name='send_grades_to_lms',
```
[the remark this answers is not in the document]
Larspolo: I had that first. But this matched better with also the usecase of another teacher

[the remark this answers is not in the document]
engelhamer: That could be an indication at the assessment section like we have for the grade passback errors.

Now it is not possible to have a combination of LTI and non-LTI users. This may be inconvenient later on.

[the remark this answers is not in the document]
Larspolo: Could be another setting when some users wants it? Because its a bit more effort

src/vue/src/components/assignment/AssignmentDetails.vue
```diff
@@ -94,11 +94,35 @@
...
+            ]"
+            class="float-right mb-3 ml-3 mt-3"
+        />
+        <h2 class="theme-h2 field-heading mb-2 mt-3">
+            Send grades to the LMS
+        </h2>
+        <small>
+            With this settings enabled, students will be requested to visit the assignment on the LMS first,
+            before they can start working on their journal.
+        </small>
```
engelhamer, line 119: These options should only be visible if an LMS connection exists at all.

engelhamer (changes requested):

src/vue/src/components/assignment/AssignmentDetails.vue
Larspolo: Would you want this on assignment creation and then later link it?

engelhamer: I would think that it is false by default and hidden for the user. If at some point an LTI connection is created, this should be shown and set to true.

Would not want users to  disable this by accident.

engelhamer: You want it to only be visible for superusers?

Larspolo: I ment as in, teachers maybe dont know about this setting

mjvkeulen: Shall we add a header for this setting to the help center? Similar to assign to, just something we can send to Rene for example. 

https://help.ejournal.app/teachers/configuring-your-assignment/assignment-settings/#assign-to

src/django/test/test_journal.py
```diff
@@ -1366,22 +1366,35 @@ def needs_lti_link(journal):
             assert name in [student2.full_name, student3.full_name]
 
         qry_journal = Journal.all_objects.filter(pk=journal.pk).allowed_journals().annotate_needs_lti_link().get()
+        print(old_needs_lti_link)
```
engelhamer, line 1369: Left

engelhamer (approved):

src/vue/src/components/assignment/AssignmentDetails.vue
Larspolo: https://github.com/eJourn-al/knowledge-base/pull/33

mjvkeulen (approved):

src/vue/src/components/journal/UserMissingLtiLinkWarning.vue
```diff
@@ -8,7 +8,10 @@
             <span v-if="assignment.is_group_assignment">
                 <b>Warning:</b> The following journal members have not visited the assignment in the active
                 LMS (Canvas) course '{{ assignment.active_lti_course.name }}' yet:
-                <ul class="pt-1 pb-1 mb-0">
+                <ul
+                    v-if="currentJournal.needs_lti_link"
```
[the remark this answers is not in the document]
Larspolo: Engel his review comment. Would rather fully exclude the UL

PR #1323, opened 2022-02-11
Improved the comment list view and viewable checks

Larspolo:
Carefully read the assert statements in the tests to see if you agree. Some stuff has changed.

Mostly problems with if you had two feedback forms, and they could see eachothers comments, but one had publishing only after due date, and the other not. Then they could still see eachothers comments.

Also teacher comments are not better excluded from the whole buildup

And I think the comments list view is now way nicer: less difficult to accidentally change something, and more clear what it does

Also made use of a lot of building functions, see if you agree with that terminology


src/django/VLE/models/journal_content/comment.py
```diff
@@ -240,7 +288,7 @@ def check_editable_by(
 
         user.check_permission('can_edit_staff_comment', self.journal.assignment)
 
-    @permission_wrap()
+    @ permission_wrap()
```
mjvkeulen, line 291: Lots of these

src/django/test/test_file_handling.py
```diff
@@ -147,8 +148,11 @@ def test_file_retrieve(self):
 
         feedback_form = factory.PeerFeedbackForm(
             journal=self.journal,
+            preset_peer_feedback_node__can_view_other_comments=True,
         )
         reviewer = feedback_form.reviewer_journal.authors.first().user
+        assert Comment.objects.viewable_by(reviewer, (published_comment.journal, feedback_form)).exists(), \
+            'Comment should be vissible for the reviewer via the feedback form'
```
mjvkeulen, line 155:
```suggestion
            'Comment should be visible for the reviewer via the feedback form'
```

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
+        return self.distinct('pk')
+
+    def filter_based_on_published_state(
+        self,
+        feedback_form: Union['VLE.models.FeedbackForm', None],
+        is_in_journal: bool,
+        is_grader: bool,
+    ):
+        is_published = self.build_q_is_published()
+        # Journal authors should only be able to view published comments
```
mjvkeulen, line 59: Unless the comment is their own right?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
-            feedback_form=feedback_form,
-            author__in=feedback_form.get_reviewers(),
+    def filter_on_reviewer_comments(self, func='filter') -> 'CommentQuerySet':
+        return getattr(self, func)(self.build_q_author_is_reviewer())
+
+    @staticmethod
+    def build_q_author_is_reviewer():
+        return (
+            Q(author=F('feedback_form__reviewer')) |
+            Q(author=F('feedback_form__reviewer_journal__authors__user'))
```
mjvkeulen, line 101: Q) We can not both assign either a single user column or multiple to the same annotation right?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
+        preset_entry_nodes_related_to_fbf_qry = VLE.models.PresetEntryNode.objects.filter_on_feedback_form(
+            feedback_form=feedback_form,
+            entry_qry=entries_related_to_fbf_qry,
+        )
+        has_relation_to_entries_related_to_fbf = self.build_q_through_node(
+            entry__in=entries_related_to_fbf_qry,
+            available_node_classes=VLE.models.Entry.get_available_node_classes(),
+        )
+        return (
+            ~Q(feedback_form=feedback_form) |
```
mjvkeulen, line 145: Q) How come?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -134,26 +197,14 @@ def check_bulk_viewable_by(
...
-            Q(author=user) |
-            # Authors of feedbackforms should share unpublished comments without
-            # them needing the can_grade permission
-            Q(feedback_form__reviewer_journal=journal_of_user, feedback_form__reviewer_journal__isnull=False) |
-            Q(feedback_form__reviewer=user, feedback_form__reviewer__isnull=False)
-        ).exists():
-            user.check_permission('can_grade', journal.assignment)
+        _, journal, feedback_form = context
+
+        if objects.viewable_by(user, (journal, feedback_form)).count() < objects.count():
```
mjvkeulen, line 206: Should this not be a subset comparison based on id?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -254,6 +298,9 @@ def save(self, *args, **kwargs):
         self.text = sanitization.strip_script_tags(self.text)
         super(Comment, self).save(*args, **kwargs)
 
+        # TODO PFB: immediately published comment do not get a notification as of now
```
mjvkeulen, line 301: Tagging

src/django/VLE/utils/import_utils.py
```diff
@@ -144,6 +144,7 @@ def import_comment(comment: Comment, copied_node: JournalEntryNode):
         comment (:model:`VLE.comment`): Comment to import.
         copied_node: Node to attach comment to.
     """
+    # TODO PFB: think what PFB comments should be imported, we cannot simply use .published
```
mjvkeulen, line 147: Good catch, any feedback form is not copied. So should these not simply be published comments without FBF?

src/django/test/test_comment.py
```diff
@@ -1055,11 +1114,14 @@ def test_list_comments_on_single_node_feedback_form_author(self):
...
-                self.published_related_comment_different_author.pk,
+                self.published_related_comment_different_reviewer.pk,
                 self.published_related_comment_journal_author.pk,
                 self.published_preset_entry_node_comment_journal_author.pk,
+                self.published_related_teacher_comment.pk,
+                self.unpublished_related_teacher_comment.pk,
             },
-            'When supplying a node & journal, the comments should be limited to only the specific node'
+            'When supplying a node & journal, the comments should be limited to only the specific node. '
+            'Also, unpublished teacher comments should be shown to reviewers'
```
mjvkeulen, line 1124: Q) What was the reasoning behind this? 

Do we feel a teacher expects their unpublished comment to be viewable to anyone other than staff?

src/django/VLE/models/journal_content/comment.py
Larspolo: Note that these issues are not introduced in this PR, simply illuminated. These should at the very least be moved to issues (if not fixed in this PR)

src/django/VLE/models/journal_content/comment.py
Larspolo: It doesn't make sense, but technically it might be possible

src/django/VLE/models/journal_content/comment.py
Larspolo: Comments outside of the feedback form might be shown. `filter_based_on_feedback_form` limits it to only the feedback form if its not allowed

src/django/VLE/models/journal_content/comment.py
Larspolo: `viewable_by` should strictly be a filter, this way no new IDs could be added. Seems like ID comparison would only increase the runtime

src/django/test/test_comment.py
Larspolo: This is only when a teacher posts a comment inside the feedback form. This way the teacher can communicate with the reviewers (unpublished) comments. Thought that might be a useful feature

src/django/VLE/models/journal_content/comment.py
mjvkeulen: Does this not include all comments which are NOT part of the feedback form?

Larspolo: Yes, intensional. This function needs to only limit stuff about the feedback form in relation to the provided user

engelhamer (approved):

src/django/VLE/models/journal_content/comment.py
Larspolo: Not sure what you are getting at, possibly missing your point. This query it not for annotations. It checks if the comment author is a reviewer

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
-            q
-            | Q(preset_peer_feedback_node=feedback_form.preset_peer_feedback_node)
-            | Q(preset_entry_node__in=VLE.models.PresetEntryNode.objects.filter_on_feedback_form(
-                feedback_form=feedback_form,
-                entry_qry=entries_related_to_fbf_qry,
-            ))
-        )
+        self = self.filter_based_on_published_state(feedback_form, is_in_journal=is_in_journal, is_grader=is_grader)
+        if feedback_form:
+            self = self.filter_based_on_feedback_form(feedback_form)
```
mjvkeulen, line 48: [NIT]: `filter_on_feedback_form` and `filter_based_on_feedback_form`, quite the difference with a seemingly equal name.

src/django/VLE/models/journal_content/comment.py
mjvkeulen: This breaks the expected behaviour of the function (`related_to_feedback_form`?). Can this not be moved outside of the function?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
+        self,
+        feedback_form: 'VLE.models.FeedbackForm',
+    ):
+        """Limit comments to comment related to the same node as the provided feedback form (peer feedback node)
 
-        return self
+        If the peer feedback node allows, also include comments that are not related to the feedback form
+        But are still in the same node
+        """
+        q_comments_related_to_feedback_form = self.build_q_comments_related_to_feedback_form(feedback_form)
```
mjvkeulen, line 84: This also included comments for which (~Q(feedback_form=feedback_form)), confusing right?

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
-                feedback_form=feedback_form,
+            if not feedback_form:
+                raise VLEForbiddenError('You are required to provide a feedback form to view comments of this journal.')
+            return self.filter(
+                Q(feedback_form=feedback_form) | is_published
             )
 
-        if reviewer_only_comments:
-            self = self.filter_on_reviewer_comments(feedback_form)
+    def filter_based_on_feedback_form(
```
mjvkeulen, line 75: We have `filter_based_on_feedback_form` and `build_q_comments_related_to_feedback_form`. Is `build_q_comments_related_to_feedback_form` intended to be reused elsewhere? Just a helper function? With the removal of ~Q(feedback_form=feedback_form) the call would function more as expected to me at least.

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
-        return self
+        If the peer feedback node allows, also include comments that are not related to the feedback form
+        But are still in the same node
+        """
+        q_comments_related_to_feedback_form = self.build_q_comments_related_to_feedback_form(feedback_form)
+        if (
+            feedback_form.preset_peer_feedback_node
+            and not feedback_form.preset_peer_feedback_node.can_view_other_comments
+        ):
+            q_comments_related_to_feedback_form = q_comments_related_to_feedback_form.add(
```
mjvkeulen, line 89: So now we have .filter(~Q(feedback_form=feedback_form) | .... | .... | .... & feedback_form=feedback_form). How about moving ~Q(feedback_form=feedback_form) to the else of this block as an `or`?

src/django/VLE/models/journal_content/comment.py
mjvkeulen: How does this work when a journal author is self-reviewing? Can they not post unpublished comments then (like other reviewers?)

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
+        elif is_grader:
+            return self
+        # Reviewers should see:
+        # - all comments within the provided feedback form
+        # - all other published comments
         else:
-            self = self.filter(
-                feedback_form=feedback_form,
+            if not feedback_form:
+                raise VLEForbiddenError('You are required to provide a feedback form to view comments of this journal.')
```
mjvkeulen, line 70: [NIT] As this explicitly regards reviewers, we could use that in the message. 

```suggestion
                raise VLEForbiddenError('You need to be invited to review this journal, to view any of its comments.')
```

src/django/VLE/models/journal_content/comment.py
```diff
@@ -10,79 +10,142 @@
...
-                VLE.models.PresetPeerFeedbackNode.IMMEDIATELY,
-                VLE.models.PresetPeerFeedbackNode.REVIEWER_DECIDES,
-            ]) |
-            Q(feedback_form__preset_peer_feedback_node__due_date__lt=timezone.now()),
+    @staticmethod
+    def build_q_is_published():
+        comment_author_is_reviewer = CommentQuerySet.build_q_author_is_reviewer()
+        has_attached_feedback_form = Q(feedback_form__isnull=False)
+        PPFBN = VLE.models.PresetPeerFeedbackNode
+        peer_feedback_node_considers_it_published = (
```
mjvkeulen, line 109: Really like the readability of this structure!

src/django/VLE/models/journal_content/comment.py
mjvkeulen: Could you elaborate? This id comparison would be in memory of the already fetched items. The runtime would not be an issue. 

I don't see how this is guaranteed to be safe atm.

src/django/VLE/models/journal_content/comment.py
mjvkeulen: Totally agree! But still tagging makes sense so that we do not merge TODOs

mjvkeulen (changes requested):

src/django/VLE/models/journal_content/comment.py
Larspolo: Not sure how that currently works. But I would find that a bit weird. Think posting only published comments there is fine as well

src/django/VLE/models/journal_content/comment.py
Larspolo: Makes sense, I moved it to `filter_based_on_feedback_form`

src/django/VLE/models/journal_content/comment.py
Larspolo: Think (hope) `.count()` does not fetch any items. This can be done in SQL as well.

If you have a set of items, and you use a function that can only take items away, then checking for the cardinality of the set is checking if it remained the same right?

src/django/VLE/models/journal_content/comment.py
Larspolo: Think we could remove the `filter_on_feedback_form` function

src/django/VLE/models/journal_content/comment.py
Larspolo: It was possible, I added the (a bit weird) behaviour of journal authors in self reflection only seeing unpublished comments when inside the form

PR #1325, opened 2022-02-11
Improved the display for maximum uses of template

Larspolo:
## Description
Two fixes:
- Improved the display when no more templates are available
With multiple templates being free use, but all locked
image
With 1 template being free use, and locked
image

- Fixed an issue where the JEN would be left behind after deleting an entry

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1325)


src/vue/src/components/timeline/TimelineNode.vue
```diff
@@ -16,6 +16,7 @@
     >
         <timeline-node-circle
             :node="node"
+            :nodes="nodes"
```
engelhamer, line 19: Better to use the store for nodes if we are realy using it everywhere.

Larspolo: Is there a nodes store? Looked like we didnt use it for rhe timeline

Larspolo: Would like to move it to an technical debt issue instead. I see many other improvements that can also be done with this move, and would like to deploy this asap

engelhamer (approved):

PR #1326, opened 2022-02-13
Improved timeline size and other small UI tweaks

engelhamer:
## Description
This makes it easier to navigate the timeline when content becomes larger.

Also fixed some minor issues / tweaks.

image


src/vue/src/components/category/CategoryDisplay.vue
```diff
@@ -10,6 +10,7 @@
             :category="category"
             :removable="editable"
             :showInfo="true"
+            class="align-middle"
```
Larspolo, line 13: image
Category text seems to be miss-aligned

src/vue/src/components/timeline/TimelineNode.vue
```diff
@@ -93,19 +94,20 @@ export default {
 
 <style lang="sass">
 .node-container
-    min-height: 90px
-    margin-left: 37px
-    padding: 0px 0px 20px 50px
+    &.add-node-container
+        min-height: 50px
```
Larspolo, line 98: Not sure where this is introduced, but the "new entry" text is not not in the middle anymore
image

src/vue/src/sass/global.sass
```diff
@@ -22,13 +22,6 @@
 @import '~sass/partials/tours.sass'
 @import '~sass/partials/colors.sass'
 
-body
-    color: $text-color
-    font-family: 'Roboto', sans-serif
-    font-size: 1em
```
Larspolo, line 28: image

Selection text seems to be bigger than other inputs

src/vue/src/components/category/CategoryDisplay.vue
Larspolo: Same goes for categories
image

src/vue/src/components/timeline/TimelineNodeCircle.vue
```diff
@@ -225,18 +226,17 @@ export default {
 
 <style lang="sass">
 .timeline-node-circle
```
Larspolo, line 228: [NIT] The pixel difference between activity types (with and without categories too) feel inconsistent. Might already have been the case, but it feels worse now that they are closer to eachother

image

Larspolo (changes requested):

src/vue/src/components/timeline/TimelineNodeCircle.vue
mjvkeulen: image

Does the lack of symmetry between nodes in the timeline not bother you Engel? I would expect there to be a similar distance between the nodes regardless of whether the hold categories or not.

src/vue/src/components/category/CategoryDisplay.vue
engelhamer: That's odd, it should be aligned in middle (as the class suggests). Which browser is this?

src/vue/src/components/timeline/TimelineNode.vue
engelhamer: Same as the previous align-middle stuff

src/vue/src/components/timeline/TimelineNodeCircle.vue
engelhamer: That means that you are wasting vertical space. The categories are making some nodes larger than others.

I reduced the feeling of asymmetry by only showing categories in full when the node is selected rather than on hover.

src/vue/src/components/category/CategoryDisplay.vue
engelhamer: Cannot reproduce it on firefox, edge and chrome (windows)

Larspolo (commented):
Improved some of the issues, but couldnt fix the theme-select font size issue

Larspolo (approved):

PR #1327, opened 2022-02-14
Improved the error message of uploading an invalid Kaltura field

Larspolo:
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1327)


src/django/VLE/validators.py
```diff
@@ -67,7 +67,10 @@ def validate_video_data(field, data):
         try:
             validate_kaltura_video_embed_code(data)
         except ValidationError:
-            validate_youtube_url_with_video_id(data)
+            try:
+                validate_youtube_url_with_video_id(data)
+            except ValidationError:
+                raise ValidationError('Entry a valid YouTube video URL or Kaltura embed code.')
```
mjvkeulen, line 73: Message is raised in validate already

engelhamer (approved):

PR #1330, opened 2022-02-17
Improved the speed of the assignment and analytics pages

Larspolo:
- Halved the load time of the journal load in group assignment
- Minimized the analytics panel first showing, reducing the loading time from about 2 + loadAssignment to just 2. loadAssignment being able to take up to 10 seconds in group assignments

---
This change is [Reviewable](https://reviewable.io/reviews/ejourn-al/ejournal/1330)


src/django/VLE/views/journal.py
```diff
@@ -93,7 +93,7 @@ def list(self, request):
         # Checking the journals specifically is done inside this function
         assignment.check_viewable_by(request.user, fields=[])
         journal_query: VLE.models.JournalQuerySet = (
-            Journal.objects.annotate_stats().filter_on_assignment(assignment).filter_on_course(course)
+            Journal.objects.annotate_stats().filter_on_assignment(assignment)
```
mjvkeulen, line 96: I am still unsure what our "course specific context" intention is. With this change, we remove this distinction entirely?

Larspolo: The serializer still contains information about the course, not sure what exactly, but now all journals will be serialized, also from other courses. Given that we already wanted to go this way, and that it now greatly reduces the loadtime, I introduced this change

engelhamer (approved):

PR #1338, opened 2022-02-20
Reduced the amount of redirects users experience when getting errors

Larspolo:
After the permission PR, we stopped redirecting forbidden errors. However, they have since been added back again. This means that users experience more redirects to error pages. In this PR I went over all forbidden raises, and changed some to bad request in the cases where I think the user should not be redirected to the error page.

Feel free to discuss each case, and possibly go over them yourself, there are about 80 of them (`raise VLEForbiddenError` & `raise VLEBadRequest`)


engelhamer (commented):
I think that this is more of a 'workaround' than a real solution: why don't we focus on the particular call to decide whether a redirect is appropriate or not? Is seems weird that you have to change the error code (albeit the original error code would suit it better, evident from the fact that it was configured as such first).

`connArgs.redirect` could be used more sparsely?

Larspolo:
I agree that another filter there would be good as well (at least, to pay more attention to it). But I also think that they should always be treated the same: when creating an entry for a deadline, the error message "Deadline already contains an entry." would preferably be send as a toast. While "You are not allowed to view this journal" would be better for a redirect

mjvkeulen:
> why don't we focus on the particular call to decide whether a redirect is appropriate or not

@engelhamer  Just checking we are on the same level, but you are proposing to use `connArgs.redirect` more often in the actual calls themselves right?

src/django/VLE/models/journal_content/assessment_feedback_form.py
```diff
@@ -353,7 +353,7 @@ def check_editable_by(
         return_boolean: bool = False,
     ):
         if any(field not in self.EDITABLE_FIELDS for field in fields):
-            raise VLEForbiddenError(
+            raise VLEBadRequest(
```
mjvkeulen, line 356: I am all for streamlining which HTTP error codes we use and when. But we should possibly do this from a higher-level perspective first? Go over the expected guidelines and see where we deviate?

mjvkeulen: I say this because currently I don't know which of these two (403 or 400) is more appropriate for a lot of these scenarios.

PR #1344, opened 2022-02-22
Removed arbitrary abbreviation limitations

Larspolo:
Some LMS'es have larger abbreviations than 10 characters, they would not be able to be ported to eJournal


src/django/VLE/models/course.py
```diff
@@ -20,10 +20,7 @@ class Course(AbstractBaseModel):
     """
 
     name = models.TextField()
-    abbreviation = models.TextField(
```
mjvkeulen, line 23: The abbreviation is shown in locations where there is limited space, I would appreciate a reasonable limit here?

mjvkeulen: Upper char limit makes sense, truncated display (also whilst configuring the abbreviation)?

Larspolo: But why limit it in the backend?

Larspolo: We do a frontend warning for users for the node title

mjvkeulen: > But why limit it in the backend?

Text fields are meant for large (unknown/unlimited) length fields, char field allows setting a limit (which has meaning). If you want to increase the limit to 20 chars fine. But what is unlimited going to help?

Larspolo: Canvas allows for at least 40 characters, so to match those courses without requiring the teacher to shorten the abbreviation

mjvkeulen: From what I could find both name and abbreviation have a limit of 255 chars in the Canvas. How about we match that setting?

Larspolo: I found both Blackboard (Learn) and Moodle limiting the name to 255 chars and abbreviation to 100 chars. Cannot fins anything about Brightspace. But changing to 255 is fine by me.

PR #1348, opened 2022-02-24
Added rubric display toggle to peer feedback activity instructions

mjvkeulen:
Added rubric display toggle to peer feedback activity instructions

Fixed a bug where rubrics could be added to a peer feedback activity, whilst being set to visible until used in an assessment.

image
image


src/vue/src/components/rubric/RubricViewToggle.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+    <div>
+        <b-button
+            v-if="rubricId"
+            class="blue-button"
+            variant="link"
+            @click="viewRubric = !viewRubric"
+        >
+            <icon name="eye"/>
+            View rubric
```
Larspolo, line 10: Would like this to toggle with the eye-slash icon

mjvkeulen: This is currently the same as the template select preview

Larspolo:
Disagree with the hotfix label

src/vue/src/components/rubric/RubricViewToggle.vue
Larspolo: Guess it can also be changed there, template import is like this
image

Also rubric import, entry preview and managing teacher entries

src/django/VLE/models/nodes/node_types/feedback_node.py
```diff
@@ -6,6 +6,7 @@
 from django.core.exceptions import ValidationError
 from django.db import models
 
+from ...rubric import Rubric
```
Larspolo, line 9: [NIT] Prefer using `VLE.models.[Model]` in the code itself, to never have to deal with circular imports

src/vue/src/components/assignment_editor/timeline_controlled/preset_node/PresetNodePeerFeedbackReviewSettings.vue
```diff
@@ -235,8 +235,9 @@
                             v-for="rubric in rubrics"
                             :key="rubric.id"
                             :value="rubric.id"
+                            :disabled="rubric.visibility != 'v'"
```
Larspolo, line 238: Looks good!

src/vue/src/store/modules/assignmentEditor.js
```diff
@@ -312,6 +312,7 @@ const mutations = {
                 id: -1,
                 name: '',
                 description: '',
+                visibility: 'v',
```
Larspolo, line 315: I actually preferred teacher having to set this setting, as teachers might not consider changing this setting otherwise, and leaking the rubric visibility.

We could add  this, but then I would prefer adding some indication in the rubric menu that the item is visible for students

Larspolo (changes requested):

src/vue/src/components/rubric/RubricViewToggle.vue
mjvkeulen: This makes use of the footer, which means the component has to be broken into pieces. I quite like the encapsulation, if its good enough for the template selection I think this is a fine addition as well.

PR #1350, opened 2022-02-25
Fixed an issue where a newly created activity would not be selected after creation

mjvkeulen:
Fixed an issue where a newly created activity would not be selected after creation

- FIXES EJOURNAL-FRONTEND-2NA
- FIXES EJOURNAL-FRONTEND-2D9
- FIXES EJOURNAL-FRONTEND-2H2


Larspolo (approved):
I think at some point we wanted to make less use of the return value of the module functions, and instead use the getters afterwards again

